PR #21, opened 2019-02-07
fix link to asciinema demo

phoracek:
Original link expired since it was not connected to an account.


phoracek:
@yuvalif could you please merge?

PR #22, opened 2019-02-08
rename k8s-node-net-conf to kubernetes-nmstate

phoracek:
At some point, repository containing this code was renamed and now
it is not possible to recompile the project. With this patch we
fix project name to match current repository and make `make build`
possible again.


manifests/state-client-pod.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   serviceAccountName: nmstate-state-controller
   containers:
   - name: state-client
-    image: yuvalif/k8s-node-net-conf-client
+    image: yuvalif/kubernetes-nmstate-client
```
yuvalif: this image does not exists on hub.docker.com under yuvalif

phoracek: Right, good point.

phoracek: @yuvalif should be fixed now, could you please take another look?

PR #24, opened 2019-02-10
build manifests based on yaml templates

phoracek:
Replace current templating system based on marshalling of binary
Go structures to a more readable format - yaml files templating.

While binary structures are better to assure that generated manifest
will be valid, it is hard to read their sources and create new. This
new approach should make it easier and faster to add new templates.

This patch also makes it possible to parametrize templates, set
different namespace or image names. This will be used in follow-up
patches to generate custom daemon sets for local cluster deployment.
This can be later also used for operator framework or stable releases.


phoracek:
@yuvalif could you please take a look? I like the original source code based approach but it is really hard to read it and extend it with daemon sets etc.

@SchSeba could you please review?

phoracek:
Oh wait, have to fix testing scripts.

PR #25, opened 2019-02-11
add local cluster support

phoracek:
This patch provides local kubernetes/openshift cluster based on
kubevirtci. This cluster can be used to try out kubernetes-nmstate and
it will be later used also for functinal tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,15 @@ build:
 	cd cmd/state-controller && go fmt && go vet && go build
 	cd cmd/policy-controller && go fmt && go vet && go build
 
-DOCKER_REPO=yuvalif
+IMAGE_REGISTRY ?= yuvalif
 
 docker: build
-	cd cmd/client && docker build -t $(DOCKER_REPO)/k8s-node-net-conf-client .
-	cd cmd/state-controller && docker build -t $(DOCKER_REPO)/k8s-node-network-state-controller .
+	cd cmd/client && docker build -t $(IMAGE_REGISTRY)/k8s-node-net-conf-client .
+	cd cmd/state-controller && docker build -t $(IMAGE_REGISTRY)/k8s-node-network-state-controller .
+
+docker-push: build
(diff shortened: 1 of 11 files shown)


phoracek:
@SchSeba @yuvalif could you please take a look?

phoracek:
@yuvalif all good :)

cluster/up.sh
```diff
@@ -0,0 +1,71 @@
...
+    ./cluster/cli.sh scp /etc/origin/master/admin.kubeconfig - > ./cluster/.kubeconfig
+
+    # Update Kube config to support unsecured connection
+    cluster_port=$(./cluster/cli.sh ports ocp | tr -d '\r')
+    ./cluster/.kubectl config set-cluster node01:8443 --server=https://127.0.0.1:$cluster_port
+    ./cluster/.kubectl config set-cluster node01:8443 --insecure-skip-tls-verify=true
+fi
+
+echo 'Install NetworkManager and Open vSwitch on nodes'
+for i in $(seq 1 ${KUBEVIRT_NUM_NODES}); do
```
SchSeba: if you start the ovs here it will break the openshift deployment no?

Makefile
```diff
@@ -5,11 +5,15 @@ build:
 	cd cmd/state-controller && go fmt && go vet && go build
 	cd cmd/policy-controller && go fmt && go vet && go build
 
-DOCKER_REPO=yuvalif
+IMAGE_REGISTRY ?= yuvalif
```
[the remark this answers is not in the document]
phoracek: We will, but in a different PR. I'd like to merge state-client and state-controller images into one (for easier mainating, they share the same dependencies and most of the code). Then we should create a new repo on nmstate quay registry and introduce Travis builds.

cluster/up.sh
phoracek: You are absolutely right!

phoracek:
@SchSeba fixed, could you please take another look?

PR #27, opened 2019-02-14
keep client and controller in a single image

phoracek:
With this PR we:
- move client to the same image as controller since they share all dependencies
- use multi-stage builds so it is easier to automate building
- set images repo to nmstate quay.io registry


phoracek:
@SchSeba could you please take a look? There are some supporting patches not really part of the PR topic. I just wanted some tweaks to piggy back on the main commit. It might be easier to review this commit by commit. Thanks!

Makefile
```diff
@@ -1,51 +1,39 @@
...
+IMAGE_REGISTRY ?= quay.io/nmstate
+IMAGE_TAG ?= latest
+PULL_POLICY ?= Always
+STATE_HANDLER_IMAGE ?= kubernetes-nmstate-state-handler
+POLICY_HANDLER_IMAGE ?= kubernetes-nmstate-configuration-policy-handler
 
-IMAGE_REGISTRY ?= yuvalif
+build:
+	cd cmd/state-handler && go fmt && go vet && go build
+	cd cmd/policy-handlerm && go fmt && go vet && go build
```
yuvalif: why "policy-handlerm"?

cmd/state-handler/main.go
```diff
@@ -0,0 +1,126 @@
...
+	kubeInformerFactory.Start(stopCh)
+	nmstateInformerFactory.Start(stopCh)
+
+	if err = controller.Run(2, stopCh); err != nil {
+		klog.Fatalf("Error running controller: %s", err.Error())
+	}
+}
+
+func client() {
+	flag.Parse()
```
yuvalif: why do you need to call "parse" again?

cmd/state-handler/main.go
```diff
@@ -0,0 +1,126 @@
...
+
+	kubeClient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatalf("Error building kubernetes clientset: %s", err.Error())
+	}
+
+	nmstateClient, err := clientset.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatalf("Error building nmstate clientset: %s", err.Error())
+	}
```
yuvalif: minor: BuildConfigFromFlags(), clientset.NewForConfig(), GetNsmespace() and GetHostName() are common to both client and controller, could be taken out of the specific functions.

cmd/state-handler/main.go
```diff
@@ -0,0 +1,126 @@
...
+		klog.Fatalf("Error listing all node network states (in %s): %v\n", ns, err)
+	}
+
+	nodeFound := false
+	name := utils.GetHostName(*hostname, cfg, ns)
+	if name == "" {
+		klog.Fatalf("Failed to get host name\n")
+	}
+
+	for _, state := range list.Items {
```
yuvalif: inside the controller code, don't we have similar initialization to the one we perform here?
or is it all based on getting updated from k8s?

Makefile
phoracek: Fixed

cmd/state-handler/main.go
phoracek: Fixed

phoracek:
@yuvalif thanks for the reviews. I addressed all of your comments. Most of them are resolved inhttps://github.com/nmstate/kubernetes-nmstate/pull/27/commits/682f6df898f87842ebcc5986330b2699ae8def0a

cmd/state-handler/main.go
phoracek: oh, missed this one. will fix it in a second

phoracek: I don't see how could we merge the logic at this point. I need to research more about the controller. Could we take this PR as it is?

PR #28, opened 2019-02-15
run basic check on travis

phoracek:
With this PR we add format check script to the project and also execute this check, build and docker build on travis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,9 @@ manifests:
 		hack/generate-manifests.sh
 
 check:
+	./hack/verify-codegen.sh
 	./hack/verify-fmt.sh
+	./hack/verify-vet.sh
 
 test:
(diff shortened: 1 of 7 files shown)


phoracek:
@SchSeba could you please take a look?

Makefile
```diff
@@ -38,6 +38,9 @@ manifests:
 	STATE_CONTROLLER_IMAGE=$(STATE_CONTROLLER_IMAGE) \
 		hack/generate-manifests.sh
 
+check:
+	./hack/verify-fmt.sh
```
SchSeba: maybe we can run `go fmt` and `go vet` then check with git if any file was change and fail the test?

phoracek: should be done now, could you take another look please?

Issue #41, opened 2019-05-17
use newer version of nmstate


phoracek:
We haven't upgraded used nmstate version in a while. We should.

PR #47, opened 2019-05-24
Use fedora RPMs to install nmstate

qinqon:
Maybe we can use fedora RPMs instead of pip and github, right now is nmstate v0.0.4

Related to #41

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,13 @@ build:
 	cd cmd/state-handler && go fmt && go vet && go build
 	cd cmd/policy-handler && go fmt && go vet && go build
 
-docker:
+docker-state-handler:
 	docker build -f cmd/state-handler/Dockerfile -t $(IMAGE_REGISTRY)/$(STATE_HANDLER_IMAGE):$(IMAGE_TAG) .
+
+docker-policy-handler:
 	docker build -f cmd/policy-handler/Dockerfile -t $(IMAGE_REGISTRY)/$(POLICY_HANDLER_IMAGE):$(IMAGE_TAG) .
+
+docker: docker-state-handler docker-policy-handler
 
 docker-push:
 	docker push $(IMAGE_REGISTRY)/$(STATE_HANDLER_IMAGE):$(IMAGE_TAG)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,9 +13,13 @@ build:
 	cd cmd/state-handler && go fmt && go vet && go build
 	cd cmd/policy-handler && go fmt && go vet && go build
 
-docker:
+docker-state-handler:
 	docker build -f cmd/state-handler/Dockerfile -t $(IMAGE_REGISTRY)/$(STATE_HANDLER_IMAGE):$(IMAGE_TAG) .
-	docker build -f cmd/policy-handler/Dockerfile -t $(IMAGE_REGISTRY)/$(POLICY_HANDLER_IMAGE):$(IMAGE_TAG) .
+
+docker-policy-handler:
```
SchSeba: you also need to split the docker-push command.

the CI is failing because of that.

qinqon: @SchSeba well the failure is using the STATE image at policy hander, let me fix it.

cmd/state-handler/Dockerfile
```diff
@@ -4,10 +4,11 @@ COPY . .
...
 FROM centos:7
-RUN yum -y install epel-release && \
+RUN yum-config-manager --add-repo https://copr.fedorainfracloud.org/coprs/nmstate/nmstate-el7/repo/epel-7/nmstate-nmstate-el7-epel-7.repo && \
+    yum -y install epel-release && \
     yum -y update && \
     yum -y install \
-        python2-pip \
+        nmstate \
         git \
         python2-pyyaml \
```
phoracek: The rest of those dependencies were needed because of the pip installation. They should be isntalled as a part of nmstate RPM now, we can drop them.

phoracek:
ci test please

@qinqon minor comment on the Dockerfile.

It looks like we have a race in our tests: https://jenkins.ovirt.org/job/nmstate_kubernetes-nmstate_standard-check-pr/28//artifact/check-patch.os-3.11.0.el7.x86_64/mock_logs/script/stdout_stderr.log

cmd/state-handler/Dockerfile
qinqon: ack, have drop them now, let's see what funct test says.

qinqon:
@phoracek about the race at our test has fixed it in the linux bridge PR, it takes a time for nodenetworkstate to be there after status-handler is ready, so we have to recheck.

ovirt-infra:
Some tests failed

phoracek:
ci test please

phoracek:
Passed on one platform, so the only problem is the race solved in already merged PR. LGTM

PR #51, opened 2019-05-27
Run a nmstatectl health check daemonset start

### LINKED ISSUE #50: Add minimal NetworkManager health check at daemon set start
during daemon startup, test that nmstatectl is working (returns any interfaces), if not, log that that is the case (user please make sure networkmanager is running) and return 1

qinqon:
To be sure that k8s host nodes has installed and running the needed
pre-requisites.

Closes #50

Error output:

```
Traceback (most recent call last):
  File "/usr/bin/nmstatectl", line 9, in <module>
    load_entry_point('nmstate==0.0.4', 'console_scripts', 'nmstatectl')()
  File "/usr/lib/python2.7/site-packages/nmstatectl/nmstatectl.py", line 54, in main
    return args.func(args)
  File "/usr/lib/python2.7/site-packages/nmstatectl/nmstatectl.py", line 125, in show
    state = _filter_state(netinfo.show(), args.only)
  File "/usr/lib/python2.7/site-packages/libnmstate/netinfo.py", line 35, in show
    report = {Constants.INTERFACES: interfaces()}
  File "/usr/lib/python2.7/site-packages/libnmstate/netinfo.py", line 56, in interfaces
    for dev in nm.device.list_devices()]
  File "/usr/lib/python2.7/site-packages/libnmstate/nm/device.py", line 345, in list_devices
    client = nmclient.client(refresh=True)
  File "/usr/lib/python2.7/site-packages/libnmstate/nm/nmclient.py", line 48, in client
    _nmclient = NM.Client.new(None)
GLib.Error: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Rejected send message, 1 matched rules; type="method_call", sender=":1.543" (uid=0 pid=17904 comm="/usr/bin/python2 /usr/bin/nmstatectl show ") interface="org.freedesktop.DBus.ObjectManager" member="GetManagedObjects" error name="(unset)" requested_reply="0" destination=":1.76" (uid=0 pid=3209 comm="/usr/sbin/NetworkManager --no-daemon ") (9)
E0528 05:56:20.012171       1 main.go:58] 
Error running nmstatectl: exit status 1,  check that NetworkManager is running at nodes

```

### DIFF
--- a/cmd/state-handler/main.go
+++ b/cmd/state-handler/main.go
@@ -54,7 +54,10 @@ func main() {
 			*hostname = envHostname
 		}
 	}
-
+	if err := nmstatectl.Check(); err != nil {
+		klog.Errorf("Error running nmstatectl: %v,  check that NetworkManager is running at nodes\n", err)
+		os.Exit(1)
+	}
 	switch *executionType {
 	case "":
 		panic("execution-type must be specified")
--- a/pkg/nmstatectl/converter.go
+++ b/pkg/nmstatectl/converter.go
@@ -11,6 +11,14 @@ import (
 
 const nmstateCommand = "nmstatectl"
 
+func Check() error {
+	cmd := exec.Command(nmstateCommand, "show")
+	if buff, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("%s\n%v", string(buff), err)
+	}
+	return nil
+}
+
 // Show is populating the passed ConfAndOperationalState object from the output of "nmstatectl show"
 func Show(currentState *v1.ConfAndOperationalState) (err error) {
 	cmd := exec.Command(nmstateCommand, "show", "--json")


cmd/state-handler/main.go
```diff
@@ -54,7 +54,10 @@ func main() {
 			*hostname = envHostname
 		}
 	}
-
+	if err := nmstatectl.Check(); err != nil {
+		klog.Fatalf("Error checking nmstatectl: %v\n", err)
```
phoracek: Please provide a nicer log message, telling we failed to execute nmstatectl, please check that NetworkManager is running on node ...

phoracek: When a user sees the log, they should be able to resolve the problem without reading/searching the traceback.

qinqon: @phoracek Can we check from here processes running at k8s hosts ?

phoracek: @qinqon no, the error message should be hint/suggestion. We don't need to know for sure, just give user the most likely cause.

qinqon: Ok, removed go traceback but keeped nmstatectl python error and added possible check on nodes.

phoracek:
@qinqon sorry, but could you keep the error output in the log? If it contains anything that might lead to fixing the issue, it should be there (not sure if we can set log level, if we can, it should be low severity).

qinqon:
@phoracek keeped the error string from running nmstatectl comand, not sure about log level, this is very terminal error.

cmd/state-handler/main.go
```diff
@@ -54,7 +54,10 @@ func main() {
 			*hostname = envHostname
 		}
 	}
-
+	if err := nmstatectl.Check(); err != nil {
+		klog.Errorf("\nError running nmstatectl: %v,  check that NetworkManager is running at nodes\n", err)
```
SchSeba: why the `/n` at the start?

cmd/state-handler/main.go
```diff
@@ -54,7 +54,10 @@ func main() {
...
+		os.Exit(-1)
```
SchSeba: `-1` is better than `1` ?

qinqon: Fixed and replacing with 1

cmd/state-handler/main.go
qinqon: Fixed

pkg/nmstatectl/converter.go
```diff
@@ -6,11 +6,19 @@ import (
...
 	"os/exec"
 )
 
 const nmstateCommand = "nmstatectl"
 
+func Check() error {
+	cmd := exec.Command(nmstateCommand, "show")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
```
phoracek: AFAIK this returns only error code. We are more interested in stderr, that should be passed as error to the caller. Please correct me if that is wrong. It would be great if you could try it and show the output.

qinqon: Ok, used the CombinedOutput so we wrap the std[out,err] into an error.

PR #54, opened 2019-05-28
Empty operator with the NodeNetworkState API

qinqon:
It just create a busibox pod when you create a NodeNetworkState and             
delete the pod if you delete it, the pod is run in the specified node.          
                                                                                
It also generates nmstatectl daemon set that runs at all nodes just             
sleeping.


Makefile
```diff
@@ -6,35 +6,42 @@ NAMESPACE ?= nmstate-default
...
+	go vet ./cmd/manager
+	CGO_ENABLED=0 GOOS=linux go build \
+		-o $@ ./cmd/manager/
 
 tests/tests.test: $(func_tests_sources) $(apis_sources)
 	ginkgo build tests
 
-docker-state-handler:
-	docker build -f cmd/state-handler/Dockerfile -t $(IMAGE_REGISTRY)/$(STATE_HANDLER_IMAGE):$(IMAGE_TAG) .
+operator: $(operator_bin)
```
phoracek: I would try to avoid calling it operator.

> An Operator is a method of packaging, deploying and managing a Kubernetes application. A Kubernetes application is an application that is both deployed on Kubernetes and managed using the Kubernetes APIs and kubectl tooling.

https://coreos.com/operators/

We however, don't use the controller to deploy the application, our controller is **the** application.

So we need two names, one for the central controller and one for nmstatectl DS. We should not call the central part controller, since it will consist of multiple controllers under a single manager (ouch). Do you have any names in mind? Maybe we can use "manager" for the central part. Then something opposing to manager for the DS. "worker" maybe? Although I'm not sure it would be considered a worker, maybe "agent"?

Makefile
```diff
@@ -6,35 +6,42 @@ NAMESPACE ?= nmstate-default
 IMAGE_REGISTRY ?= quay.io/nmstate
 IMAGE_TAG ?= latest
 PULL_POLICY ?= Always
-STATE_HANDLER_IMAGE ?= kubernetes-nmstate-state-handler
-POLICY_HANDLER_IMAGE ?= kubernetes-nmstate-configuration-policy-handler
+OPERATOR_IMAGE ?= kubernetes-nmstate
+NMSTATECTL_IMAGE ?= nmstatectl
```
phoracek: I don't want to conflict with whatever nmstate/nmstate people are doing. What if we find out that we need some custom kubernetes-nmstate code in the DS part? Then it won't be a generic "nmstatectl" image anymore. I blab more about the naming in following comment.

Makefile
```diff
@@ -6,35 +6,42 @@ NAMESPACE ?= nmstate-default
...
 
 tests/tests.test: $(func_tests_sources) $(apis_sources)
 	ginkgo build tests
 
-docker-state-handler:
-	docker build -f cmd/state-handler/Dockerfile -t $(IMAGE_REGISTRY)/$(STATE_HANDLER_IMAGE):$(IMAGE_TAG) .
+operator: $(operator_bin)
+
+docker-operator: $(operator_bin)
+	docker build -f build/Dockerfile . -t $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
```
phoracek: There are operator-sdk build commands, we could use them. However, we have a second image to build - nmstatectl. What if we change the structure so we have build/operator and build/nmstatectl, to keep it symmetrical. I think it should not conflict with the operator-sdk generate commands.

hack/generate-manifests.sh
```diff
@@ -26,20 +26,24 @@
 set -euo pipefail
```
phoracek: Do we wan't to use our custom script to generate manifests? Operator-sdk has `generate k8s` (or something similar). That would generate CRDs with proper OpenAPI checking and more (RBAC requirements).

Makefile
phoracek: For images:
- kubernetes-nmstate-manager
- kubernetes-nmstate-handler

For objects (on kube API, namespace/name):
- deployment kubernetes-nmstate/manager
- DS kubernetes-nmstate/handler

pkg/controller/controller.go
```diff
@@ -0,0 +1,18 @@
+package controller
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// AddToManagerFuncs is a list of functions to add all Controllers to the Manager
+var AddToManagerFuncs []func(manager.Manager) error
```
phoracek: Not a part of this PR, but once we have it, you can add a controller managing a resource by calling `operator-sdk add controller --api-version=app.example.com/v1alpha1 --kind=AppService`

pkg/apis/addtoscheme_nmstate_v1.go
```diff
@@ -0,0 +1,10 @@
+package apis
+
+import (
+	v1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1"
+)
+
+func init() {
+	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes, v1.SchemeBuilder.AddToScheme)
```
phoracek: It looks like you kept types.go we had there originally, right?

operator-sdk has a lot of metadata, could you remove all we had for now, and then add NodeNetworkState using operator-sdk

`operator-sdk add api --api-version=app.example.com/v1alpha1 --kind=AppService`

pkg/apis/nmstate/v1/nodenetworkstate_types.go
```diff
@@ -357,3 +357,7 @@ const (
 type LinkAggregationMode string
 
 // TODO: add const strings
+
+func init() {
```
phoracek: this Whole file should go away and be created again, with proper review process. operator-sdk also adds some metadata that will be generated AFAIK. Would you mind dropping `pkg/apis` all together for the first PR?

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -0,0 +1,154 @@
+package nodenetworkstate
```
phoracek: Could you please remove the controller from the first PR? Keep only an empty manager, let's add types (API) and controller (this) later in separate PRs

PR #56, opened 2019-05-29
Empty operator no api no controller

qinqon:
operator-sdk new kubernetes-nmstate-manager --dep-manager dep


phoracek:
Please add build/_output to .gitignore, the rest looks fine.

.gitignore
```diff
@@ -0,0 +1,77 @@
+# Temporary Build Files
```
phoracek: Please don't use one of those, it's a mess. IDE specific ignores should be in one's global gitignore. Would you mind dropping the generated part?

PR #59, opened 2019-05-29
add basic sanity checks and travis integration

phoracek:

PR #61, opened 2019-05-29
introduce local cluster

phoracek:
This should be merged only after #59

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,12 @@
-MANAGER_IMAGE ?= kubernetes-nmstate-manager
+IMAGE_REGISTRY ?= quay.io
+IMAGE_REPO ?= nmstate
+MANAGER_IMAGE_NAME ?= kubernetes-nmstate-manager
+MANAGER_IMAGE_TAG ?= latest
+MANAGER_IMAGE_FULL_NAME ?= $(IMAGE_REGISTRY)/$(IMAGE_REPO)/$(MANAGER_IMAGE_NAME):$(MANAGER_IMAGE_TAG)
 
 all: check manager
 
-check: format vet 
+check: format vet
 
 format:
 	gofmt -d cmd/ pkg/
@@ -11,7 +15,22 @@ vet:
 	go vet ./cmd/... ./pkg/...
 
 manager:
-	operator-sdk build $(MANAGER_IMAGE)
+	operator-sdk build $(MANAGER_IMAGE_FULL_NAME)
+
+push-manager:
+	docker push $(MANAGER_IMAGE_FULL_NAME)
+
+cluster-up:
+	./cluster/up.sh
+
+cluster-down:
+	./cluster/down.sh
+
+cluster-sync:
+	./cluster/sync.sh
+
+cluster-clean:
+	./cluster/clean.sh
 
 .PHONY: \
 	all \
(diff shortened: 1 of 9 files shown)


cluster/clean.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+echo 'Cleaning up ...'
+
+./cluster/kubectl.sh delete --ignore-not-found -f deploy/
```
qinqon: no file specified

Makefile
```diff
@@ -0,0 +1,42 @@
+IMAGE_REPO ?= quay.io/nmstate
```
qinqon: Missing IMAGE_REGISTRY so we can push to local registry

```make
IMAGE_REGISTRY ?= quay.io
IMAGE_REPO ?= $(IMAGE_REGISTRY)/nmstate
```

Makefile
```diff
@@ -0,0 +1,42 @@
...
+IMAGE_TAG ?= latest
+MANAGER_IMAGE ?= kubernetes-nmstate-manager
+
```
qinqon: We can compose here the MANAGER_IMAGE_URL to reuse it 
```make
MANAGER_IMAGE_URL ?= $(IMAGE_REPO)/$(MANAGER_IMAGE):$(MANAGER_IMAGE_TAG)
```

Makefile
```diff
@@ -0,0 +1,42 @@
+IMAGE_REPO ?= quay.io/nmstate
+IMAGE_TAG ?= latest
```
qinqon: We have be using different tag for different images

```make
MANAGER_IMAGE_TAG ?= latest
```

Makefile
```diff
@@ -0,0 +1,42 @@
...
+check: check-gofmt check-vet
+
+check-gofmt:
+	./hack/check-gofmt.sh
+
+check-vet:
+	./hack/check-vet.sh
+
+build-manager:
+	operator-sdk build $(IMAGE_REPO)/$(MANAGER_IMAGE):$(IMAGE_TAG)
```
qinqon:
```make
operator-sdk build $(MANAGER_IMAGE_URL)
```

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+push-manager:
+	docker push $(IMAGE_REPO)/$(MANAGER_IMAGE):$(IMAGE_TAG)
```
qinqon:
```make
docker push $(MANAGER_IMAGE_URL)
```

cluster/clean.sh
phoracek: This should deploy everything under the directory.

Makefile
phoracek: Done

Makefile
phoracek: it's not an URL really (misses the protocol prefix), called it MANAGER_IMAGE_FULL_NAME

Makefile
phoracek: Done

Makefile
phoracek: Done

phoracek:
@qinqon works for me. We might want to change the makefile so it has one target to deploy operator on the cluster and one to run it locally. Both dependent on the sync (of RBAC and friends).

PR #63, opened 2019-05-29
run functional tests on std-ci

phoracek:
Merge only after #61


phoracek:
@qinqon looks like it works! Could you please take a look and merge?

PR #66, opened 2019-05-31
Add unit test for State handling

### LINKED ISSUE #68: Run ginkgo command from vendor directory
Right now contributor has to do a go get of ginkgo with correct version, investigate if we can use and installing version from vendor.

qinqon:
We are going to pass nmstatectl yaml as "transparent" data to do so a custom yaml marshaller is implemented that just take it as string, so k8s manifest looks like normal struct but behing the scenes is just a string not a golang struct.

Closes #68 


test/unit/state_test.go
```diff
@@ -0,0 +1,42 @@
...
+  nodeName: node01
+  desiredState:
+    interfaces:
+      - name: eth1
+        type: ethernet
+        state: up
+`
+	)
+	BeforeEach(func() {
+		err := yaml.Unmarshal([]byte(oneIf), &obtained)
```
phoracek: I think this could be done more drastically.

```go
var _ = Describe("State", func() {
	obtained nmstatev1.NodeNetworkState
	err := yaml.Unmarshal([]byte(`
apiVersion: nmstate.io/v1
kind: NodeNetworkState
metadata:
  name: node01
spec:
  managed: true
  nodeName: node01
  desiredState:
    interfaces:
      - name: eth1
        type: ethernet
        state: up
`), &obtained)
	if err != nil {
		panic(err)
	}
```

Or even better, introduce a helper function for that, we will transform manifest string to the structure everywhere.

Also, it should be defined under context. We use this manifest for one specific purpose now. If we later see it can be reuse, we can move it higher. But let's keep it close to its usage to avoid junkyard.

test/unit/state_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"gopkg.in/yaml.v2"
+
+	nmstatev1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1"
+)
+
+var _ = Describe("State", func() {
```
phoracek: NodeNetworkState

test/unit/state_test.go
```diff
@@ -0,0 +1,42 @@
...
+        type: ethernet
+        state: up
+`
+	)
+	BeforeEach(func() {
+		err := yaml.Unmarshal([]byte(oneIf), &obtained)
+		Expect(err).ToNot(HaveOccurred())
+
+	})
+	Context("with one interface", func() {
```
phoracek: when unmarshalled from a manifest with a non-empty desiredState

test/unit/state_test.go
```diff
@@ -0,0 +1,42 @@
...
+		It("should not be nil", func() {
```
phoracek: should not have empty DesiredState field

Or something like that. Be a little more specific, so we are not clueless when it fails.

Gopkg.toml
```diff
@@ -61,6 +61,10 @@ required = [
   # branch = "master" #osdk_branch_annotation
   version = "=v0.8.0" #osdk_version_annotation
 
+[[override]]
```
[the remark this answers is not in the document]
qinqon: Done

test/unit/unit_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package unit
```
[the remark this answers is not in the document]
qinqon: Done

.travis.yml
```diff
@@ -2,9 +2,14 @@ language: go
 go: "1.11"
 
 before_install:
-# install operator-sdk
+- echo "install operator-sdk"
 - sudo curl -JL https://github.com/operator-framework/operator-sdk/releases/download/v0.8.0/operator-sdk-v0.8.0-x86_64-linux-gnu -o /usr/bin/operator-sdk
 - sudo chmod +x /usr/bin/operator-sdk
+- echo "install unit test framework ginkgo and gomega"
+- go get github.com/onsi/ginkgo/ginkgo
```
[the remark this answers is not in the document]
qinqon: Ok, I am going to remove this for now

test/unit/state_test.go
qinqon: I have changed it a lot, let review it again.

test/unit/state_test.go
qinqon: Done

test/unit/state_test.go
qinqon: Let's re review

test/unit/state_test.go
qinqon: let's re review

qinqon:
Humm maybe we have to use k8s API, https://github.com/kubernetes/client-go/issues/193

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+    type: ethernet
+    state: down`)
+
+		NNSManifest = `
+apiVersion: nmstate.io/v1
+kind: NodeNetworkState
+metadata:
+  name: node01
+  creationTimestamp: "1970-01-01T00:00:00Z"
+spec:
```
phoracek: Should be

```yaml
apiVersion: nmstate.io/v1
kind: NodeNetworkState
metadata:
  name: node01
spec:
  managed: true
  nodeName: node01
  desiredState:
    interfaces:
    - name: eth1
      type: ethernet
      state: up
state:
  currentState:
    interfaces:
    - name: eth1
      type: ethernet
      state: down
```

See my comments on types

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+`
+		NNSStruct = NodeNetworkState{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "nmstate.io/v1",
+				Kind:       "NodeNetworkState",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:              "node01",
+				CreationTimestamp: metav1.Unix(0, 0),
```
phoracek: we don't need CreationTimestamp here

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+			Spec: NodeNetworkStateSpec{
+				Managed:      true,
+				NodeName:     "node01",
+				DesiredState: desiredState,
+			},
+			CurrentState: currentState,
+		}
+	)
+
+	BeforeEach(func() {
```
phoracek: why do we need BeforeEach? Can't we just unmarshal and marshal it after definition?

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+			// We cannot compare the whole structs since the raw yaml strings
+			// are not going to match, so we have to match them field by field
+			Expect(obtainedStruct.Spec.DesiredState).
+				To(MatchYAML([]byte(NNSStruct.Spec.DesiredState)))
+
+			Expect(obtainedStruct.CurrentState).
+				To(MatchYAML([]byte(NNSStruct.CurrentState)))
+
+			Expect(obtainedStruct.Spec.Managed).
+				To(Equal(NNSStruct.Spec.Managed))
```
phoracek: could you keep it on the same line? we don't need 79 character lines in Go (or can't really have)

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			Expect(obtainedStruct.Spec.NodeName).
+				To(Equal(NNSStruct.Spec.NodeName))
+
+			Expect(obtainedStruct.TypeMeta).To(Equal(NNSStruct.TypeMeta))
+
```
phoracek: we can drop those empty lines between every check, save some space

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+			Expect(obtainedStruct.ObjectMeta).To(Equal(NNSStruct.ObjectMeta))
+		})
+	})
+
+	Context("when marshal the result", func() {
+		It("should match the NNS manifest", func() {
```
phoracek: please use full name NodeNetworkState

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	yaml "github.com/ghodss/yaml"
+)
+
+var _ = Describe("NodeNetworkState", func() {
+	var (
+		obtainedStruct   NodeNetworkState
```
phoracek: nodeNetworkStateStruct

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+		obtainedManifest []byte
```
phoracek: nodeNetworkStateManifest

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+        type: ethernet
+        state: up
+currentState:
+  interfaces:
+    - name: eth1
+      type: ethernet
+      state: down
+
+`
+		NNSStruct = NodeNetworkState{
```
phoracek: we don't need obtainedStruct intantiation if we have the definition right here

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	BeforeEach(func() {
+		err := yaml.Unmarshal([]byte(NNSManifest), &obtainedStruct)
+		Expect(err).ToNot(HaveOccurred())
+
+		obtainedManifest, err = yaml.Marshal(NNSStruct)
+		Expect(err).ToNot(HaveOccurred())
+
+	})
+	Context("when unmarshal the result", func() {
```
phoracek: when reading NodeNetworkState from a raw manifest

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+		It("should be like the NNS struct", func() {
```
phoracek: should successfully parse desiredState yaml

then other **It**s

should successfully parse currentStata yaml

...

pkg/apis/nmstate/v1/unit_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package v1
```
phoracek: I think it should be named v1_suite_test.go

qinqon:
Let's use what kubernetes is [using](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/yaml/decoder.go#L30)

pkg/apis/nmstate/v1/nodenetworkstate_types.go
```diff
@@ -7,20 +7,18 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// State containes the namestatectl yaml network state
+// +k8s:openapi-gen=true
+type State []byte
+
 // NodeNetworkStateSpec defines the desired state of NodeNetworkState
 // +k8s:openapi-gen=true
 type NodeNetworkStateSpec struct {
```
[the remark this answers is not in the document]
qinqon: You mean "NodeNetworkStateStatus" I removed since it's not giving any value to the data type, is just containe one type and we are not reusing it.

pkg/apis/nmstate/v1/state_helper.go
```diff
@@ -0,0 +1,24 @@
+package v1
```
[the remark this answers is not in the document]
qinqon: Done

pkg/apis/nmstate/v1/nodenetworkstate_types.go
```diff
@@ -7,20 +7,18 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// State containes the namestatectl yaml network state
```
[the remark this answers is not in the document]
qinqon: Done

pkg/apis/nmstate/v1/state_test.go
```diff
@@ -0,0 +1,103 @@
...
+    type: ethernet
+    state: up`)
+
+		currentState = State(`
+interfaces:
+  - name: eth1
+    type: ethernet
+    state: down`)
+
+		NNSManifest = `
```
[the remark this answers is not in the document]
qinqon: Done

pkg/apis/nmstate/v1/state_test.go
qinqon: I did remove state: on purpose we really need it ? commented at your comment.

pkg/apis/nmstate/v1/state_test.go
qinqon: We do, if we don't add it it add a default value the string counterpart depends on also adding this default value, with this we just fix the value.

Maybe we can use ginkgo/gstruct to match partially the struct, but I don't really like how it shows structs when it fails.

pkg/apis/nmstate/v1/state_test.go
qinqon: Splited in two, one for for marshal the other for unmarshal, looks like BeforeEach function get override at Context level (Or that's what I hope)

pkg/apis/nmstate/v1/state_test.go
qinqon: Done

pkg/apis/nmstate/v1/state_test.go
qinqon: Done

pkg/apis/nmstate/v1/state_test.go
qinqon: Done, I like obtained/expected idiom at unit test, but since this is unmarshal it's implicit.

pkg/apis/nmstate/v1/state_test.go
qinqon: Done

pkg/apis/nmstate/v1/state_test.go
qinqon: We use this at both marshal and unmarshal so we need the global definition

pkg/apis/nmstate/v1/state_test.go
qinqon: Done

pkg/apis/nmstate/v1/state_test.go
qinqon: Done

pkg/apis/nmstate/v1/unit_suite_test.go
qinqon: Done. Also I don't know if we want different suite for policy in the future for example, but let's just remove and see what we need later on.

pkg/apis/nmstate/v1/nodenetworkstate_types.go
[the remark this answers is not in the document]
phoracek: `Spec` is like a kubernetes standard. The object should look like:

```shell
apiVersion: nmstate.io/v1
kind: NodeNetworkState
metadata:
  name: node01
spec:
  managed: true
  nodeName: node01
  desiredState:
    interfaces:
    - ...
status:
  currentState:
    capabilities: null
    interfaces:
    - ...
```

Currently there is only currentState in the Spec. However, we will need to extend it later, e.g. with state - whether we were able to apply desiredState, and why not (error message). All bunch of stuff that is currently buried in logs, but should be visible to user.

Makefile
```diff
@@ -20,6 +22,12 @@ manager:
 push-manager:
 	docker push $(MANAGER_IMAGE_FULL_NAME)
 
+gen-k8s:
+	operator-sdk generate k8s
+
+unit-test:
+	ginkgo $(GINKGO_ARGS) ./pkg/apis/nmstate/v1
```
phoracek: as discussed, let's try to replace this with `go test` and drop `go get ` on ginkgo and gomega

qinqon: Looks like ginkgo has nice features to randomize tests, let's keep it go get is just taking the command line tool but I think we are compiling towards vendoring one.

pkg/apis/nmstate/v1/nodenetworkstate_types_test.go
```diff
@@ -0,0 +1,105 @@
...
+			},
+		}
+	)
+
+	Context("when reading NodeNetworkState from a raw manifest", func() {
+
+		var nodeNetworkStateStruct NodeNetworkState
+
+		BeforeEach(func() {
+			err := yaml.Unmarshal([]byte(nnsManifest), &nodeNetworkStateStruct)
```
phoracek: sorry, maybe I'm missing something. Why do we need to Unmarshal (and Marshal in the other test) the manifest before every test, wouldn't be defining it in the scope of `Describe("NodeNetworkState"` enough?

phoracek:
Also, can you please run those tests in verbose by default? I like to see the flow instead of `........`. :)

pkg/apis/nmstate/v1/nodenetworkstate_types_test.go
qinqon: We have to test to different thinks marshal and unmarshal so we have two context for each, and it context to their thing at BeaforeEach.

Makefile
```diff
@@ -37,8 +46,9 @@ cluster-clean:
 	check \
 	format \
 	vet \
-	manager
+	manager \
 	push-manager \
+	test/unit \
```
phoracek: gotta update this one

Makefile
```diff
@@ -20,6 +23,12 @@ manager:
 push-manager:
 	docker push $(MANAGER_IMAGE_FULL_NAME)
 
+gen-k8s:
+	operator-sdk generate k8s
+
+unit-test:
+	$(GINKGO) $(GINKGO_ARGS) ./pkg/apis/nmstate/v1
```
phoracek: whwont you just run it on the whole PKG/?

PR #73, opened 2019-06-07
Add handler

qinqon:
Add nmstate-handler as a proxy to nodes
    
    It's a privileged DaemonSet that will have nmstate library installed
    with this in place nmstate-manager can exec into nmstate-handler pods
    and run nmstatectl to retrieve and set network config.
    
    Added make target "make cluster-sync-handler" to just sync the handler

### DIFF
--- a/build/handler/Dockerfile
+++ b/build/handler/Dockerfile
@@ -1,0 +1,9 @@
+FROM centos:7
+
+RUN yum -y install epel-release && \
+    yum -y update && \
+    yum -y install \
+        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.6/2.el7/noarch/python2-libnmstate-0.0.6-2.el7.noarch.rpm \
+        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.6/2.el7/noarch/nmstate-0.0.6-2.el7.noarch.rpm \
+    yum clean all
+CMD exec /bin/bash -c "trap : TERM INT; sleep infinity & wait"
--- a/deploy/handler.yaml
+++ b/deploy/handler.yaml
@@ -1,0 +1,28 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: nmstate-handler
+spec:
+  selector:
+    matchLabels:
+      nmstate.io: nmstate-handler
+  template:
+    metadata:
+      labels:
+        nmstate.io: nmstate-handler
+    spec:
+      serviceAccountName: kubernetes-nmstate
+      containers:
+        - name: nmstate-handler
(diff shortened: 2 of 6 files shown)


deploy/handler.yaml
```diff
@@ -0,0 +1,40 @@
...
+    spec:
+      containers:
+        - name: nmstate-handler
+          image: REPLACE_IMAGE
+          imagePullPolicy: Always
+          volumeMounts:
+          - name: dbus-socket
+            mountPath: /run/dbus/system_bus_socket
+          env:
+          - name: NODE_NAME
```
phoracek: Do we need any of those env variables?

build/handler/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM centos:7
+RUN yum -y install epel-release && \
+    yum -y update && \
+    yum -y install \
+        https://kojipkgs.fedoraproject.org//packages/nmstate/0.0.5/1.el7/noarch/python2-libnmstate-0.0.5-1.el7.noarch.rpm \
```
phoracek: // should probably be /

deploy/handler.yaml
qinqon: Nope, removed.

build/handler/Dockerfile
qinqon: Done, also upgraded nmstate to 0.0.6

Was entertaining the idea of using ubi7 or even ubi8 but looks like we need subscription-manager to install deps.

deploy/handler.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: extensions/v1beta1
```
phoracek: please try running this with openshift. i think we will need to add scc there

qinqon: So check-patch.os-3.11.0.el7.x86_64 is passing so we are all good ?

phoracek: Not sure, it may get created, but failt to schelute pods. Please trz it locally

qinqon: confirmed that it fails 
        Warning   FailedCreate                      daemonset-controller                 Error creating: pods "nmstate-handler-" is forbidden: unable to validate against any security context constraint: [spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
1s        33s       3         ansible-service-broker.15a725c5e9548158            ClusterServiceBroker

SchSeba: we need to add some SCC config

qinqon: Done

PR #75, opened 2019-06-07
Exec nmstatectl

qinqon:
Update currentState with nmstatectl show from handler

It will connect to nmstate-handler pod and run nmstatectl show command on it
the output is stored at currentState.


SchSeba:
Hi @qinqon nice PR.

I just have a design question here.
should be a better solution if every handler watch is node for change and update is status without involving the manager?

phoracek:
@SchSeba we want to use operator-sdk on maximum. It seems like it is not meant to be used as a daemon set, so we decided to follow this centralized approach. Maybe we will hit some issues, but I don't know about anything unsolvable. Overall, it feels like a cleaner design. One brain, multiple workers. But if you have any specific concerts, please do raise them.

pkg/handler/handler.go
```diff
@@ -0,0 +1,92 @@
...
+		return handler, fmt.Errorf("No nmstate-handler pods at cluster")
+	}
+	for _, pod := range podList.Items {
+		if pod.Spec.NodeName == nodeName {
+			handler.pod = pod
+			return handler, nil
+		}
+	}
+	// Just select the first one in case that more exists
+	return handler, fmt.Errorf("No nmstate handlers at %", nodeName)
```
phoracek: invalid format string

test/e2e/nodes_test.go
```diff
@@ -27,9 +29,22 @@ var _ = Describe("Nodes", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				interfaces := currentState["interfaces"]
+				Expect(interfaces).ToNot(BeEmpty())
+				obtainedInterfaces := interfacesName(interfaces)
+				Expect(obtainedInterfaces).To(SatisfyAll(
+					ContainElement("lo"),
+					ContainElement("cni0"),
+					ContainElement("eth0"),
+					ContainElement("eth1"),
+					ContainElement("flannel.1"),
```
phoracek: please try it on both openshift and kubernetes. i suspect you won't find flannel on openshift

test/e2e/utils.go
```diff
@@ -80,6 +84,16 @@ func writePodsLogs(namespace string, writer io.Writer) error {
 	return nil
 }
 
+func interfacesName(interfaces []map[string]interface{}) []string {
+	var names []string
+	for _, iface := range interfaces {
+		name, hasName := iface["name"]
+		Expect(hasName).To(BeTrue())
```
phoracek: If we fail on this, all we will see is that False is not True in this function. Please add here (and all Expects where it makes sense) message describing why we failed and were we checking

test/e2e/nodes_test.go
```diff
@@ -27,9 +29,22 @@ var _ = Describe("Nodes", func() {
 			for _, node := range nodes {
 				key := types.NamespacedName{Namespace: namespace, Name: node}
 				nodeNetworkState := nodeNetworkState(key)
-				//TODO: exec nmstatectl is not in place let's just compare
-				//      with the stuff we have harcoded there
-				Expect(nodeNetworkState.Status.CurrentState).To(MatchYAML("interfaces: []"))
+				currentStateYaml := nodeNetworkState.Status.CurrentState
+				Expect(currentStateYaml).ToNot(BeEmpty())
```
phoracek: please make the code breathe. empty line after every

```
x := X()
Expect(...)
```

pkg/handler/handler.go
qinqon: Done

test/e2e/nodes_test.go
qinqon: Maybe we can discover were are we running and test according.

test/e2e/utils.go
qinqon: What would be nice is to validate this towards (nmstate schmea)[https://github.com/nmstate/nmstate/blob/master/libnmstate/schemas/operational-state.yaml]

test/e2e/nodes_test.go
qinqon: Done

test/e2e/utils.go
phoracek: I'd suggest open an issue for that. Sounds like a nice little self-contained task.

test/e2e/nodes_test.go
phoracek: I see no benefit in checking for flannel.1. Just check eth0 and eth1 (we are still expecting those tests to be executed against local cluster. Later we can use nmstate to create a set of interfaces in BeforeAll and then verify them.

test/e2e/utils.go
qinqon: Done, but only for test, for the operator this have to be a feature after we generate the config from policies.

https://github.com/nmstate/kubernetes-nmstate/issues/76

test/e2e/nodes_test.go
qinqon: Done.

Makefile
```diff
@@ -90,14 +92,19 @@ cluster-down:
 	./cluster/down.sh
 
 cluster-clean:
-	./cluster/clean.sh
+	$(KUBECTL) delete --ignore-not-found -f build/_output/
+	$(KUBECTL) delete --ignore-not-found -f deploy/
+	$(KUBECTL) delete --ignore-not-found -f deploy/crds/nmstate_v1_nodenetworkstate_crd.yaml
+	if [[ "$$KUBEVIRT_PROVIDER" =~ ^os-.*$$ ]]; then \
```
phoracek: Personally i prefer not to write shell constructs in Makefiles, once we find out we need a little more, it will be an headache. However, I will hope you will be the one that will suffer this and will not fight against this.

automation/test.sh
```diff
@@ -9,6 +9,6 @@ trap '{ make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
 
 make cluster-down
 make cluster-up
-make cluster-sync-handler test/local/e2e
-make cluster-sync
-make test/cluster/e2e
+#TODO: Not working at stdci
```
phoracek: please add WIP or /hold unless we resolve these

test/e2e/utils.go
```diff
@@ -50,6 +54,16 @@ func writePodsLogs(namespace string, writer io.Writer) error {
 	return nil
 }
 
+func interfacesName(interfaces []interface{}) []string {
+	var names []string
+	for _, iface := range interfaces {
+		name, hasName := iface.(map[string]interface{})["name"]
+		Expect(hasName).To(BeTrue(), "should have name field in the interfaces, nhttps://github.com/nmstate/nmstate/blob/master/libnmstate/schemas/operational-state.yam")
```
phoracek: there is trailing "n" before the link, missing l at the end

test/e2e/utils.go
```diff
@@ -33,6 +34,9 @@ func writePodsLogs(namespace string, writer io.Writer) error {
 	podsClientset := framework.Global.KubeClient.CoreV1().Pods(namespace)
 
 	for _, pod := range podList.Items {
+		if !strings.Contains(pod.Name, "nmstate") {
```
phoracek: doesn't seem stable. can we check against some labels instead? maybe we have "app" label set to nmstate

test/e2e/nodes_test.go
```diff
@@ -5,17 +5,36 @@ import (
...
+				var currentStateYaml nmstatev1.State
+				Eventually(func() nmstatev1.State {
+					currentStateYaml = nodeNetworkState(key).Status.CurrentState
+					return currentStateYaml
+				}).ShouldNot(BeEmpty(), "Node %s should have currentState", node)
+
+				By("unmarshal state yaml into unstructured golang")
+				var currentState map[string]interface{}
+				err := yaml.Unmarshal(currentStateYaml, &currentState)
+				Expect(err).ToNot(HaveOccurred(), "should parse correctly yaml: %s", currentStateYaml)
```
phoracek: Other messages start with capital -> Should

automation/test.sh
```diff
@@ -9,6 +9,6 @@ trap '{ make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
 
 make cluster-down
 make cluster-up
-make test/local/e2e
```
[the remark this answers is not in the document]
qinqon: I though local testing was the issue with this PR stdci so I commented it out, but looks like cluster/e2e is failing too, I will keep it if it works after fixing cluster at stdci

deploy/handler.yaml
```diff
@@ -11,7 +11,6 @@ spec:
       labels:
         nmstate.io: nmstate-handler
     spec:
-      serviceAccountName: kubernetes-nmstate
```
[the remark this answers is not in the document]
qinqon: It's related we need it for opensthift since you need a serviceAccount to reference scc

Makefile
qinqon: Agree, the idea is to move things to cluster/ hacks/ when we have a clear view of what we need, but right now is just easier to have it here.

test/e2e/utils.go
qinqon: Done

test/e2e/utils.go
qinqon: Good idea, Done.

test/e2e/nodes_test.go
qinqon: Done

automation/test.sh
qinqon: It's the local testing, we can merge without it since we don't know if we are going to have it at distributed.

qinqon:
Looks like something is not right at automation nmstatectl runs without error but it reports empty interfaces 
```json
{"level":"info","ts":1560406319.1828363,"logger":"handler","msg":"stdout: ---\ndns-resolver:\n  config:\n    search: []\n    server: []\n  running: {}\nroutes:\n  config: []\n  running: []\ninterfaces: []\n, stderr: , err: <nil>"}
{"level":"info","ts":1560406319.1972275,"logger":"controller_nodenetworkstate","msg":"Reconciling NodeNetworkState","Request.Namespace":"default","Request.Name":"node01"}
```

phoracek:
@qinqon could you verify that nmstate is running on the node? What if you run nmstatectl directly, both on the node and in pod?

qinqon:
@phoracek I am trying to verify that network-manager is working correctly in the node, is difficutl since ./cluster/cli.sh ssh cannot be dumped so the logs from there does not appear

I have try to no installa network-manager at up.sh and the result is the same, have to be related to it.

automation/test.sh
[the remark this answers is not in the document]
qinqon: It's working, the issue was the cli.sh I will add it again.

automation/test.sh
qinqon: Going to remove test/local/e2e is working

Makefile
```diff
@@ -90,14 +92,19 @@ cluster-down:
...
+		$(KUBECTL) delete --ignore-not-found -f deploy/openshift/; \
+	fi
 
 cluster-sync-resources:
 	for resource in $(resources); do \
 		$(KUBECTL) apply -f $$resource; \
 	done
 	if [[ "$$KUBEVIRT_PROVIDER" =~ ^os-.*$$ ]]; then \
-		$(KUBECTL) apply -f deploy/openshift-scc.yaml; \
+		$(KUBECTL) apply -f deploy/openshift/scc.yaml; \
```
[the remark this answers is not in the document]
qinqon: Done

test/e2e/utils.go
qinqon: I will try to fix in same commit a do smaller PRs.

qinqon:
@phoracek looks like test/local/e2e is not working on openshift... let's just drop it from automation/test.sh

deploy/handler.yaml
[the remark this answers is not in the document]
qinqon: I know I totally lost track of stuff, I supose I am use to not have commits in the same review (gerrit and all)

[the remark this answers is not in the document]
phoracek: I feel you. The problem is that in GitHub we have only two levels (commits in PRs), while in Gerrit we had three (commits versioned in patches in branches). You can use GitHub PRs either as patches (and then squash), that is nice for reviewing, it is obvious what are latest changes. Or you can use PRs as branches (and then force push commits in those and handle rebases when first commit needs changes). Pick whatever suits you better.

PR #80, opened 2019-06-12
Replace the central operator architecture with a distributed one

### LINKED ISSUE #78: Run per-node
In order to avoid connectivity issues, we should run per node. After we get our current PRs merged, let's follow:

- [x] Remove leader election and run tests again, when executed on a single node, it should still pass
- [x] Share a single container between operator and nmstatectl, drop exec logic and just call it locally
- [x] Add selector reconciling only local node

With this change, we should avoid problems with connectivity lost and execution synchronisation.

qinqon:
Merges manager and handler as daemonset
    
Replaces the central operator architecture with a distributed one
implemented as daemonset, so there is a pod at each node and they
update the NodeNetworkState of their node.

close #78


qinqon:
ci test please

Makefile
```diff
@@ -6,11 +6,6 @@ MANAGER_IMAGE_TAG ?= latest
 MANAGER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(MANAGER_IMAGE_NAME):$(MANAGER_IMAGE_TAG)
 MANAGER_IMAGE ?= $(IMAGE_REGISTRY)/$(MANAGER_IMAGE_FULL_NAME)
 
-HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
```
phoracek: we agreed to keep only handler, not manager (because of it sounds centralized)

qinqon: Is PR is draft, no ready for review, my idea is to rename it to handler, it was easier to use it as manager since the name was ar more places.

pkg/controller/node/node_controller.go
```diff
@@ -101,50 +102,17 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
 
 	//TODO: Manage deletes
-	nodeNetworkStateKey := types.NamespacedName{
-		Namespace: "default",
-		Name:      request.Name,
-	}
-	// Create NodeNetworkState for this node
 	nodeNetworkState := &nmstatev1.NodeNetworkState{}
-	err = r.client.Get(context.TODO(), nodeNetworkStateKey, nodeNetworkState)
+	err = nmstate.GetNodeNetworkState(r.client, request.Name, nodeNetworkState)
```
phoracek: node_controller should initialize with empty state. node_network_status_controller should populate it. Let's make sure we split those responsibilities. Later, when we have Conditions implemented, we can mark it here as Initialized=True, but Populated=False. Or something similar.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -36,8 +41,22 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	forThisPod := predicate.Funcs{
```
phoracek: forThisNode.

Do we care about Delete? There is nothing we can do, is it? It would probably fail on Update.

pkg/helper/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nmstatev1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1"
+)
+
+const nmstateCommand = "nmstatectl"
+const namespace = "default"
```
phoracek: Isn't NodeNetworkState cluster-wide? You don't need to specify Namespace name when handling cluster-wide resources.

pkg/controller/node/node_controller.go
```diff
@@ -45,8 +43,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// If they are not set by default they are true so we set others than
-	// CreateFunc to false.
-	onCreation := predicate.Funcs{
+	// CreateFunc to false
```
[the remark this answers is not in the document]
qinqon: Done

pkg/controller/node/node_controller.go
qinqon: Changing currentState to empty means calling "Update" on client and the reponsable of this is node_network_status_controller, do we really need to put empty state here (In case is not already empty).

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
qinqon: Maybe just bypass them, since the only one delete state if node controller and the node is going down.

pkg/helper/client.go
qinqon: It's not working if I don't use namespace.

pkg/controller/node/node_controller.go
phoracek: we don't change anything here, we are creating initial NodeNetworkState. there is no NodeNetworkState to update, we trigger initial creation. the rest will be done by other component. sounds like a good distributed system approach. maybe i just can't see your concern, could you elaborate little more?

pkg/helper/client.go
phoracek: Aha, we are not cluster-wide yet: https://github.com/nmstate/kubernetes-nmstate/blob/operators-wip/deploy/crds/nmstate_v1_nodenetworkstate_crd.yaml#L12. Nevermind then, we have to fix that in another PR

pkg/controller/node/node_controller.go
qinqon: So do we have to change something here or not ?

pkg/helper/client.go
qinqon: Damn it next PR will be before starting with desiredState.

pkg/controller/node/node_controller.go
phoracek: - You are reconciling a created Node here
- You try to find existing NodeNetworkState on 106
- That always fails on 108, we are the ones creating it initially later in this function

We should only call CreateNodeNetworkState(), you can drop nodeNetworkState parameter, since it will be always empty while creating.

qinqon: Better logic would be to re-create the object if it's there to cover corner case, what do you think ?

pkg/helper/node.go
```diff
@@ -0,0 +1,14 @@
+package helper
+
+import (
+	"os"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func IsForThisPod(meta v1.Object) bool {
```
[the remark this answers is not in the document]
qinqon: Done.

pkg/controller/node/node_controller.go
phoracek: I'm fine with smarter logic, like somebody removed NodeNetworkState? We will recreate if needed. However, it has to be an empty config (in this controller).

qinqon: Done

pkg/helper/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+func nmstatectl(arguments ...string) (string, error) {
+	//cmd := exec.Command(nmstateCommand, arguments...)
```
SchSeba: nit: remove comment

pkg/helper/client.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating NodeNetworkState: %v, %+v", err, nodeNetworkState)
+	}
+
+	return nil
+}
+
+func DeleteNodeNetworkState(client client.Client, nodeNetworkState *nmstatev1.NodeNetworkState) error {
+
+	// There is no NodeNetworkState for this node let's create it
```
SchSeba: wrong comment?

pkg/helper/client.go
```diff
@@ -0,0 +1,89 @@
...
+	cmd.Stdout = &outb
+	cmd.Stderr = &errb
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("Failed to execute nmstatectl show: '%v'", err)
+	}
+	return outb.String(), nil
+}
+
+func GetNodeNetworkState(client client.Client, nodeName string, nodeNetworkState *nmstatev1.NodeNetworkState) error {
+
```
SchSeba: nit: remove empty line

pkg/helper/client.go
```diff
@@ -0,0 +1,89 @@
...
+	nodeNetworkStateKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      nodeName,
+	}
+
+	return client.Get(context.TODO(), nodeNetworkStateKey, nodeNetworkState)
+}
+
+func CreateNodeNeworkState(client client.Client, nodeName string) error {
+
```
SchSeba: nit: remove empty line

pkg/helper/client.go
```diff
@@ -0,0 +1,89 @@
...
+	err := client.Create(context.TODO(), &nodeNetworkState)
+	if err != nil {
+		return fmt.Errorf("error creating NodeNetworkState: %v, %+v", err, nodeNetworkState)
+	}
+
+	return nil
+}
+
+func DeleteNodeNetworkState(client client.Client, nodeNetworkState *nmstatev1.NodeNetworkState) error {
+
```
SchSeba: nit: remove empty line

pkg/helper/node.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"os"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Return true if the event name is the name of
+// the pods's node (reading the env var NODE_NAME)
+func EventIsForThisNode(meta v1.Object) bool {
```
SchSeba: can we save the node that we are running as part of the controller? no need to check this on every request this is static for all the pod running life time.

test/e2e/utils.go
```diff
@@ -117,3 +154,12 @@ func nodeNetworkState(key types.NamespacedName) nmstatev1.NodeNetworkState {
 	}, timeout, interval).ShouldNot(HaveOccurred())
 	return state
 }
+
+func deleteNodeNeworkStates() {
+	nodeNetworkStateList := &nmstatev1.NodeNetworkStateList{}
+	err := framework.Global.Client.List(context.TODO(), &dynclient.ListOptions{}, nodeNetworkStateList)
+	Expect(err).ToNot(HaveOccurred())
+	for _, nodeNetworkState := range nodeNetworkStateList.Items {
+		framework.Global.Client.Delete(context.TODO(), &nodeNetworkState)
```
SchSeba: I think we should also get the error here and configure a `Expect(err).ToNot(HaveOccurred())` because this can change the tests if we run the `AfterEach` and we are not able to remove all the objects before the next test.

deploy/operator.yaml
```diff
@@ -1,21 +1,20 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
```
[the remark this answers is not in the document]
qinqon: Let's stick with operator-sdk conventions.

build/Dockerfile
```diff
@@ -1,12 +1,19 @@
-FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
+FROM centos:7
 
-# Cannot change the binary to handler-manager since the name
+RUN yum -y install epel-release && \
+    yum -y update && \
+    yum -y install \
+        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.6/2.el7/noarch/python2-libnmstate-0.0.6-2.el7.noarch.rpm \
```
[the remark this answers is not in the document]
qinqon: Latest 0.0.5 RPM has an issue as solution they (nmstate team) give us a direct koji build for 0.0.6.

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/node.go
qinqon: Is about reading environment variables let's not add complexity in the controller.

test/e2e/utils.go
qinqon: We want cleanup to continue, I will try to store a flag and check after loop.

deploy/operator.yaml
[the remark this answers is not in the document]
phoracek: I think that even as a daemonset, we are still an operator for OKD4 standards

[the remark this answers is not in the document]
phoracek: I think that even as a daemonset, we are still an operator for OKD4 standards

qinqon:
ci test please

phoracek:
```
default                             nmstate-handler-5mg86                          0/1       ContainerCreating   0          52s
```

Looks like nothing is failing and it is just a timeout issue waiting for the container to be created. However, 52 seconds it quite long for a container that has image available locally.

qinqon:
@phoracek At my laptop is does not take that much :-(

pkg/controller/node/node_controller.go
```diff
@@ -101,50 +102,23 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
+			return reconcile.Result{}, fmt.Errorf("error at node reconcile accessing NodeNetworkState: %v", err)
 		}
-		nodeNetworkState.Spec = nmstatev1.NodeNetworkStateSpec{
-			NodeName: request.Name,
-		}
-		// There is no NodeNetworkState for this node let's create it
-		err = r.client.Create(context.TODO(), nodeNetworkState)
+	} else {
+		// NodeNetworkState is no suppose to be there let's delete it
+		err = nmstate.DeleteNodeNetworkState(r.client, nodeNetworkState)
```
phoracek: Maybe someone wants to define NodeNetworkState configuration and only then register a Node? I think we should just ignore this case until we hit an issue with it.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -36,8 +41,22 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return nmstate.EventIsForThisNode(createEvent.Meta)
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			return nmstate.EventIsForThisNode(deleteEvent.Meta)
+		},
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
+			return nmstate.EventIsForThisNode(updateEvent.MetaNew)
+		},
+		GenericFunc: func(genericEvent event.GenericEvent) bool {
```
phoracek: what is this GenericFunc? Do we care about it?

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -36,8 +41,22 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	forThisNode := predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return nmstate.EventIsForThisNode(createEvent.Meta)
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			return nmstate.EventIsForThisNode(deleteEvent.Meta)
```
phoracek: What about ignoring it for now? There is no point on updating content of a NodeNetworkState that has been marked to be removed

pkg/helper/client.go
```diff
@@ -0,0 +1,84 @@
...
+	var outb, errb bytes.Buffer
+	cmd.Stdout = &outb
+	cmd.Stderr = &errb
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("Failed to execute nmstatectl show: '%v'", err)
+	}
+	return outb.String(), nil
+}
+
+func GetNodeNetworkState(client client.Client, nodeName string, nodeNetworkState *nmstatev1.NodeNetworkState) error {
```
phoracek: Any benefit in populating pointer over returning the instance?

pkg/helper/client.go
```diff
@@ -0,0 +1,84 @@
...
+	nodeNetworkStateKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      nodeName,
+	}
+
+	return client.Get(context.TODO(), nodeNetworkStateKey, nodeNetworkState)
+}
+
+func CreateNodeNeworkState(client client.Client, nodeName string) error {
```
phoracek: Maybe InitializeNodeNetworkState to emphasize that we are creating it blank

pkg/helper/client.go
```diff
@@ -0,0 +1,84 @@
...
+	nodeNetworkState := nmstatev1.NodeNetworkState{}
```
phoracek: why don't you populate meta and spec right here?

pkg/helper/client.go
```diff
@@ -0,0 +1,84 @@
...
+	// Create NodeNetworkState for this node
+	nodeNetworkState.ObjectMeta = metav1.ObjectMeta{
+		Name:      nodeName,
+		Namespace: namespace,
+	}
+	nodeNetworkState.Spec = nmstatev1.NodeNetworkStateSpec{
+		NodeName: nodeName,
+	}
+	// There is no NodeNetworkState for this node let's create it
```
phoracek: I don't see this logic in here. It is known to a caller of this function, but not to the function itself. For all we know, it may exist

pkg/helper/client.go
```diff
@@ -0,0 +1,84 @@
...
+	err := client.Create(context.TODO(), &nodeNetworkState)
+	if err != nil {
+		return fmt.Errorf("error creating NodeNetworkState: %v, %+v", err, nodeNetworkState)
+	}
+
+	return nil
+}
+
+func DeleteNodeNetworkState(client client.Client, nodeNetworkState *nmstatev1.NodeNetworkState) error {
+	err := client.Delete(context.TODO(), nodeNetworkState)
```
phoracek: Maybe it would be better to only pass Node name for these?

test/e2e/utils.go
```diff
@@ -76,11 +79,45 @@ func prepare(t *testing.T) (*framework.TestCtx, string) {
...
+func WaitForOperatorDaemonSet(t *testing.T, kubeclient kubernetes.Interface, namespace, name string, retryInterval, timeout time.Duration) error {
+	return waitForDaemonSet(t, kubeclient, namespace, name, retryInterval, timeout, true)
+}
+
+func waitForDaemonSet(t *testing.T, kubeclient kubernetes.Interface, namespace, name string, retryInterval, timeout time.Duration, isOperator bool) error {
+	if isOperator && framework.Global.LocalOperator {
+		t.Log("Operator is running locally; skip waitForDeployment")
+		return nil
+	}
+	err := wait.PollImmediate(retryInterval, timeout, func() (done bool, err error) {
```
phoracek: What about using Gomega's Eventually? Then you just implement checkDaemonSet (to read it and check whether it's ready. The rest will be done using Eventually.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
qinqon: It's for events that are unknown, maybe refresh is one of them.

pkg/controller/node/node_controller.go
qinqon: Have removed in the refresh PR but let's do it here too.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
qinqon: Done.

pkg/helper/client.go
qinqon: Nope here let's simplify it.

pkg/helper/client.go
qinqon: Make sense, done.

pkg/helper/client.go
qinqon: Done.

pkg/helper/client.go
qinqon: Going to delete this comment, is from the copy paste.

pkg/helper/client.go
qinqon: Well I am going to remove this function al together since we are ditching the logit that use it (node reconcile with already existing nodenetworkstate)

test/e2e/utils.go
qinqon: This is copy paste from operators-sdk to wait for Deployments, my idea is to create a PR there in their framework with this, so I prefer to keep it like their code.

phoracek:
@SchSeba could you take another look (and approve+merge)?

qinqon:
@phoracek @SchSeba Let me fix conflicts with the kubevirtci PR

PR #84, opened 2019-06-14
Run automation/test.sh with no colors

qinqon:
There is no color support at jenkins and the output is kind of messy,
with this patch we maintain color at local dev but remove it at
jenkins run.


phoracek:
OS fail is unrelated. (no) Colors are fine

PR #85, opened 2019-06-14
Add more info at stdci teardown

qinqon:
It will list k8s pods and events before cluster-down

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -2,10 +2,16 @@
 
 kubectl() { cluster/kubectl.sh "$@"; }
 
+teardown() {
+    kubectl get event || true
+    kubectl get pod || true
+    make cluster-down
+}
+
 export KUBEVIRT_PROVIDER=$TARGET
 
 # Make sure that the VM is properly shut down on exit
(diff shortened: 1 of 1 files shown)


automation/test.sh
```diff
@@ -2,10 +2,16 @@
 
 kubectl() { cluster/kubectl.sh "$@"; }
 
+teardown() {
+    kubectl get event || true
+    kubectl get pod || true
```
phoracek: don't we need to specify namespace?

automation/test.sh
```diff
@@ -2,10 +2,16 @@
 
 kubectl() { cluster/kubectl.sh "$@"; }
 
+teardown() {
+    kubectl get --all-namespaces event || true
+    kubectl get --all-namespaces pod || true
```
phoracek: this would be madness on openshift, too many pods. are we interested in anything but nmstate namespace?

qinqon: Let's have it all, you never know what you need and this is just after finishing it.

phoracek: (*demonic voice* you will regret it JAJAJA)

ack

PR #88, opened 2019-06-14
Use cluster scope for NodeNetworkState

qinqon:
Followed instructions at [1]

https://github.com/operator-framework/operator-sdk/blob/master/doc/operator-scope.md

Signed-off-by: Quique Llorente <quique.llorente@gmail.com>


deploy/role.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
phoracek: We have to be more strict here.

Please use cluster-role only where we needed (AFAIK monitoring Nodes and handling NodeNetworkState).

The rest should be limited to our namespace (we should not have rights to remove all existing objects).

Finally, I'm sure we don't need all the resource that are listed here, but that is for another PR.

qinqon: Done, let's do cleanup after we merge to master with all the changes so we know what we access.

qinqon: It's taking me time to put the correct rules at the two places.

qinqon: I needed to put "pods" and "pods/exec" at ClusterRole, maybe after the distributed handler we can change that, at least now is more restrictive.

phoracek: weird. but it won't be needed in the final version with shared containre between nmstatectl and operator, so it's fine.

phoracek:
ci test please

PR #90, opened 2019-06-17
Refresh NodeNetworkState every 5 seconds

qinqon:
This PR include the following functionalit:
- re-create NodeNetworkState in case it's deleted
- refresh currentState every 5 seconds from nmstatectl show
- filter NodeNetworkState updates by generation and finalizers so we don't reconcile on status updates.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	log         = logf.Log.WithName("controller_nodenetworkstate")
+	log         = logf.Log.WithName("controller_node")
 	nodeRefresh = 5 * time.Second
 )
 
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -3,6 +3,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -20,7 +21,10 @@ import (
 	nmstate "github.com/nmstate/kubernetes-nmstate/pkg/helper"
 )
 
-var log = logf.Log.WithName("controller_node")
+var (
+	log         = logf.Log.WithName("controller_nodenetworkstate")
+	nodeRefresh = 5 * time.Second
+)
 
 // Add creates a new Node Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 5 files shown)


test/e2e/nodes_test.go
```diff
@@ -1,49 +1,63 @@
...
 			}
 		})
+		Context("and node network state is deleted", func() {
+			BeforeEach(func() {
+				deleteNodeNeworkStates()
+			})
+			It("should recreate it with currentState", func() {
+				for _, node := range nodes {
+					var currentStateYaml nmstatev1.State
+					currentState(namespace, node, &currentStateYaml).ShouldNot(BeEmpty())
```
phoracek: could you do the ShouldNot check inside currentState and just take return value with State here?

pkg/controller/node/node_controller.go
```diff
@@ -20,7 +21,10 @@ import (
 	nmstate "github.com/nmstate/kubernetes-nmstate/pkg/helper"
 )
 
-var log = logf.Log.WithName("controller_node")
+var (
+	log         = logf.Log.WithName("controller_nodenetworkstate")
```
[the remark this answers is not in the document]
qinqon: Done

pkg/helper/client.go
```diff
@@ -64,7 +64,7 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1.NodeNe
 		CurrentState: nmstatev1.State(currentState),
 	}
 
-	err = client.Status().Update(context.TODO(), nodeNetworkState)
+	err = client.Status().Update(context.Background(), nodeNetworkState)
```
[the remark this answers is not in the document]
qinqon: I have follow [best practices](https://blog.openshift.com/kubernetes-operators-best-practices/) here about update status.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -51,7 +56,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			return false
 		},
 		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
-			return nmstate.EventIsForThisNode(updateEvent.MetaNew)
+			return nmstate.EventIsForThisNode(updateEvent.MetaNew) &&
+				(updateEvent.MetaNew.GetGeneration() != updateEvent.MetaOld.GetGeneration() ||
```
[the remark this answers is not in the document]
qinqon: Done

test/e2e/nodes_test.go
qinqon: The idea is to be the test the one that put the assert, we may want to do currentState(..).Should(BeEmpty())

phoracek: What if you move Expect outer:

Expect(currentState).Should... ?

qinqon: What do we gain with that ?

phoracek: Ah sorry, misunderstood. Well, if you think this makes more sense, let's keep it.

PR #95, opened 2019-06-19
Apply DesiredState to nodes using nmstatectl set

qinqon:
So it can change network config at all the nodes, like creating
linux bridge or bond.


pkg/helper/client.go
```diff
@@ -15,15 +16,34 @@ import (
...
+	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
-		return "", fmt.Errorf("Failed to execute nmstatectl show: '%v'", err)
+		return "", fmt.Errorf("Failed to execute nmstatectl show: '%v', %s", err, stderr.String())
 	}
-	return outb.String(), nil
+	return stdout.String(), nil
+}
+
+func set(state string) error {
```
phoracek: I think we should capture stdout (or whatever contains log) output and log it in the pod. I think it is a valuable information, especially now when we may face a lot of issues with config.

pkg/helper/client.go
```diff
@@ -15,15 +16,34 @@ import (
...
-	cmd.Stdout = &outb
-	cmd.Stderr = &errb
+func show(arguments ...string) (string, error) {
+	cmd := exec.Command(nmstateCommand, "show")
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
-		return "", fmt.Errorf("Failed to execute nmstatectl show: '%v'", err)
+		return "", fmt.Errorf("Failed to execute nmstatectl show: '%v', %s", err, stderr.String())
```
phoracek: errors should start lower-case. Would you mind fixing it here while we edit these?

pkg/helper/client.go
```diff
@@ -15,15 +16,34 @@ import (
...
+	return stdout.String(), nil
+}
+
+func set(state string) error {
+	cmd := exec.Command(nmstateCommand, "set")
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("Failed to create pipe for writing  into nmstate: %v", err)
```
phoracek: Double space between writing and into

pkg/helper/client.go
```diff
@@ -15,15 +16,34 @@ import (
...
+	}
+	defer stdin.Close()
```
phoracek: please put an empty line after this

pkg/helper/client.go
```diff
@@ -15,15 +16,34 @@ import (
...
+	_, err = io.WriteString(stdin, state)
+	if err != nil {
+		return fmt.Errorf("Failed to write state into stdin: %v", err)
+	}
```
phoracek: and after this

pkg/helper/client.go
```diff
@@ -15,15 +16,34 @@ import (
...
+	if err = cmd.Run(); err != nil {
+		return fmt.Errorf("Failed to execute nmstate set: '%v' '%s'", err, stderr.String())
+	}
```
phoracek: and this

test/e2e/nodenetworkstate_test.go
```diff
@@ -0,0 +1,56 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstatev1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1"
+)
+
+var _ = Describe("NodeNetworkState", func() {
+	Context("when desiredState is configured", func() {
+		var (
+			desiredState nmstatev1.State
```
phoracek: please do it on a single line, save trees

test/e2e/nodenetworkstate_test.go
```diff
@@ -0,0 +1,56 @@
...
+		)
+		JustBeforeEach(func() {
+			for _, node := range nodes {
+				updateDesiredState(namespace, node, desiredState)
```
phoracek: This is confusing. Combining operations performed in BeforeEach and JustBeforeEach gives me headache, jumping up and down. Cannot we just do all operations in BeforeEach directly? You can create a helper function to config desired state everywhere.

You could create onEachNode(func) helper that you could use in BeforeEach and It

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

pkg/helper/client.go

test/e2e/nodenetworkstate_test.go
qinqon: Done

test/e2e/nodenetworkstate_test.go
qinqon: I was doing this ginkgo way, so we put the scenarios at BeforeEach, JustBeforeEach do just the stuff to updateDesiredState BeforeEach set the specific config, but we can replace it by helper.

phoracek: I don't like when I need to think too much while reading tests from top to bottom, but if it is **the** Ginkgo way, so be it.

test/e2e/nodenetworkstate_test.go
```diff
@@ -0,0 +1,69 @@
...
+	Context("when desiredState is configured", func() {
+
+		var desiredState nmstatev1.State
+
+		JustBeforeEach(func() {
+			for _, node := range nodes {
+				updateDesiredState(namespace, node, desiredState)
+			}
+		})
+		JustAfterEach(func() {
```
phoracek: This feels weird to me. Could you help Petr and tell him why is that better than having just BeforeEach and AfterEach?

qinqon: I am going to replace it by helper functions I don't like either and I hate myself for it.

phoracek: :open_mouth:

test/e2e/nodenetworkstate_test.go
```diff
@@ -42,14 +28,19 @@ var _ = Describe("NodeNetworkState", func() {
...
     type: linux-bridge
     state: absent
 `)
+			)
+
+			BeforeEach(func() {
+				updateDesiredState(namespace, br1Up)
+			})
+			AfterEach(func() {
+				updateDesiredState(namespace, br1Absent)
```
phoracek: We have to be blocking, don't we?

qinqon: Well we have to check that the bridge is really not there, so is safe to continue.

phoracek: In this PR?

qinqon: Let's remove the bridge directly in the node so we don't use the stuff we are testing, and we test bridge deletion at different thest.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -38,6 +38,10 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	return &ReconcileNodeNetworkState{client: mgr.GetClient(), scheme: mgr.GetScheme()}
 }
 
+func desiredState(object runtime.Object) nmstatev1.State {
+	return object.(*nmstatev1.NodeNetworkState).Spec.DesiredState
```
phoracek: We should make this bit safer. It is code in controller that might panic if object is not the expected type (unlikely, but trust nobody, maybe it's fine now, but we might forget, use the function in another riskier context and fail hard). Would you please try to cast the type using case and return error in case it doesn't match?

phoracek:
@qinqon please rebase

PR #96, opened 2019-06-19
Set k8s-1.13.3 as the default provider

qinqon:
The kubevirtci uses k8s-1.13.3 as the default let's do the same
at this project so ./kubevirtci/cluster-up/kubectl.sh works
out of the box.


SchSeba:
@qinqon can we remove the 1.11 lane? to reduce some resources on every PR we create for the STD-CI?

qinqon:
@SchSeba yep totally, let's remove it, we can add it later on if needed.

PR #97, opened 2019-06-20
Install kubevirtci from different organization

qinqon:
Also move to a script the code to install it.


hack/install-kubevirtci.sh
```diff
@@ -0,0 +1,17 @@
...
+
+script_dir=$(dirname "$(readlink -f "$0")")
+kubevirtci_dir=kubevirtci
+
+rm -rf $kubevirtci_dir
+git clone https://github.com/$organization/kubevirtci $kubevirtci_dir
+pushd $kubevirtci_dir
+git checkout $commit
+popd
+
```
phoracek: please drop those trailing lines

Makefile
```diff
@@ -80,10 +81,8 @@ $(local_handler_manifest): deploy/operator.yaml
 	sed "s#REPLACE_IMAGE#$(LOCAL_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)#" \
 		deploy/operator.yaml > $@
 
-$(CLUSTER_DIR)/%: kubevirtci.version
-	rm -rf kubevirtci
-	git clone https://github.com/kubevirt/kubevirtci
-	cd kubevirtci && git checkout $$(cat  ../kubevirtci.version)
+$(CLUSTER_DIR)/%: $(install_kubevirtci)
```
phoracek: this should depend on kubevirtci as well, in case it is missing

test/e2e/utils.go
```diff
@@ -172,20 +173,21 @@ func run(node string, command ...string) {
...
+	GinkgoWriter.Write([]byte(strings.Join(ssh_command, " ") + "\n"))
 	cmd.Stdout = GinkgoWriter
 	cmd.Stderr = GinkgoWriter
 	Expect(cmd.Run()).To(Succeed())
 }
 
 func createDummy(nodes []string, dummyName string) {
 	for _, node := range nodes {
-		run(node, "sudo", "ip", "link", "add", dummyName, "type", "dummy")
+		run(node, "sudo", "nmcli", "con", "add", "type", "dummy", "con-name", dummyName, "ifname", dummyName)
```
phoracek: how is this related to the PR?

qinqon: stdci was failing at "ip link add" with "RTNETLINK answers: File exists" but the interface was not there, looks like it's relatd to flushing the interface, doing it with nmcli works and also kind of make sense since we are testing networkmanager.

So this is all to make CI happy.

phoracek:
Do we need this now when the PR has been merged?

phoracek:
I like the script, but we should drop the mention of different repo. Just pin it to specific commit on kubevirtci.

qinqon:
We are not going to be able to see commits of different branches outside of main organization, let's keep this as it is and point to the merged stuff.

We may need a long time deviation on a kubevirtci wip, this is still needed.

Makefile
qinqon: kubevirtci is part of the CLUSTER_DIR if it's not there make will call the dependencies.

CLUSTER_DIR = kubevirtci/cluster-up/

hack/install-kubevirtci.sh
qinqon: Done

phoracek:
@qinqon do you want to wait for the multi-node fix?

qinqon:
@phoracek let's merge this will fix cli.sh already and we are not testing multinode now.

PR #118, opened 2019-06-26
Add 'make release' target

qinqon:
It uses do a docker push tagged with version/version.go and uses
github-release command [1] to create release with manifests and
info from version/name and version/description.

[1] https://github.com/aktau/github-release


phoracek:
ci test please

Makefile
```diff
@@ -2,8 +2,9 @@ IMAGE_REGISTRY ?= quay.io
 IMAGE_REPO ?= nmstate
 
 HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
-HANDLER_IMAGE_TAG ?= latest
-HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME):$(HANDLER_IMAGE_TAG)
+#HANDLER_IMAGE_SUFFIX ?= :latest
```
phoracek: why do we keep this and why the latest has `:`?

Makefile
```diff
@@ -14,6 +15,7 @@ E2E_TEST_EXTRA_ARGS ?=
 E2E_TEST_ARGS ?= $(strip -test.v -ginkgo.v $(E2E_TEST_EXTRA_ARGS))
 
 OPERATOR_SDK ?= build/_output/bin/operator-sdk
+GITHUB_RELEASE ?= build/_output/bin/github-release
```
phoracek: there are several groups of variable definitions, why is the one with OPERATOR_SDK and LOCAL_REGISTRY the best fit for GITHUB_RELEASE? I think we are starting to have a junkyard of variables here. It's getting out of hand, we should organize and comment them properly, in a separate PR.

Makefile
```diff
@@ -26,10 +28,12 @@ KUBECTL ?= $(CLUSTER_DIR)/kubectl.sh
 CLUSTER_UP ?= $(CLUSTER_DIR)/up.sh
 CLUSTER_DOWN ?= $(CLUSTER_DIR)/down.sh
 CLI ?= $(CLUSTER_DIR)/cli.sh
-SSH ?= $(CLI) ssh
+SSH ?= $(CLUSTER_DIR)/ssh.sh
```
phoracek: How is this change related?

Makefile
```diff
@@ -49,6 +53,9 @@ $(GINKGO): Gopkg.toml
 $(OPERATOR_SDK): Gopkg.toml
 	GOBIN=$$(pwd)/build/_output/bin/ go install ./vendor/github.com/operator-framework/operator-sdk/cmd/operator-sdk
 
+$(GITHUB_RELEASE): Gopkg.toml
+	GOBIN=$$(pwd)/build/_output/bin/ go install ./vendor/github.com/aktau/github-release
```
phoracek: Can we pin github-release to a specific release? This doesn't feel safe

Makefile
```diff
@@ -68,12 +75,24 @@ test/e2e: $(OPERATOR_SDK)
...
 
-
 $(local_handler_manifest): deploy/operator.yaml
 	mkdir -p $(dir $@)
 	sed "s#REPLACE_IMAGE#$(LOCAL_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)#" \
 		deploy/operator.yaml > $@
 
+
+$(versioned_operator_manifest): HANDLER_IMAGE_SUFFIX = :$(shell hack/version.sh)
+$(versioned_operator_manifest): version/version.go
```
phoracek: what happens when you have multiple targets with the same name?

Makefile
```diff
@@ -68,12 +75,24 @@ test/e2e: $(OPERATOR_SDK)
...
+	mkdir -p $(dir $@)
+	sed "s#REPLACE_IMAGE#$(HANDLER_IMAGE)#" \
+		deploy/operator.yaml > $@
+
+$(): deploy/operator.yaml
```
phoracek: what is this?

Makefile
```diff
@@ -100,9 +119,8 @@ cluster-sync-resources: $(KUBECTL)
 		$(KUBECTL) apply -f deploy/openshift/; \
 	fi
 
-cluster-sync-handler: cluster-sync-resources $(local_handler_manifest)
-	IMAGE_REGISTRY=localhost:$(shell $(CLI) ports registry | tr -d '\r') \
-		make push-handler
+cluster-sync-handler: IMAGE_REGISTRY = localhost:$(shell $(CLI) ports registry | tr -d '\r')
```
phoracek: Please move unrelated changes out. It is hard to review this PR for me since there are so many changes I'm not so confident reviewing.

Makefile
```diff
@@ -112,6 +130,20 @@ cluster-sync-handler: cluster-sync-resources $(local_handler_manifest)
 
 cluster-sync: cluster-sync-handler
 
+$(description): version/description
+	mkdir -p $(dir $@)
+	sed "s#HANDLER_IMAGE#$(HANDLER_IMAGE)#" \
+		version/description > $@
+
+release: HANDLER_IMAGE_SUFFIX = :$(shell hack/version.sh)
```
phoracek: Please explain how this whole mechanism works. Having all the logic in Makefile without any comments is unmaintaintable and unreviewable.

version/description
```diff
@@ -0,0 +1,11 @@
+This release has rewrite the project from scratch using operator-sdk
```
phoracek: I want to see documentation of "how to do a release" somewhere in the project.

Makefile
qinqon: Reading around, looks like tagging with latest is not good practices, you have to just do no tag and push and you can still get latest, that's why this HANDLER_IMAGE_SUFFIX is added here.

But the commented one is trash, let me remove it.

Makefile
qinqon: Yep will organize, in a other PR.

Makefile
qinqon: Little boy scout here, but I can change it back.

Makefile
qinqon: Done

Makefile
qinqon: It's not multiple target this is just adding a Make variable to the target, so for exampe if I change this and one of the dependencie is handler-push it will use the changed HANDLER_IMAGE_SUFFIX

This way we don't have to call "make handler-push" but just add handler-push as a dependency.

Makefile
qinqon: Brain fart, it was the local version of the operator.yaml, removed by accident.

Makefile
qinqon: Done

Makefile
qinqon: Will do a bigger PR with documentation on it.

version/description
qinqon: Done

PR #119, opened 2019-06-27
Add bond e2e test

qinqon:
The test case is taken from cnv-test use of bond interfaces


test/e2e/nodenetworkstate_test.go
```diff
@@ -30,8 +42,41 @@ var _ = Describe("NodeNetworkState", func() {
...
+        stp:
+          enabled: false
+      port:
+        - name: bond1
+          stp-hairpin-mode: false
+          stp-path-cost: 100
+          stp-priority: 32
+`)
 
+		br1Absent = nmstatev1alpha1.State(`interfaces:
```
phoracek: what about absent bonds? do we cleanup connections on host (is there leftover of deactivated connection)?

qinqon: I leave it on purpose, as it does exactly same thing as bridges just link del.

phoracek: i don't understand.

why do we have it for bridge and not for bond?

do we have leftover connections on node? if so, it affects the test. in real situation, we would have no pre-existing connection. but in this test, we reuse the same over and over again (wild guess)

test/e2e/nodenetworkstate_test.go
```diff
@@ -75,10 +116,81 @@ var _ = Describe("NodeNetworkState", func() {
...
+							HaveKeyWithValue("type", expectedBond["type"]),
+							HaveKeyWithValue("state", expectedBond["state"]),
+							HaveKeyWithValue("link-aggregation", expectedBond["link-aggregation"]),
+						)),
+						ContainElement(SatisfyAll(
+							HaveKeyWithValue("name", expectedBridge["name"]),
+							HaveKeyWithValue("type", expectedBridge["type"]),
+							HaveKeyWithValue("state", expectedBridge["state"]),
+							HaveKeyWithValue("bridge", HaveKeyWithValue("port", expectedBridgePorts)),
+						))))
```
phoracek: lisp vibes

PR #121, opened 2019-06-27
Add basic NodeNetworkConfigurationPolicy API

qinqon:
It expose the following:
- spec.matcher.nodes: nodes to apply the policy
- spec.priority: order in witch config is merged before apply to same node
- spec.desiredState: nmstate configuration to apply to the node


deploy/crds/nmstate_v1alpha1_nodenetworkstate_crd.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     listKind: NodeNetworkStateList
     plural: nodenetworkstates
     singular: nodenetworkstate
-  scope: Cluster
+  scope: Namespaced
```
phoracek: ?

qinqon:
ci test please

deploy/crds/nmstate_v1alpha1_nodenetworkstate_crd.yaml
```diff
@@ -28,8 +28,23 @@ spec:
         metadata:
           type: object
         spec:
+          properties:
```
phoracek: This PR should not do any changes on NodeNetworkState.

phoracek: Ah, it needs to be rebased once #122 is merged. nvm

deploy/crds/nmstate_v1alpha1_nodenetworkconfigurationpolicy_crd.yaml
```diff
@@ -0,0 +1,56 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: nodenetworkconfigurationpolicies.nmstate.io
+spec:
+  group: nmstate.io
+  names:
+    kind: NodeNetworkConfigurationPolicy
+    listKind: NodeNetworkConfigurationPolicyList
+    plural: nodenetworkconfigurationpolicies
+    singular: nodenetworkconfigurationpolicy
```
phoracek: can we keep a short here as well? nnsp?

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -0,0 +1,74 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
phoracek: please remove all these noisy comments

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -0,0 +1,74 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NodeNetworkConfigurationPolicyMatch define the matching criteria to apply
+// the policy
+// +k8s:openapi-gen=true
+type NodeNetworkConfigurationPolicyMatch struct {
```
phoracek: in the original design doc, Match was only for interface matching, nodeSelector was under Spec directly. I'd be in favor of that, to avoid too much nesting.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -0,0 +1,74 @@
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Node names to apply the policy
+	Nodes []string `json:"nodes,omitempty"`
```
phoracek: Let's stick to existing kubernetes API and use nodeSelector: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector

It does matching based on labels, so it is pretty flexible.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -0,0 +1,74 @@
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// In case of multiple policies applying for the same node this
+	// priority define the order from low to high
+	Priority int `json:"priority,omitempty"`
+
+	// Criteria to apply this policy
+	Match NodeNetworkConfigurationPolicyMatch `json:"match,omitempty"`
```
phoracek: We can drop it from this initial version. Just keep nodeSelector on this level.

deploy/crds/nmstate_v1alpha1_nodenetworkconfigurationpolicy_crd.yaml
qinqon: nncp, Done.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
qinqon: Done

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
qinqon: Like k8s https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L2878

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
qinqon: ditto

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
phoracek: Don't understand your point here. nodeSelector is a map that should live in the top scope of Spec. `NodeNetworkConfigurationPolicyMatch` you define here is not needed in this PR, `match` should be later used to match interfaces running on a node.

deploy/crds/nmstate_v1alpha1_nodenetworkconfigurationpolicy_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: nmstate.io/v1alpha1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: example-nodenetworkconfigurationpolicy
+spec:
+  # Add fields here
+  size: 3
```
phoracek: please provide a real example matching what we have available in the CRD

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -0,0 +1,64 @@
...
+
+// NodeNetworkConfigurationPolicySpec defines the desired state of NodeNetworkConfigurationPolicy
+// +k8s:openapi-gen=true
+type NodeNetworkConfigurationPolicySpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// In case of multiple policies applying for the same node this
+	// priority define the order from low to high
```
phoracek: from high to low, right? or the highest to the lowest. we take priority 100 before 1.

qinqon: From the design document 

> The implementation may process NodeNetworkConfigurationPolicy entries based on priority (from low to high), such that each entry may override a previous value in the desired state. It is left up to the implementation to define such behaviour, especially in cases where the priority on an entry is equal with other “colliding” entries.

So it's says we apply policies from low to high priority, so lastest applied will override (or this is what I understand), so 100 will override stuff at 1.

phoracek: Aha! Sorry, that makes sense.

phoracek:
ci test please

PR #122, opened 2019-06-27
Generate the openapi validation code

qinqon:
Also add a make gen-openapi


pkg/apis/nmstate/v1alpha1/nodenetworkstate_types.go
```diff
@@ -43,6 +45,8 @@ type NodeNetworkState struct {
 	Status NodeNetworkStateStatus `json:"status,omitempty"`
 }
 
+// +genclient
+// +genclient:nonNamespaced
```
phoracek: Not needed for List, only for single object

PR #123, opened 2019-06-28
Basic policy controller

qinqon:
It should update desiredState of all nodes in case of the
match node list is empty or the specified one.
    
If more than one policy applys merge desiredState using
priority flag.


Makefile
```diff
@@ -6,8 +6,14 @@ HANDLER_IMAGE_TAG ?= latest
...
 
-GINKGO_EXTRA_ARGS ?=
-GINKGO_ARGS ?= -v -r --randomizeAllSpecs --randomizeSuites --race --trace $(GINKGO_EXTRA_ARGS)
+UNIT_TEST_EXTRA_ARGS ?=
+UNIT_TEST_ARGS ?= -v -r --randomizeAllSpecs --randomizeSuites --race --trace $(UNIT_TEST_EXTRA_ARGS)
+ifdef UNIT_TEST_FOCUS
+	UNIT_TEST_ARGS += --focus $(UNIT_TEST_FOCUS)
+endif
+ifdef UNIT_TEST_SKIP
+	UNIT_TEST_ARGS += --skip $(UNIT_TEST_SKIP)
```
phoracek: this look familiar, except in the other code i saw, you had `+={space space}--...`, so there are spaces between flags

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,177 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var (
+	log = logf.Log.WithName("controller_nodenetworkconfigurationpolicy")
+)
+
+/**
+* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
```
phoracek: please drop this

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,177 @@
+package nodenetworkconfigurationpolicy
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
```
phoracek: split group with nmstate internal imports and third party

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,177 @@
...
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	return &ReconcileNodeNetworkConfigurationPolicy{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+}
+func getNodeName() (string, error) {
```
phoracek: empty space above

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,177 @@
...
+	nodeName := os.Getenv("NODE_NAME")
```
phoracek: could we keep it in `init()` so it is loaded only once? it is fine if we panic when missing

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,177 @@
...
+func matches(nodeSelector map[string]string, labels map[string]string) bool {
+	for key, value := range nodeSelector {
+		if foundValue, hasKey := labels[key]; !hasKey || foundValue != value {
+			return false
+		}
+	}
+	return true
+}
+
+func nodeSelectorMatchisThisNode(cl client.Client, eventObject runtime.Object) bool {
```
phoracek: Matches

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,272 @@
+package nodenetworkconfigurationpolicy
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	. "github.com/onsi/ginkgo/extensions/table"
+
```
phoracek: drop this line

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var _ = Describe("NodeNetworkConfigurationPolicy controller predicates", func() {
+	type PredicateCase struct {
```
phoracek: why is this public?

Makefile
qinqon: Organized it a little, let me check if is good now.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: This is an early optimization, not needed.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: No only that. It makes sure we die immediately during the start if the mandatory variable is missing (there is no recover from that). With this code we loop on error log (don't we?)

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,174 @@
...
+		return false
+	}
+	node := corev1.Node{}
+	err = cl.Get(context.TODO(), types.NamespacedName{Name: nodeName}, &node)
+	if err != nil {
+		log.Info("Cannot find corev1.Node", "nodeName", nodeName)
+		return false
+	}
+
+	policyNodeSelector := eventObject.(*nmstatev1alpha1.NodeNetworkConfigurationPolicy).Spec.NodeSelector
```
phoracek: node selector is not mandatory, is it? This could fail (or the following, passing nil as selector)

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,174 @@
...
+		return err
+	}
+
+	// Watch for changes to primary resource NodeNetworkConfigurationPolicy
+	err = c.Watch(&source.Kind{Type: &nmstatev1alpha1.NodeNetworkConfigurationPolicy{}}, &handler.EnqueueRequestForObject{}, forThisNodePredicate(mgr.GetClient()))
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
phoracek: forgotten error checking?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,174 @@
...
+type ReconcileNodeNetworkConfigurationPolicy struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a NodeNetworkConfigurationPolicy object and makes changes based on the state read
+// and what is in the NodeNetworkConfigurationPolicy.Spec
+// TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
```
phoracek: drop this

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
```
phoracek: you are adding import, but no user

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+var _ = Describe("NodeNetworkConfigurationPolicy controller predicates", func() {
+	type predicateCase struct {
+		EnvNodeName  string
+		ObjNodeName  string
+		NodeSelector map[string]string
+		NodeLabels   map[string]string
+		Reconcile    bool
+	}
+	DescribeTable("all events",
```
phoracek: maybe "testing predicates"? it gives more context to the output IMHO

ovirt-infra:
Some tests failed

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Ohh let's go to init + panic :-(

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: it's not a pointer they cannot pass nil there, if they don't configure it it will be an empty map I think

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
qinqon: This is old I think

ovirt-infra:
Some tests failed

ovirt-infra:
Some tests failed

qinqon:
test ci please

Makefile
phoracek: If I call focus and skip now, does it work? Will it include an empty space between the two?

Makefile
```diff
@@ -21,13 +28,15 @@ ifdef E2E_TEST_EXTRA_ARGS
 	E2E_TEST_ARGS +=  $(E2E_TEST_EXTRA_ARGS)
 endif
 
+GINKGO?= build/_output/bin/ginkgo
```
phoracek: `GINKGO ?=`

Makefile
```diff
@@ -21,13 +28,15 @@ ifdef E2E_TEST_EXTRA_ARGS
...
 OPERATOR_SDK ?= build/_output/bin/operator-sdk
 GITHUB_RELEASE ?= build/_output/bin/github-release
 LOCAL_REGISTRY ?= registry:5000
 
 export KUBEVIRT_PROVIDER ?= k8s-1.13.3
 export KUBEVIRT_NUM_NODES ?= 1
 export KUBEVIRT_NUM_SECONDARY_NICS ?= 1
+export NODE_NAME=node01
```
phoracek: where do we use it? which Node's name is this? first node? node used for some purpose? what about multi-node?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var (
+	log      = logf.Log.WithName("controller_nodenetworkconfigurationpolicy")
+	nodeName string
+)
+
+func init() {
+	var isSet = false
```
phoracek: this is not needed, you just 

```
nodeName, isSet := os.LookupEnv("NODE_NAME")
```

on the next line

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -0,0 +1,160 @@
...
+			reqLogger.Info(fmt.Sprintf("the NodeNetworkState for %s is not there yet, let's requeue", nodeName))
+			// If there is no NodeNetworkState let's requeue could be that
+			// we are in the middle of the creation
+			return reconcile.Result{Requeue: true}, nil
+		}
+		// Error reading the nodeNetworkState - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// FIXME: We have to merge it somehow in case of multiple policies applied
```
phoracek: please add TODO for Patch operation instead of Update. Otherwise, we will end up without observedState and conditions for some time.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_suite_test.go
```diff
@@ -0,0 +1,19 @@
...
+	. "github.com/onsi/gomega"
+)
+
+func TestUnit(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "NodeNetworkConfigurationPolicy controller Test Suite")
+}
+
+var _ = BeforeSuite(func() {
+	os.Setenv("NODE_NAME", "node01")
```
phoracek: Does this apply when we do it from init?

Makefile
qinqon: Yep it will include a space " += " has one, I will fix the GINKGO part though

Makefile
qinqon: This is just used just by unit tests, let me move there.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: no, this is wrong, is you do this it create a private nodeName and it fails.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_suite_test.go
qinqon: I have to remove this damn, good catch.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: -_- ok

qinqon: Is the same ?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Ok, I will add a check there, the idea is that if they don't specify nodeSelector all nodes will be selected, but let's put that as a unit test.

qinqon: Looks like map is never nil
https://play.golang.org/p/gPZB49dQUzF

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,242 @@
...
+			}),
+	)
+})
+
+var _ = Describe("NodeNetworkConfigurationPolicy controller reconciler", func() {
+	var (
+		name          = "nodenetworkconfigurationpolicy-test"
+		namespace     = "default"
+		numberOfNodes = 2
+		nodes         []nmstatev1alpha1.NodeNetworkState
```
phoracek: nodeNetworkStates?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+		cl      client.Client
+		r       *ReconcileNodeNetworkConfigurationPolicy
+		req     reconcile.Request
+		policy1 = nmstatev1alpha1.NodeNetworkConfigurationPolicy{
```
phoracek: policy1? what's that

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,242 @@
...
+				Namespace: namespace,
+			},
+		}
+		res reconcile.Result
+	)
+
+	BeforeEach(func() {
+
+		By("populate the NodeNetworkState for nodes")
+		nodes = nil
```
phoracek: wasm't it a list?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,242 @@
...
+		for n := 1; n <= numberOfNodes; n++ {
+			nodes = append(nodes, nmstatev1alpha1.NodeNetworkState{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("node%02d", n),
+					Namespace: namespace,
```
phoracek: state should not be namespaced

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+	})
+
+	JustBeforeEach(func() {
+
+		// Register operator types with the runtime scheme.
+		By("register state and policies types")
+		s := scheme.Scheme
+		s.AddKnownTypes(nmstatev1alpha1.SchemeGroupVersion, &policy1)
+		s.AddKnownTypes(nmstatev1alpha1.SchemeGroupVersion, &nodes[0])
```
phoracek: you should just initialize and pass the struct here. do not assume the list has expexted type and is not empty

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,242 @@
...
+		// Objects to track in the fake client
+		objs := []runtime.Object{&policy1}
+
+		By("create fake client")
+		cl = fake.NewFakeClient(objs...)
+
+		r = &ReconcileNodeNetworkConfigurationPolicy{client: cl, scheme: s}
+
+	})
+	Context("when there is no NodeNetworkState for the node", func() {
```
phoracek: please add empty lines between cotexts, its, describes

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: This is the policy that we reconcile.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: we don't care append will initialize it.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Removed namespace everywhere for CRs

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
phoracek: Then it should be reconciledPolicy (or something like that). Also, it shoudn't be a global variable - you need to make sure it's clean before ever other test.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,247 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: name,
+			},
+		}
+		res reconcile.Result
+	)
+
+	BeforeEach(func() {
+
+		By("populate the NodeNetworkState for nodeNetworkStates")
```
phoracek: populate usually starts with higher-case, it's not part of the Describe-Conext-It chain.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+		}
+	})
+
+	JustBeforeEach(func() {
+
+		// Register operator types with the runtime scheme.
+		By("register state and policies types")
+		s := scheme.Scheme
+		s.AddKnownTypes(nmstatev1alpha1.SchemeGroupVersion, &nmstatev1alpha1.NodeNetworkConfigurationPolicy{})
```
phoracek: Could we do this in suite initialization? It is too much init code that will be needed in most of the tests.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,247 @@
...
+		s.AddKnownTypes(nmstatev1alpha1.SchemeGroupVersion, &nmstatev1alpha1.NodeNetworkState{})
+
+		// Objects to track in the fake client
+		objs := []runtime.Object{&policy1}
+
+		By("create fake client")
+		cl = fake.NewFakeClient(objs...)
+
+		r = &ReconcileNodeNetworkConfigurationPolicy{client: cl, scheme: s}
+
```
phoracek: Feel free to argue, but Petr's favorite style is:

```
Foo
if err != nil {
	...
}

Bar
if err != nil {
	...
}

Foo({
	bar
	bar

	baz
})

Shrubbery
```

Keep blocks on the same level separated by empty line, keep error handling in the block where the error was returned, if starting a nested block, no need to divide it with empty line.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,247 @@
...
+				NodeLabels:   map[string]string{},
+				NodeSelector: map[string]string{},
+				Reconcile:    false,
+			}),
+	)
+})
+
+var _ = Describe("NodeNetworkConfigurationPolicy controller reconciler", func() {
+	var (
+		name              = "nodenetworkconfigurationpolicy-test"
```
phoracek: name and number of nodes can be constants.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -0,0 +1,247 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			res, err = r.Reconcile(req)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should not requeue", func() {
+			Expect(res.Requeue).ToNot(BeTrue())
+		})
+
+		Context(" and it has non empty desiredState", func() {
```
phoracek: no need for the initial space

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Only the schema part, it's better to have new client every time we run tests, but I don't see much gain moving it to suite init.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
qinqon: Done

PR #128, opened 2019-07-10
Load nns refresh time from config map

### LINKED ISSUE #12: add controller configuration for retry and polling rate
should be done via a configmap + env variables to be statically set on startup

HlinaCZ:
/cc @phoracek

Resolves #12

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -41,6 +41,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: NODE_NETWORK_STATE_REFRESH_INTERVAL
+              valueFrom:
+                configMapKeyRef:
+                  name: nmstate-config
+                  key: node_network_state_refresh_interval
           volumeMounts:
           - name: dbus-socket
             mountPath: /run/dbus/system_bus_socket
@@ -51,3 +56,11 @@ spec:
         hostPath:
           path: /run/dbus/system_bus_socket
           type: Socket
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nmstate-config
+  namespace: nmstate
+data:
+  node_network_state_refresh_interval: "5"
--- a/pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
+++ b/pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
@@ -3,7 +3,9 @@ package nodenetworkstate
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
+	"strconv"
 	"time"
 
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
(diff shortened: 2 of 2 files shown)


ovirt-infra:

deploy/operator.yaml
```diff
@@ -51,3 +56,11 @@ spec:
...
           path: /run/dbus/system_bus_socket
           type: Socket
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nmstate-config
+  namespace: nmstate
+data:
+  node_network_state_refresh_interval: "5"
```
phoracek: please add missing newline

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -24,9 +26,21 @@ import (
 
 var (
 	log                     = logf.Log.WithName("controller_nodenetworkstate")
-	nodenetworkstateRefresh = 5 * time.Second
+	nodenetworkstateRefresh time.Duration
 )
 
+func init() {
```
qinqon: Do this at main and pass to constructed controller, this is very implicit.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -24,9 +26,21 @@ import (
...
+	refreshTime, isSet := os.LookupEnv("NODE_NETWORK_STATE_REFRESH_INTERVAL")
+	if !isSet {
+		panic("NODE_NETWORK_STATE_REFRESH_INTERVAL is mandatory")
+	}
+	intRefreshTime, err := strconv.Atoi(refreshTime)
```
qinqon: Can't we verify that at configmap type ?

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
phoracek: It is not clear how would we pass any extra data from main() down. Also, it would require us to either create a singleton for it or pass it through several functions all the way down here. I'd personally prefer init() in this case. But if you think passing it down would be better (totally fair, makes sense to keep single interface between app and the environment, i won't fight for init()).

phoracek:
ci add to whitelist

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
phoracek: ConfigMap values are only string.

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
qinqon: Puff I see a lot of dependency injections let's just do at init (but I still don't like them)

phoracek:
>  buy I hate panic and init

@qinqon it is not a runtime error, if it happens, it happens right at the beginning because of something immutable (env variable passed via manifest) is missing/invalid. I think we are on the safe side in this case.

PR #129, opened 2019-07-10
Multiple policies

qinqon:
Deal with multiple policies
    
If controller is reconciling one policie it should combine its
desiredState with other ones that matches nodeSelector, the
combination logic is just combine is there isn't any conflict
at 'interfaces' field.

### DIFF
--- a/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
@@ -115,8 +115,8 @@ type ReconcileNodeNetworkConfigurationPolicy struct {
 	scheme *runtime.Scheme
 }
 
-// It will return the policies with same labels at the one from argument
-func (r *ReconcileNodeNetworkConfigurationPolicy) filterByNodeLabels(node corev1.Node) (nmstatev1alpha1.NodeNetworkConfigurationPolicyList, error) {
+// It will return policies with their nodeSelector matching node labels
+func (r *ReconcileNodeNetworkConfigurationPolicy) policiesMatchingNode(node corev1.Node) (nmstatev1alpha1.NodeNetworkConfigurationPolicyList, error) {
 	policyList := nmstatev1alpha1.NodeNetworkConfigurationPolicyList{}
 	filteredPolicyList := nmstatev1alpha1.NodeNetworkConfigurationPolicyList{}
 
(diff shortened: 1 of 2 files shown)


ovirt-infra:
Some tests failed

ovirt-infra:
Some tests failed

qinqon:
test ci please

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -113,6 +115,107 @@ type ReconcileNodeNetworkConfigurationPolicy struct {
 	scheme *runtime.Scheme
 }
 
+// It will return the policies with same labels at the one from argument
```
phoracek: You meant to say policies matching the same node/labels? Maybe it would be better to make it policiesMatchingNode.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -113,6 +115,107 @@ type ReconcileNodeNetworkConfigurationPolicy struct {
...
+		if nodeSelectorMatchesThisNode(r.client, &policy) {
+			filteredPolicyList.Items = append(filteredPolicyList.Items, policy)
+		}
+	}
+
+	return filteredPolicyList, nil
+}
+
+func unmarshalInterfaces(state nmstatev1alpha1.State) (map[string]interface{}, []interface{}, error) {
+	// Unmarshall interfaces state into unstructured golang
```
phoracek: This should document the whole function. Also, please document what is being returned in the first and the second value. It looks weird that we unwrap insides and return both unstructuredState and its interfaces. Can't we handle it from the caller?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -113,6 +115,107 @@ type ReconcileNodeNetworkConfigurationPolicy struct {
...
+	var unstructuredState map[string]interface{}
+	err := yaml.Unmarshal(state, &unstructuredState)
+	if err != nil {
+		return unstructuredState, []interface{}{}, fmt.Errorf("error unmarshaling state: %v", err)
+	}
+	return unstructuredState, unstructuredState["interfaces"].([]interface{}), nil
+}
+
+func mapByName(interfacesList []interface{}) (map[string]interface{}, error) {
```
phoracek: having "interface" as in network interface and "interface" as golang thing on one line is killing me... Can we specify a new interface networkInterface/iface to make it easier for poor Petr's eyes? Or rename all variables to be ifaceList etc.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -113,6 +115,107 @@ type ReconcileNodeNetworkConfigurationPolicy struct {
...
+	intersectionKeys := []string{}
+	for key, _ := range lhs {
+		if _, hasKey := rhs[key]; hasKey {
+			intersectionKeys = append(intersectionKeys, key)
+		}
+	}
+	return intersectionKeys
+}
+
+// It will merge "interfaces" if they are not comflicting (there is not changes from the same interface) in case
```
phoracek: conflicting

phoracek:
Multiple policies will be applied without merging

PR #135, opened 2019-07-15
filter out veths

### LINKED ISSUE #109: Filter out veth interfaces from status
The observed state currently lists all interfaces of the host. With Kubernetes, that can be even around hundred veth interfaces connecting Pods to cluster network. I'd say those interfaces are not as important for node networking and we should filter them out by default (user can request to list them by setting a flag on NodeNetworkState).

HlinaCZ:
This function takes argument from config map and removes uninteresting outputs for customers by filtering it out.

Resolves #109


pkg/helper/client.go
```diff
@@ -5,17 +5,26 @@ import (
...
+	"os"
 	"os/exec"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/gobwas/glob"
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	yaml "gopkg.in/yaml.v2"
```
qinqon: use the k8s yaml parser yaml "sigs.k8s.io/yaml"

pkg/helper/client.go
```diff
@@ -79,14 +88,16 @@ func InitializeNodeNeworkState(client client.Client, nodeName string) error {
...
+	currentStateRaw, err := show()
 	if err != nil {
 		return fmt.Errorf("error running nmstatectl show: %v", err)
 	}
 
-	// Let's update status with current network config from nmstatectl
-	nodeNetworkState.Status = nmstatev1alpha1.NodeNetworkStateStatus{
-		CurrentState: nmstatev1alpha1.State(currentState),
+	currentState := nmstatev1alpha1.State(currentStateRaw)
+	if interfacesFilter == "" {
```
qinqon: Move this check to filterOut since it's part of the filter logic.

pkg/helper/client.go
```diff
@@ -104,3 +115,30 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 	}
 	return set(string(nodeNetworkState.Spec.DesiredState))
 }
+
+func filterOut(currentState nmstatev1alpha1.State, interfacesFilter string) nmstatev1alpha1.State {
+	var state map[string]interface{}
```
qinqon: Add here the empty check
if interfacesFilter == "" {
    return currentState
}

pkg/helper/client.go
```diff
@@ -104,3 +115,30 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 	}
 	return set(string(nodeNetworkState.Spec.DesiredState))
 }
+
+func filterOut(currentState nmstatev1alpha1.State, interfacesFilter string) nmstatev1alpha1.State {
```
qinqon: Return error

pkg/helper/client.go
```diff
@@ -104,3 +115,30 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+	var state map[string]interface{}
+	err := yaml.Unmarshal([]byte(currentState), &state)
+	if err != nil {
+		panic(fmt.Sprintf("%v", err))
```
qinqon: Don't panic return error.

pkg/helper/client.go
```diff
@@ -79,14 +88,16 @@ func InitializeNodeNeworkState(client client.Client, nodeName string) error {
...
 	}
 
-	// Let's update status with current network config from nmstatectl
-	nodeNetworkState.Status = nmstatev1alpha1.NodeNetworkStateStatus{
-		CurrentState: nmstatev1alpha1.State(currentState),
+	currentState := nmstatev1alpha1.State(currentStateRaw)
+	if interfacesFilter == "" {
+		nodeNetworkState.Status.CurrentState = currentState
+	} else {
+		nodeNetworkState.Status.CurrentState = filterOut(currentState, interfacesFilter)
```
qinqon: Handle error

pkg/helper/client.go
```diff
@@ -5,17 +5,26 @@ import (
...
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/gobwas/glob"
```
qinqon: Why don't we use golang regex ? 
https://gobyexample.com/regular-expressions

pkg/helper/client.go
```diff
@@ -104,3 +115,30 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+		g := glob.MustCompile(interfacesFilter)
+		if !g.Match(name.(string)) {
+			filteredInterfaces = append(filteredInterfaces, iface)
+		}
+	}
+
+	state["interfaces"] = filteredInterfaces
+	filteredState, err := yaml.Marshal(state)
+	if err != nil {
+		panic(fmt.Sprintf("%v", err))
```
qinqon: Return error don't panic

pkg/helper/client_test.go
```diff
@@ -0,0 +1,53 @@
...
+	Context("when given 2 interfaces and 1 is veth", func() {
+		state := nmstatev1alpha1.State(`interfaces:
+- name: eth1
+  state: up
+  type: ethernet
+- name: vethab6030bd
+  state: down
+  type: ethernet
+`)
+		returnedState := filterOut(state, "veth*")
```
qinqon: use var blocks and BeforeEach is the gomega way to isolate specs.

pkg/helper/client_test.go
```diff
@@ -0,0 +1,53 @@
...
+		It("should return filtered 1 interface without veth", func() {
+			filteredState := nmstatev1alpha1.State(`interfaces:
+- name: eth1
+  state: up
+  type: ethernet
+`)
+			Expect(returnedState).To(Equal(filteredState))
+		})
```
qinqon: Since we are going to move "" empty scenario to filterOut add a unit test to it.

And since we are going to return errors add a bad yaml format one too.

pkg/helper/client.go
phoracek: IMO regular expressions are overkill in this case. They are hard to read and write. Do you see any case where regex would be needed? All I see now is filtering out stuff with name prefix.

qinqon: For this case we don't need complicated regex veth.* so there is no need for another third party that user has to learn.

veth.* vs veth* this is not that different

qinqon: Also what we are filtering here is an arbitrary name not url or path, so better go with regex than globing.

qinqon: Globing is usually used to filter stuff like github.com/kubevirt/* or the same for some path, while regex is mor suitable for names for example, a user want to filter a range of interfaces for whatever reason you cannot do it with globing, but you can do it with regex eth[0-5] or the like.

qinqon: Bah let's go with gobling, is good enough, if we need more complex scenarios we can add a fix

phoracek: You had a fair point though. I don't have a strong opinion on this. @HlinaCZ won't be around next weeks, so we have a time to think this through.

pkg/helper/client_test.go
```diff
@@ -0,0 +1,99 @@
+package helper
```
SchSeba: please add a test that we add another interface in the filter config.

check that we can filter both veth and vnet.

pkg/helper/client.go
```diff
@@ -136,3 +147,34 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+	if err != nil {
+		return currentState, err
+	}
+
+	interfaces := state["interfaces"]
+	var filteredInterfaces []interface{}
+
+	for _, iface := range interfaces.([]interface{}) {
+		name := iface.(map[string]interface{})["name"]
+		g := glob.MustCompile(interfacesFilter)
```
SchSeba: maybe move this to the init funtion and save it? we don't need to do it on every call.

The `MustCompile` function comment
```
MustCompile is the same as Compile, except that if Compile returns error, this will panic
```

This will exit I don't think we want that.

HlinaCZ: If I move this `MustCompile` to init function we would couldn't use unit test for testing this feature as it would filter names only from environment variable or configuration file. 
Do you have any better idea how to solve this issue?

SchSeba: maybe I don't understand the issue but I think you can create a function for this and you call the function in the init function.

for the unit test just test the function you create that is the same you call from the init

pkg/helper/client.go
```diff
@@ -148,8 +147,18 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 	return setOutput + outputVlanFiltering, nil
 }
 
-func filterOut(currentState nmstatev1alpha1.State, interfacesFilter string) (nmstatev1alpha1.State, error) {
-	if interfacesFilter == "" {
+func getFilter() {
```
phoracek: This should return pointer to the compiled filter. User must not know that a singleton is being used.

pkg/helper/client.go
```diff
@@ -148,8 +147,18 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+		interfacesFilter := os.Getenv("INTERFACES_FILTER")
+		globCompiled = glob.MustCompile(interfacesFilter)
+		globCompiledIsSet = true
+	}
+}
+
+func filterOut(currentState nmstatev1alpha1.State) (nmstatev1alpha1.State, error) {
+	getFilter()
+
+	if globCompiled.Match("") {
```
phoracek: We should compare with glob.Nothing (or what was it). That should be less expensive (if it works)

pkg/helper/client.go
```diff
@@ -19,11 +19,10 @@ import (
 
 const nmstateCommand = "nmstatectl"
 
-var interfacesFilter string
-
-func init() {
-	interfacesFilter = os.Getenv("INTERFACES_FILTER")
-}
+var (
+	globCompiled      glob.Glob
```
phoracek: Please give it a better name describing its use, not only what it is. interfacesFilterGlob maybe? Up to you.

pkg/helper/client_test.go
```diff
@@ -115,13 +137,22 @@ var _ = Describe("FilterOut", func() {
...
-			returnedState, err := filterOut(state, "{veth*,vnet*}")
+			returnedState, err := filterOut(state)
 
 			Expect(err).ToNot(HaveOccurred())
 			Expect(returnedState).To(Equal(filteredState))
 		})
 	})
 })
+
+func setEnvFilter(filter string) {
```
phoracek: great idea to have this helper!

pkg/helper/client.go
HlinaCZ: `match.Nothing` isn't expression so it can't be compared with the compiled glob.
So sticking with empty match for now.

pkg/helper/client_test.go
HlinaCZ: Thanks.

phoracek:
@HlinaCZ please rebase it on master manually. This PR has some bonding commits that have been already merged.

deploy/operator.yaml
```diff
@@ -67,4 +72,5 @@ metadata:
   name: nmstate-config
   namespace: nmstate
 data:
-  node_network_state_refresh_interval: "5"
\ No newline at end of file
+  node_network_state_refresh_interval: "5"
+  interfaces_filter: "veth*"
```
SchSeba: related to the test of two interfaces should this be `"{veth*,}"`

SchSeba: and maybe we should update the documentation about this option

phoracek: We should absolutely document this in https://github.com/nmstate/kubernetes-nmstate/blob/master/docs/developer-guide-state.md. @HlinaCZ could you add one paragraph there, mentioning that we have this config map with the filter value, it is glob, show examples you used in test.

HlinaCZ: I thought this file https://github.com/nmstate/kubernetes-nmstate/blob/master/docs/user-guide-state-reporting.md would be more appropriate for its use case.

pkg/helper/client.go
```diff
@@ -145,3 +155,44 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+func getFilter() *glob.Glob {
+	if !interfacesFilterGlobIsSet {
+		interfacesFilter := os.Getenv("INTERFACES_FILTER")
+		interfacesFilterGlob = glob.MustCompile(interfacesFilter)
+		interfacesFilterGlobIsSet = true
+	}
+	return &interfacesFilterGlob
+}
+
+func filterOut(currentState nmstatev1alpha1.State) (nmstatev1alpha1.State, error) {
```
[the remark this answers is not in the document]
HlinaCZ: If I am getting this right it should like like in this commit: https://github.com/nmstate/kubernetes-nmstate/pull/135/commits/3a4dcb8ab6b59bd7136658106f2a9373c52c95e0#diff-658839e1b7faf17e49850dcc315fe151R160
But have the glob compiled in init and set as an argument. Am I right or did I misunderstood?

docs/user-guide-state-reporting.md
```diff
@@ -14,6 +14,28 @@ on [local cluster](deployment-local-cluster.md) or your
 
 The operator will periodically update the reported state of node interfaces.
 
+We can set the period of update time in seconds in config map in variable
```
phoracek: Please add a new section "Additional Configuration" to the bottom of the file with this text

Makefile
```diff
@@ -52,6 +52,8 @@ description = build/_output/description
 
 resources = deploy/namespace.yaml deploy/service_account.yaml deploy/role.yaml deploy/role_binding.yaml
 
+export INTERFACES_FILTER = "{veth*}"
```
phoracek: Please comment that we need it for unit tests

pkg/helper/client.go
```diff
@@ -5,17 +5,33 @@ import (
...
+	yaml "sigs.k8s.io/yaml"
 
+	"github.com/gobwas/glob"
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
 const nmstateCommand = "nmstatectl"
 
+var (
+	interfacesFilter     string
```
phoracek: We can drop this variable, it is not used anywhere except the init. Only Glob is used

docs/user-guide-state-reporting.md
```diff
@@ -249,3 +249,28 @@ status:
 ```
 
 If you want to learn more about the `observedState` API, see [nmstate documentation](https://nmstate.github.io/).
+
+## Additional configuration
```
SchSeba: small nit: It can be nice to add a note that the user need to remove manually all the nmstate-handler pods if he updates the configMap because the daemonset will not restart the pods on a change

pkg/helper/client.go
```diff
@@ -98,11 +113,13 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
 		return fmt.Errorf("error running nmstatectl show: %v", err)
 	}
 
-	// Let's update status with current network config from nmstatectl
-	nodeNetworkState.Status = nmstatev1alpha1.NodeNetworkStateStatus{
-		CurrentState: nmstatev1alpha1.State(currentState),
+	filteredState, err := filterOut(nmstatev1alpha1.State(currentState), interfacesFilterGlob)
+	if err != nil {
+		return fmt.Errorf("error filtering out interfaces from NodeNetworkState: %v", err)
```
qinqon: This should be a warning, curentState should work, showing all the interfaces.

pkg/helper/client_test.go
```diff
@@ -0,0 +1,117 @@
...
+`)
+			interfacesFilterGlob = glob.MustCompile("{veth*,vnet*}")
+		})
+
+		It("it should filter out all interfaces matching any of these prefixes and keep the others", func() {
+			returnedState, err := filterOut(state, interfacesFilterGlob)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(returnedState).To(Equal(filteredState))
+		})
+	})
```
qinqon: Can we add a bad filter expression test ? or this concept does not exist at globg ?

docs/user-guide-state-reporting.md
SchSeba: yep still work in progress there are some external operators that watch on the config map change and kill all the relevant pods but again is a separate project.

pkg/helper/client_test.go
phoracek: Filter our can be broken by used only when they set invalid glob. That would however fail early in init(), not as a part of filterOut. I think we cannot test it (and there is no real need, as it would fail to run the handler completely).

docs/user-guide-state-reporting.md
phoracek: Done

SchSeba:
ci test please

phoracek:
Failed on the DHCP test, relevant tests have passed.

PR #140, opened 2019-07-18
Configure vlan_filtering with iproute utils

qinqon:
The linux bridge vlan filtering is only supported a newer versions
of NetworkManager and not yet implemented by nmstate at all, so we
introduce a temporary workaround here to configure it the iproute
utilities "ip" and "bridge" using hostNetworking at nmstate-handler
to run then at node networking.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1012,6 +1012,38 @@
   version = "0.2.0"
 
 [[projects]]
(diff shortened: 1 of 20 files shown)


phoracek:
This is being implemented in nmstate. Our bash workaround works fine, let's wait for the proper solution.

PR #142, opened 2019-07-23
Configure linux bridge vlan-filtering manually

qinqon:
Only future version of nmstate/[NM](https://github.com/nmstate/nmstate/pull/440) supports linux bridges vlan filtering,
to bypass this problem this PR calls iproute commands ip/bridge to set the
vlan_filtering flag and add all vlans to the bridges itself and to outbound
interfaces.

### DIFF
--- a/build/bin/vlan-filtering
+++ b/build/bin/vlan-filtering
@@ -13,4 +13,4 @@ for bridge in $@; do
    done
 
 done
-bridge vlan show
+bridge -j vlan show
--- a/build/bin/vlan-filtering
+++ b/build/bin/vlan-filtering
@@ -13,4 +13,3 @@ for bridge in $@; do
    done
 
 done
-bridge -j vlan show
--- a/build/bin/vlan-filtering
+++ b/build/bin/vlan-filtering
@@ -1,23 +1,16 @@
 #!/bin/bash -e
-IFS=$'\n'
-sed_expression="^.*: (.*) state .* : .*master (.*) state.*$"
-vlan_range=1-4094
-for port in $(bridge link show | grep -v @ | grep "state forwarding"); do
-   outbound_name=$(echo $port | sed -r "s/$sed_expression/\1/")
-   bridge_name=$(echo $port | sed -r "s/$sed_expression/\2/")
-   bridges=$(printf $bridge_name\\n$bridges | uniq)
-   outbounds=$(printf $outbound_name\\n$outbounds)
-done
 
-# We need activate vlan_filtering first to be able
-# to set vlans on outbound ports
-for bridge in $bridges; do
+vlan_range=1-4094
+
+for bridge in $@; do
    ip link set dev $bridge type bridge vlan_filtering 1
    bridge vlan add dev $bridge vid $vlan_range self
-done
 
-for outbound in $outbounds; do
-   bridge vlan add dev $outbound vid $vlan_range
-done
+   IFS=$'\n'
+   for port in $(bridge link show $bridge | grep -v @ | grep "state forwarding"); do
+      outbound=$(echo $port | sed -r "s/^.*: (.*) state .* :.*$/\1/")
+      bridge vlan add dev $outbound vid $vlan_range
+   done
 
+done
 bridge vlan show
--- a/build/bin/vlan-filtering
+++ b/build/bin/vlan-filtering
@@ -1,0 +1,23 @@
+#!/bin/bash -e
+IFS=$'\n'
+sed_expression="^.*: (.*) state .* : .*master (.*) state.*$"
+vlan_range=1-4094
+for port in $(bridge link show | grep -v @ | grep "state forwarding"); do
+   outbound_name=$(echo $port | sed -r "s/$sed_expression/\1/")
+   bridge_name=$(echo $port | sed -r "s/$sed_expression/\2/")
+   bridges=$(printf $bridge_name\\n$bridges | uniq)
+   outbounds=$(printf $outbound_name\\n$outbounds)
+done
+
+# We need activate vlan_filtering first to be able
+# to set vlans on outbound ports
+for bridge in $bridges; do
+   ip link set dev $bridge type bridge vlan_filtering 1
+   bridge vlan add dev $bridge vid $vlan_range self
+done
+
+for outbound in $outbounds; do
+   bridge vlan add dev $outbound vid $vlan_range
+done
+
+bridge vlan show
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -14,6 +14,9 @@ spec:
         app: kubernetes-nmstate
         name: nmstate-handler
     spec:
+      # Needed to force vlan filtering config with iproute commands until
+      # future nmstate/NM is in place.
+      # https://github.com/nmstate/nmstate/pull/440
       hostNetwork: true
       serviceAccountName: nmstate-handler
       nodeSelector:
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -14,6 +14,7 @@ spec:
         app: kubernetes-nmstate
         name: nmstate-handler
     spec:
+      hostNetwork: true
       serviceAccountName: nmstate-handler
       nodeSelector:
         beta.kubernetes.io/arch: amd64
--- a/pkg/helper/bridges.go
+++ b/pkg/helper/bridges.go
@@ -1,0 +1,31 @@
+package helper
+
+import (
+	"fmt"
+
+	"github.com/tidwall/gjson"
+
+	yaml "sigs.k8s.io/yaml"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+func getBridgesUp(desiredState nmstatev1alpha1.State) ([]string, error) {
+	foundBridges := []string{}
+
+	desiredStateYaml, err := yaml.YAMLToJSON([]byte(desiredState))
+	if err != nil {
+		return foundBridges, fmt.Errorf("error converting desiredState to JSON: %v", err)
+	}
+
+	queryResults := gjson.ParseBytes(desiredStateYaml).
+		Get("interfaces.#(type==linux-bridge)#").
+		Get("#(state==up)#.name").
+		Array()
+
+	for _, queryResult := range queryResults {
+		foundBridges = append(foundBridges, queryResult.String())
+	}
+
+	return foundBridges, nil
+}
--- a/pkg/helper/bridges_test.go
+++ b/pkg/helper/bridges_test.go
@@ -1,0 +1,132 @@
+package helper
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var (
+	badYaml = nmstatev1alpha1.State("}")
+	empty   = nmstatev1alpha1.State("")
+
+	noBridges = nmstatev1alpha1.State(`interfaces:
+  - name: eth1
+    type: ethernet
+    state: up
+  - name: bond1
+    type: bond
+    state: up
+    link-aggregation:
+      mode: active-backup
+      slaves:
+        - eth1
+      options:
+        miimon: '120'
+`)
+	noBridgesUp = nmstatev1alpha1.State(`interfaces:
+  - name: eth1
+    type: ethernet
+    state: up
+  - name: br1
+    type: linux-bridge
+    state: down
+  - name: br2
+    type: linux-bridge
+    state: absent
+`)
+	someBridgesUp = nmstatev1alpha1.State(`interfaces:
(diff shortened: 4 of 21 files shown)


dankenigsberg:
> Only newere version of nmstate/NM supports linux bridges vlan filtering,
> to bypass this problem this PR calls iproute commands ip/bridge to set the
> vlan_filtering flag and add all vlans to the bridges itself and to outbound
> interfaces.
> 
> Signed-off-by: Quique Llorente [ellorent@redhat.com](mailto:ellorent@redhat.com)

newere -> future,  and add a link to Miguel's PR.

deploy/operator.yaml
```diff
@@ -14,6 +14,7 @@ spec:
         app: kubernetes-nmstate
         name: nmstate-handler
     spec:
+      hostNetwork: true
```
dankenigsberg: please add a comment that we add this only for this hack.

build/bin/vlan-filtering
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+IFS=$'\n'
+sed_expression="^.*: (.*) state .* : .*master (.*) state.*$"
+vlan_range=1-4094
+for port in $(bridge link show | grep -v @); do
+   outbound_name=$(echo $port | sed -r "s/$sed_expression/\1/")
+   bridge_name=$(echo $port | sed -r "s/$sed_expression/\2/")
+   bridges=$(printf $bridge_name\\n$bridges | uniq)
```
dankenigsberg: a safer way to list all host bridges:
```
ip -o link show type bridge
```

build/bin/vlan-filtering
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+IFS=$'\n'
+sed_expression="^.*: (.*) state .* : .*master (.*) state.*$"
+vlan_range=1-4094
+for port in $(bridge link show | grep -v @); do
+   outbound_name=$(echo $port | sed -r "s/$sed_expression/\1/")
```
dankenigsberg: I think you are going to capture here also `tap` devices, which you should not.

qinqon:
Bonding e2e test is not working I am on it, but we can start review.

pkg/helper/client.go
```diff
@@ -102,5 +113,13 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 	if len(desiredState) == 0 {
 		return "Ignoring empty desired state", nil
 	}
-	return set(string(nodeNetworkState.Spec.DesiredState))
+	outputSet, err := set(string(nodeNetworkState.Spec.DesiredState))
```
dankenigsberg: you might be better off parsing the desired state yaml, iterating over desired bridges and the interfaces explicitly set inside them. please consider.

build/bin/vlan-filtering
qinqon: I am setting vlan filtering config to bridges that have a port to outbound interface filtering ports with "grep -v @".
ip -o link show type bridge give you all bridges even if they don't have a port to an outbound interface.

pkg/helper/client.go
qinqon: Then maybe we want to reconsider this https://github.com/nmstate/kubernetes-nmstate/pull/140, when NM/nmstate is ready customers will configure vlan-filtering explicitly at k8s-nmstate, with the above PR we can provide that before having nmstate/NM

build/bin/vlan-filtering
qinqon: Isn't VMIs tap interfaces at different network namespaces ? or you mean VM out from kubevirt ?

pkg/helper/client.go
qinqon: In case we want to parse desiredState we all need to ignore bridges with "state: absent", would be nice to be able to ask nmstatectl about interfaces managed by it.

qinqon: After talking with nmstate as expected we are the owners so we should know what interface do we manage, so parsing desiredState have to be, since #140 maybe is too much work.

build/bin/vlan-filtering
dankenigsberg: you are right about `tap`, but for each VMI you'd see here a half of a `veth` pair which you should not touch.

build/bin/vlan-filtering
dankenigsberg: I think that you should set `vlan_filtering=1` even on floating nic-less bridges.

pkg/helper/bridges.go
```diff
@@ -0,0 +1,31 @@
...
+
+func getBridgesUp(desiredState nmstatev1alpha1.State) ([]string, error) {
+	foundBridges := []string{}
+
+	desiredStateYaml, err := yaml.YAMLToJSON([]byte(desiredState))
+	if err != nil {
+		return foundBridges, fmt.Errorf("error converting desiredState to JSON: %v", err)
+	}
+
+	queryResults := gjson.ParseBytes(desiredStateYaml).
```
phoracek: Interesting approach :D I'd expect us to convert it to native Go structure and then iterate that. Won't pushback since it is a temporary workaround.

pkg/helper/bridges_test.go
```diff
@@ -0,0 +1,132 @@
...
+      options:
+        miimon: '120'
+`)
+	noBridgesUp = nmstatev1alpha1.State(`interfaces:
+  - name: eth1
+    type: ethernet
+    state: up
+  - name: br1
+    type: linux-bridge
+    state: down
```
phoracek: Is state mandatory? If not, what is the default?

pkg/helper/bridges_test.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+	Context("when desired state is not a yaml", func() {
+		BeforeEach(func() {
+			desiredState = badYaml
+		})
+		It("should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+	Context("when desired state is empty", func() {
```
phoracek: could you handle all non-failing ones via a table?

test/e2e/nodenetworkstate_test.go
```diff
@@ -96,7 +96,6 @@ var _ = Describe("NodeNetworkState", func() {
 
 				// Let's clean the bridge directly in the node
 				// bypassing nmstate
-				deleteConnectionAtNodes("eth1")
```
phoracek: how is this related?

pkg/helper/client.go
```diff
@@ -102,5 +113,17 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
```
phoracek: setOutput

phoracek: please add an empty line above

pkg/helper/client.go
```diff
@@ -102,5 +113,17 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+	if err != nil {
+		return outputSet, err
+	}
+	bridgesUp, err := getBridgesUp(nodeNetworkState.Spec.DesiredState)
```
phoracek: empty line above (after error check), the same for the rest of the function

phoracek: please add a comment referencing the issue

pkg/helper/client.go
```diff
@@ -102,5 +113,17 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+	if err != nil {
+		return "", fmt.Errorf("error retrieving up bridges from desired state: %v", err)
```
phoracek: This and the next return are questionable. We have no way to rollback if our hack after `nmstatectl set` fails. How would we report it in conditions? Which state are we in? I think we should at least log a big warning.

pkg/helper/bridges.go
qinqon: I like this approach better for this cases, even for future policies implementation when you have to query the nmstate yaml to get values.

pkg/helper/bridges_test.go
qinqon: I think it's mandatory https://github.com/nmstate/nmstate/blob/master/libnmstate/schemas/operational-state.yaml#L71 it does not have minimun: 0

test/e2e/nodenetworkstate_test.go
qinqon: Now eth1 is a real secondary nic from kubevirtci, to emulate a real outbound interface.

phoracek: We still create a connection profile for it, therefore we should clean up after the test is done. I will not remove the interface.

pkg/helper/bridges.go
phoracek: Granted we keep it as bytes, you might be right. Would make it way easier to implement for sure.

test/e2e/nodenetworkstate_test.go
qinqon: We cannot remove it

phoracek: Why not? What happens if you do? Who handles clean-up after the test?

pkg/helper/client.go
qinqon: We need to run this after nmstate since, we create the bridges there, the proper option would be to use commit/rollback and do the rollback in case vlan-filtering fails, so we consider the two operatoins as one.

For now I am going to add a big warning

Is that ok ?

pkg/helper/bridges_test.go
qinqon: This is temporary workaround don't want to spend more time on it, prefer to use time on e2e

build/bin/vlan-filtering
qinqon: I am setting it now at all bridges from desiredState so I cover that, I am going to mark this as resolved.

pkg/helper/client.go
qinqon: I have add filtering by desiredState bridges so I am resolving this.

pkg/helper/bridges_test.go
phoracek: "temporary" is the biggest lie! But ok, it's not that bad. But next time, please consider `DescribeTable`

pkg/helper/client.go
phoracek: Yes, big warning is fine. And a TODO to handle the rollback once we have needed tools in `nmstatectl`.

test/e2e/nodenetworkstate_test.go
qinqon: Humm we should remove them, don't know why I delete the delete, adding them back.

pkg/helper/client.go
qinqon: Well as commented it's going to re-concile if vlan-filtering fails so that's like a big warning, so added the TODO about commit/rollback

build/bin/vlan-filtering
qinqon: But I am filtering out veths since they appear with "@" so we are not seting vlans for them VMIs or normal pods.

build/bin/vlan-filtering
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+
+vlan_range=1-4094
+
+for bridge in $@; do
+   ip link set dev $bridge type bridge vlan_filtering 1
+   bridge vlan add dev $bridge vid $vlan_range self
+
+   IFS=$'\n'
+   for port in $(bridge link show $bridge | grep -v @ | grep "state forwarding"); do
```
phoracek: not sure if it helps and if it is available on RHEL, but you can use -json to get structured output

qinqon: Thanks! it's working bridge -j vlan show, that makes e2e easier !!!

build/bin/vlan-filtering
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+
+vlan_range=1-4094
```
maiqueb: There's one thing that puzzles me with the selected range.

So you activate vlan filtering **and** whitelist **all** the vlans (on all ports) ? It's unclear to me what are the intentions.

One possibility is that there are more "users" of that bridge and you don't want to impair their networking, but still, would like to point this out.

dankenigsberg: We need to allow all incoming traffic from the host interface. When we configure the bridge we have no idea which NetAttDef are going to be defined in the cluster (nor do we want to know)

qinqon:
@dankenigsberg @phoracek @maiqueb I have manually confirmed that nmstate does not remove the ports not configured by it (AKA veths) so we are good there.

build/bin/vlan-filtering
qinqon: @dankenigsberg so we are good ? I don't see any change needed since we are not touching veths at all here

maiqueb:
> Remember that by default it also sets vlan 1 as pvid, meaning you will get a 'freely' provisioned vlan with the following config:
> "1 PVID Egress Untagged"
> 
> Not sure you want that, or explicitly want to get rid of it.

Since you're redefining the vlan 1, the default is overwritten, disregard this comment.

PR #144, opened 2019-07-26
Control nns status at different controller

qinqon:
It will do reconcile after some time to refresh status the nns controller
will not reconcile after some time.


pkg/controller/nodenetworkstatestatus/nodenetworkstatestatus_controller.go
```diff
@@ -0,0 +1,128 @@
+package nodenetworkstatestatus
```
phoracek: It is not for the whole status. e.g. conditions will be edited by both. please rename it to something closer to node network state reporting controller. The other should be renamed to node network state configuration controller.

pkg/controller/nodenetworkstatestatus/nodenetworkstatestatus_controller.go
```diff
@@ -0,0 +1,128 @@
...
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return nmstate.EventIsForThisNode(createEvent.Meta)
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			// This controller responsability is updates, receiving
+			// deletes is of no use
+			return false
+		},
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
+			return nmstate.EventIsForThisNode(updateEvent.MetaNew)
```
phoracek: Does this get triggered at the moment a new desired status is applied?

qinqon: Yep and this is good, although we don't know if it's updated.

pkg/controller/nodenetworkstatereport/nodenetworkstatereport_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	forThisNode := predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return nmstate.EventIsForThisNode(createEvent.Meta)
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			// This controller responsability is updates, receiving
+			// deletes is of no use
+			return false
+		},
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
```
phoracek: Just an idea, in order to rereconcile right after a configuration is set, we could write `something somewhere` in the CR once saved (maybe version is already doing that), that change will be observed by report controller and status refreshed. I guess it happens even now, not sure

qinqon: The only "problem" we have with this is that in case desiredState is apply before our Reconcile here at update we will have to wait for next run... I don't see this as a big problem.

PR #156, opened 2019-08-28
WIP handle bridge-over-default-nic scenario via a script

phoracek:
There are issues using kubernetes-nmstate to configure bridge over
the default NIC. Moreover, these issues are really difficult to debug.

This patch covers this specific scenario via a script (nmcli+ip link).

### DIFF
--- a/build/bin/bridge-over-nic
+++ b/build/bin/bridge-over-nic
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -xe
+
+bridge_name=$1
+port_name=$2
+vlan_range_min=$3
+vlan_range_max=$4
+
+VLAN_RANGE="${vlan_range_min}-${vlan_range_max}"
+
+if [ -z "${bridge_name}" ]; then
+    echo 'Please define name of the bridge via BRIDGE_NAME variable'
+    exit 1
+fi
+
+if [ -z "${port_name}" ]; then
+    echo 'Please define name of the port via PORT_NAME variable'
+    exit 1
+fi
+
+if [[ ! $(ip address show $port_name) ]]; then
+    echo "Interface ${port_name} has not been found in the system"
+    exit 1
+fi
+
+if [[ ! $(nmcli --fields connection.id con show ${bridge_name}) ]]; then
+    echo 'Bridge connection profile does not exists yet, creating it'
+    nmcli con add type bridge ifname ${bridge_name} con-name ${bridge_name}
+    nmcli con mod ${bridge_name} connection.autoconnect-slaves 1
+fi
+
+if [[ ! $(nmcli --fields connection.id con show ${port_name}) ]]; then
+    echo 'Port connection profile does not exists yet, creating it'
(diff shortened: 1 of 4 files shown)


phoracek:
@SchSeba @oshoval @dankenigsberg  please review this madness, still in progress though

build/bin/bridge-over-nic
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -xe
+
+bridge_name=$1
+nic_name=$2
+vlan_range_min=$3
+vlan_range_max=$4
+
+VLAN_RANGE="${vlan_range_min}-${vlan_range_max}"
```
oshoval: vlan_range_min/max should be mandatory ?
if so maybe worth to add a check like the ones below ?

oshoval:
lgtm

build/bin/bridge-over-nic
phoracek: good point, thanks

pkg/controller/nodenetworkstateconfiguration/nodenetworkstateconfiguration_controller.go
```diff
@@ -145,3 +156,88 @@ func (r *ReconcileNodeNetworkStateConfiguration) Reconcile(request reconcile.Req
...
+        - name: eth0
+          vlans:
+          - vlan-range-min: 1
+            vlan-range-max: 4094
+EOF
+
+*/
+
+// TODO in order to maintain compatibility in future, trigger workaround if the requested configuration matches
+func detectBridgeWorkaround(desiredState nmstatev1alpha1.State) (string, string, string, string) {
```
dankenigsberg: I don't understand how the detection works. I expected to see here "if name == mad-workaround-policy", to make sure it does not kick in for run-of-the-mill policies.

phoracek: The issue with that would be that we would eventually need to introduce upgrade path or maintain the script forever. Instead, this specific scenario is handled by the script. Once we properly support it via nmstate, we can drop the workaround and this `desiredState` will do exactly the same job.

dankenigsberg: Ok

phoracek: BTW I have tried it on bare-metal KNI and it worked! Not as a part of kubernetes-nmstate though, just the script alone, deployed as a DaemonSet.

test/e2e/nodenetworkstate_test.go
```diff
@@ -115,6 +141,42 @@ var _ = Describe("NodeNetworkState", func() {
 				}
 			})
 		})
+		FContext("with a linux bridge workaround (dhcp+trunk) up", func() {
```
qinqon: s/FContext/Context/g

phoracek:
This has been fixed properly and I'm glad we did it.

PR #159, opened 2019-09-04
Removed redundant ethernet setups

HlinaCZ:
Removed redundant ethernet setups from e2e tests and doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,10 +76,6 @@ metadata:
 spec:
   desiredState:
     interfaces:
-    - name: eth1
-      description: An ethernet network interface
-      type: ethernet
-      state: up
     - name: br1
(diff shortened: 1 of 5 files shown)


phoracek:
@HlinaCZ it seems like we have an issue where we start testing sooner than NodeNetworkState for the node is initialized. I think the issue is here https://github.com/nmstate/kubernetes-nmstate/blob/master/test/e2e/utils.go#L141, we use Eventually and retry if we fail to Update, but we fail hard immediately if we fail on Get. We should return Get error if not nil, not Expect. Would you please create another PR solving it? Thanks!

phoracek:
@HlinaCZ https://github.com/nmstate/kubernetes-nmstate/pull/159#pullrequestreview-283689233

PR #162, opened 2019-09-06
Change docker image to fedora:30

HlinaCZ:


phoracek:
@HlinaCZ we can close this PR. @SchSeba is switching to Fedora in https://github.com/nmstate/kubernetes-nmstate/pull/163. We need to get that quickly in the code-base, sorry.

PR #163, opened 2019-09-06
Fix vlan filter bash

SchSeba:
This PR fix the vlan filter function.

Now we pass one bridge every time with all it's interfaces.
We also confiure the vlan filtering only on the uplink interfaces
an not on the bridge itself.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/kubernetes-nmstate/163)

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,12 +1,7 @@
-FROM centos:7
+FROM fedora:30
 
-RUN yum -y install epel-release && \
-    yum -y update && \
-    yum -y install \
-        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.8/1.el7/noarch/python2-libnmstate-0.0.8-1.el7.noarch.rpm \
-        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.8/1.el7/noarch/nmstate-0.0.8-1.el7.noarch.rpm \
-        iproute && \
-    yum clean all
+RUN sudo dnf install -y nmstate iproute && \
+    sudo dnf clean all
 
 # Cannot change the binary to nmstate-handler since the name
 # is taken from the directory name [1]
--- a/build/bin/vlan-filtering
+++ b/build/bin/vlan-filtering
@@ -1,15 +1,13 @@
-#!/bin/bash -e
+#!/bin/bash -xe
+
+bridge=$1
+shift
+ports=$@
 
 vlan_range=2-4094
 
-for bridge in $@; do
-   ip link set dev $bridge type bridge vlan_filtering 1
-   bridge vlan add dev $bridge vid $vlan_range self
+ip link set dev $bridge type bridge vlan_filtering 1
 
-   IFS=$'\n'
-   for port in $(bridge link show $bridge | grep -v @ | grep "state forwarding"); do
-      outbound=$(echo $port | sed -r "s/^.*: (.*) state .* :.*$/\1/")
-      bridge vlan add dev $outbound vid $vlan_range
-   done
-
+for port in $ports; do
+    bridge vlan add dev $port vid $vlan_range
 done
--- a/pkg/helper/bridges.go
+++ b/pkg/helper/bridges.go
@@ -10,22 +10,30 @@ import (
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-func getBridgesUp(desiredState nmstatev1alpha1.State) ([]string, error) {
-	foundBridges := []string{}
+func getBridgesUp(desiredState nmstatev1alpha1.State) (map[string][]string, error) {
+	foundBridgesWithPorts := map[string][]string{}
 
 	desiredStateYaml, err := yaml.YAMLToJSON([]byte(desiredState))
 	if err != nil {
-		return foundBridges, fmt.Errorf("error converting desiredState to JSON: %v", err)
+		return foundBridgesWithPorts, fmt.Errorf("error converting desiredState to JSON: %v", err)
 	}
 
-	queryResults := gjson.ParseBytes(desiredStateYaml).
+	bridgesUp := gjson.ParseBytes(desiredStateYaml).
 		Get("interfaces.#(type==linux-bridge)#").
-		Get("#(state==up)#.name").
+		Get("#(state==up)#").
 		Array()
 
-	for _, queryResult := range queryResults {
-		foundBridges = append(foundBridges, queryResult.String())
+	for _, bridgeUp := range bridgesUp {
+		portList := []string{}
+		for _, port := range bridgeUp.Get("bridge.port.#.name").Array() {
+			portList = append(portList, port.String())
+		}
+
+		// Add only the bridge we have some outbound port to configure
+		if len(portList) > 0 {
+			foundBridgesWithPorts[bridgeUp.Get("name").String()] = portList
+		}
 	}
 
-	return foundBridges, nil
+	return foundBridgesWithPorts, nil
 }
--- a/pkg/helper/bridges_test.go
+++ b/pkg/helper/bridges_test.go
@@ -36,6 +36,13 @@ var (
     type: linux-bridge
     state: absent
 `)
+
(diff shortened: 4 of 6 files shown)


phoracek:
We have to fix the vlan test, it is failing on Fedora 30 (maybe a different format of `bridge vlan show`?)

build/bin/vlan-filtering
```diff
@@ -2,14 +2,13 @@
...
+for bridge in $1; do
    ip link set dev $bridge type bridge vlan_filtering 1
-   bridge vlan add dev $bridge vid $vlan_range self
+   #bridge vlan add dev $bridge vid $vlan_range self
 
    IFS=$'\n'
-   for port in $(bridge link show $bridge | grep -v @ | grep "state forwarding"); do
-      outbound=$(echo $port | sed -r "s/^.*: (.*) state .* :.*$/\1/")
-      bridge vlan add dev $outbound vid $vlan_range
+   for port in ${*:2}; do
```
phoracek: what does the `${*:2}` do?

pkg/helper/client.go
```diff
@@ -27,8 +27,11 @@ func show(arguments ...string) (string, error) {
 	return stdout.String(), nil
 }
 
-func applyVlanFiltering(bridgesUp []string) (string, error) {
-	cmd := exec.Command("vlan-filtering", bridgesUp...)
+func applyVlanFiltering(bridgeName string,ports []string) (string, error) {
+	command := []string{bridgeName}
+	command = append(command, ports...)
+
+	cmd := exec.Command("vlan-filtering",command...)
```
phoracek: Please run gofmt on your changes

pkg/helper/bridges_test.go
```diff
@@ -78,7 +78,7 @@ var (
 
 var _ = Describe("Network desired state bridge parser", func() {
 	var (
-		obtainedBridges []string
+		obtainedBridges map[string][]string
```
phoracek: please rename the variable to reflect what does it keep, it is not only bridge now

SchSeba:
ci test please

pkg/helper/bridges.go
```diff
@@ -24,8 +24,20 @@ func getBridgesUp(desiredState nmstatev1alpha1.State) ([]string, error) {
 		Array()
 
 	for _, queryResult := range queryResults {
-		foundBridges = append(foundBridges, queryResult.String())
+		portsQueryResults := gjson.ParseBytes(desiredStateYaml).
```
qinqon: We can simplify this with:
```golang
package helper                                                                  
                                                                                
import (                                                                        
    "fmt"                                                                       
                                                                                
    "github.com/tidwall/gjson"                                                  
                                                                                
    yaml "sigs.k8s.io/yaml"                                                     
                                                                                
)                                                                               
                                                                                
func getBridgesUp(desiredState nmstatev1alpha1.State) (map[string][]string, error) {
    foundBridgesWithPorts := map[string][]string{}                              
                                                                                
    desiredStateYaml, err := yaml.YAMLToJSON([]byte(desiredState))              
    if err != nil {                                                             
        return foundBridgesWithPorts, fmt.Errorf("error converting desiredState to JSON: %v", err)
    }                                                                           
                                                                                
    bridgesUp := gjson.ParseBytes(desiredStateYaml).                            
        Get("interfaces.#(type==linux-bridge)#").                               
        Get("#(state==up)#").                                                   
        Array()                                                                 
                                                                                
    for _, bridgeUp := range bridgesUp {                                        
        portList := []string{}                                                  
        for _, port := range bridgeUp.Get("bridge.port.#.name").Array() {       
            portList = append(portList, port.String())                          
        }                                                                       
        foundBridgesWithPorts[bridgeUp.Get("name").String()] = portList         
    }                                                                           
                                                                                
    return foundBridgesWithPorts, nil                                           
}                                                            
```

pkg/helper/bridges_test.go
```diff
@@ -126,7 +126,13 @@ var _ = Describe("Network desired state bridge parser", func() {
 		})
 		It("should return the slice with the bridges", func() {
 			Expect(err).ToNot(HaveOccurred())
-			Expect(obtainedBridges).To(ConsistOf("br1", "br2"))
+			Expect(len(obtainedBridges)).To(Equal(2))
```
qinqon: Can we add a test with some bridges down and also son bridges up but without ports ?

build/bin/vlan-filtering
```diff
@@ -2,14 +2,13 @@
 
 vlan_range=2-4094
 
-for bridge in $@; do
+for bridge in $1; do
```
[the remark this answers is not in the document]
qinqon: the normal thing is to take first one and use shift.
```bash
bridge=$1
shift
ports=$@
```
and use those

build/bin/vlan-filtering
qinqon: Not needed if we just use "shift" to get $@ at the top of the script

qinqon:
@SchSeba to fix e2e tests you need to remove vlans check for bridges, since we are removing it in the script 

```golang
iff --git a/test/e2e/nodenetworkstate_test.go b/test/e2e/nodenetworkstate_test.go
index 9a4d27c7..cc97b861 100644
--- a/test/e2e/nodenetworkstate_test.go
+++ b/test/e2e/nodenetworkstate_test.go
@@ -225,7 +225,7 @@ var _ = Describe("NodeNetworkState", func() {
                                                parsedVlans := gjson.Parse(bridgeVlans)
 
                                                hasVlans(parsedVlans.Get("bond1"), 4094)
-                                               hasVlans(parsedVlans.Get("br1"), 4094)
+                                               hasVlans(parsedVlans.Get("br1"), 1)
 
                                                eth1Vlans := parsedVlans.Get("eth1").Array()
                                                Expect(eth1Vlans).To(BeEmpty())

```

qinqon:
@SchSeba please take a look have added the missing parts

build/Dockerfile
```diff
@@ -1,12 +1,7 @@
...
+FROM fedora:30
 
-RUN yum -y install epel-release && \
-    yum -y update && \
-    yum -y install \
-        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.8/1.el7/noarch/python2-libnmstate-0.0.8-1.el7.noarch.rpm \
-        https://kojipkgs.fedoraproject.org/packages/nmstate/0.0.8/1.el7/noarch/nmstate-0.0.8-1.el7.noarch.rpm \
-        iproute && \
-    yum clean all
+RUN sudo dnf install -y nmstate iproute && \
```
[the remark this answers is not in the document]
phoracek: We take whatever is released in repositories our OS provides. Pinning version is something you see with Python and PyPI where multiple versions are available at the same time, but that is not the case here. (pushing back just because of I'm not used to see version request in `dnf`)

pkg/helper/bridges_test.go
phoracek: I'm all for more tests, but I'd rather have those in separate follow-up PRs (like this https://github.com/nmstate/kubernetes-nmstate/pull/160). We should get basic fix merged ASAP to be able to test it on real environments.

build/Dockerfile
```diff
@@ -1,12 +1,7 @@
-FROM centos:7
+FROM fedora:30
```
phoracek: Commenting for the commit message: Please specify what was the problem (both in the body and if possible in the title as well). Please say why it was wrong in its original state.

s/it's/its/
s/an/and/

qinqon: Done

build/bin/vlan-filtering
```diff
@@ -1,15 +1,13 @@
-#!/bin/bash -e
+#!/bin/bash -xe
 
-vlan_range=2-4094
+bridge=$1
+shift
```
phoracek: don't we have to switch twice?

qinqon:
```bash
[ellorent@localhost kubernetes-nmstate]$ bash foo.sh br1 eth1 eth2
bridge: br1
port eth1
port eth2
```
```bash
bridge=$1                                                                       
shift                                                                           
ports=$@                                                                        
echo bridge: $bridge                                                            
for port in $ports; do                                                          
    echo port $port                                                             
done                 
```

PR #166, opened 2019-09-10
Increase nmstate set timeout to 60 seconds

qinqon:
It's harcoded to 20 seconds, this PR patch the python code to
have 60 seconds.


SchSeba:
lets just wait for CI

PR #168, opened 2019-09-10
Remove unnneded linux bridge stp attributes

qinqon:
With the latest version from nmstate if stp is not enabled the
stp attributes at linux bridge port are not mandatory.

### DIFF
--- a/test/e2e/nodenetworkstate_test.go
+++ b/test/e2e/nodenetworkstate_test.go
@@ -76,9 +76,6 @@ var _ = Describe("NodeNetworkState", func() {
           enabled: false
       port:
         - name: bond1
-          stp-hairpin-mode: false
-          stp-path-cost: 100
-          stp-priority: 32
 `)
 
 		br1Absent = nmstatev1alpha1.State(`interfaces:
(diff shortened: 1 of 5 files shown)


qinqon:
If this passes we can close https://github.com/nmstate/nmstate/issues/443.

test/e2e/nodenetworkstate_test.go
```diff
@@ -213,10 +210,9 @@ var _ = Describe("NodeNetworkState", func() {
 			})
 			It("should have the bond in the linux bridge as port at currentState", func() {
 				var (
-					expectedInterfaces  = interfaces(br1WithBond1Up)
-					expectedBond        = interfaceByName(expectedInterfaces, "bond1")
-					expectedBridge      = interfaceByName(expectedInterfaces, "br1")
-					expectedBridgePorts = expectedBridge["bridge"].(map[string]interface{})["port"]
```
phoracek: why is this dropped? just a cleanup?

qinqon: expectedBridgePorts is not needed we are now harcoding "bond1" as the port name, if you don't then you have to do a partial search of attribute "name" in the port since we don't have the stp attribute in the expected data.

PR #170, opened 2019-09-12
Test reboots

qinqon:


phoracek:
@qinqon please rebase

test/e2e/reboot_test.go
```diff
@@ -0,0 +1,60 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var _ = Describe("Nmstate network connnections persistency", func() {
+	var (
+		vlan101 = nmstatev1alpha1.State(`interfaces:
+interfaces:
```
phoracek: Help me here please. Having tests focused on a single thing is important. However, our main concern is whether the default interface comes up with the same IP, otherwise we lose access to the host. Could we test that specifically?

phoracek: (I want a test that fails due to changed IP and forces us to fix that bug)

test/e2e/reboot_test.go
```diff
@@ -0,0 +1,60 @@
...
+    state: up
+    vlan:
+      base-iface: eth0
+      id: 101
+`)
+	)
+	Context("after rebooting a node with applied configuration", func() {
+		BeforeEach(func() {
+			By("Apply policy")
+			setDesiredStateWithPolicy("vlan-eth0.101", vlan101)
```
fedepaol: The test body verifies what was done in the `BeforeEach` section, with the result that
- the BeforeEach / AfterEach functions are less reusable (there will probably won't be another test in this context)
- the code is not readable

Would it make sense to get rid of BeforeEach / AfterEach and having all in the test body?

test/e2e/reboot_test.go
qinqon: Well this is just a test to ensure that network configuration is apply before kubernetes-nmstate is up and ready, but we can increase the reboot test with the issue we have found at this PR or a follow up PR whatever you see it's better.

test/e2e/reboot_test.go
qinqon: I think you are right although is not very ginkgo way, what do you think @phoracek

phoracek: I don't agree that it is less readable. It is more lines, but it very explicitly splits the setup, the test and the teardown.

I don't mind keeping everything in the test body, but if you chose to do it, please do it in separate PR and keep it consistent across the whole project.

fedepaol: BeforeEach / AfterEach are meant to contain setup / tear down code independent of the body of the test, while here we have the actual thing that the test is going to verify.
In order to understand what the test does you need to jump between the functions which it makes a bit less readable for my taste but again, it's not a big deal.

PR #173, opened 2019-09-18
Stabilize e2e tests and skip default bridge test

qinqon:
- Use kubernetes-nmstate to do tear down
- Skip moving ip from to bridge test since it's failing and we need to investigate why
- Increase ginkgo "slow test" threshold to 60 seconds

### DIFF
--- a/test/e2e/default_bridged_network_test.go
+++ b/test/e2e/default_bridged_network_test.go
@@ -46,7 +46,9 @@ var _ = Describe("NodeNetworkConfigurationPolicy default bridged network", func(
     type: linux-bridge
     state: absent
 `)
-	Context("when there is a default interface with dynamic address", func() {
+	// FIXME: This is a pending spec since we have to discover why the
+	//        cluster never goes back at kubevirtci provider
+	XContext("when there is a default interface with dynamic address", func() {
 		addressByNode := map[string]string{}
 		BeforeEach(func() {
 			By("Check eth0 is the default route interface and has dynamic address")
@@ -93,6 +95,9 @@ var _ = Describe("NodeNetworkConfigurationPolicy default bridged network", func(
 			By("Creating the policy")
 			setDesiredStateWithPolicy("default-network", createBridgeOnTheDefaultInterface)
 
+			By("Waiting until the node becomes ready again")
+			waitForNodesReady()
+
 			By("Checking that obtained the same IP address")
 			for _, node := range nodes {
 				Eventually(func() string {
@@ -105,20 +110,11 @@ var _ = Describe("NodeNetworkConfigurationPolicy default bridged network", func(
 				Eventually(func() string {
 					return defaultRouteNextHopInterface(node)
 				}, 30*time.Second, 1*time.Second).Should(Equal("brext"))
+
+				By("Verify that VLAN configuration is done properly")
+				hasVlans(node, "eth0", 2, 4094).Should(Succeed())
+				hasVlans(node, "brext", 1, 1).Should(Succeed())
 			}
-
-			By("Waiting until the node becomes ready again")
-			waitForNodesReady()
-
-			By("Verify that VLAN configuration is done properly")
-			Eventually(func() bool {
-				for _, bridgeVlans := range bridgeVlansAtNodes() {
-					hasVlans(bridgeVlans, "eth0", 2, 4094)
-					hasVlans(bridgeVlans, "brext", 1, 1)
-				}
-				return true
-			}, 15*time.Second, 1*time.Second).Should(BeTrue(), "Incorrect vlan ids at brext bridge")
-
 		})
 	})
 })
--- a/test/e2e/simple_bridge_and_bond.go
+++ b/test/e2e/simple_bridge_and_bond.go
@@ -1,18 +1,34 @@
 package e2e
 
 import (
-	"time"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
-	"github.com/tidwall/gjson"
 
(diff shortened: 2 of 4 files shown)


test/e2e/default_bridged_network_test.go
```diff
@@ -90,9 +90,13 @@ var _ = Describe("NodeNetworkConfigurationPolicy default bridged network", func(
 		})
 
 		It("should successfully move default IP address on top of the bridge", func() {
+			Skip("Cluster does not comes back")
 			By("Creating the policy")
 			setDesiredStateWithPolicy("default-network", createBridgeOnTheDefaultInterface)
 
+			By("Waiting until the node becomes ready again")
```
phoracek: Why do we have all these changes on the test if we skip? I think we should not change test that is not being ran.

test/e2e/simple_bridge_and_bond.go
```diff
@@ -90,49 +99,17 @@ var _ = Describe("NodeNetworkState", func() {
...
-				time.Sleep(1 * time.Second)
-
-				// Let's clean the bridge directly in the node
-				// bypassing nmstate
-				deleteConnectionAtNodes("eth1")
-				deleteConnectionAtNodes("eth2")
-				deleteConnectionAtNodes("br1")
-			})
-			It("should have the linux bridge at currentState", func() {
+				updateDesiredState(br1Absent)
```
phoracek: Man this is great!

test/e2e/simple_bridge_and_bond.go
```diff
@@ -90,49 +99,17 @@ var _ = Describe("NodeNetworkState", func() {
...
 				for _, node := range nodes {
-					interfacesNameForNode(node).Should(ContainElement("br1"))
+					interfacesNameForNode(node).ShouldNot(ContainElement("br1"))
```
phoracek: Don't we need eventually here?

test/e2e/utils.go
```diff
@@ -328,22 +328,45 @@ func toUnstructured(y string) interface{} {
...
 
-func bridgeVlansAtNodes() []string {
-	outputs, _ := runAtNodes("sudo", "bridge", "-j", "vlan", "show")
-	return outputs
+func bridgeVlansAtNode(node string) (string, error) {
+	return run(node, "sudo", "bridge", "-j", "vlan", "show")
 }
 
-func hasVlans(bridgeVlans string, connection string, minVlan int, maxVlan int) {
+func hasVlans(node string, connection string, minVlan int, maxVlan int) AsyncAssertion {
```
phoracek: We always use this in combination with Should(Succeed()). Would you mind just checking it inside the function (maybe rename it it checkVlans). I'd like to do some tiny refactoring on tests code and get rid of returning AsyncAssertion. It is killing my brain. Even worse when I have to explain how it works. I think our helper function should either return some basic Go value or do the whole assertion. This is on a mid way, but I don't think it makes the code more readable.

Feel free to argue. I don't believe AsyncAssertion gives us much value with `hasVlans(node, "bond1", 2, 4094).Should(Succeed())`. `hasVlans` either succeeds or not. If you want it for Eventually, return error, if not, assert inside the function.

Similar to vlansCardinality.

test/e2e/utils.go
```diff
@@ -328,22 +328,45 @@ func toUnstructured(y string) interface{} {
...
+		for expectedVlan := minVlan; expectedVlan <= maxVlan; expectedVlan++ {
+
+			vlanByIdAndConection := fmt.Sprintf("%s.#(vlan==%d)", connection, expectedVlan)
+			if !parsedBridgeVlans.Get(vlanByIdAndConection).Exists() {
+				return fmt.Errorf("bridge connection %s has no vlan %d, obtainedVlans: \n %s", connection, expectedVlan, bridgeVlans)
+			}
+		}
+		return nil
+	}, ReadTimeout, ReadInterval)
+
```
phoracek: please drop the empty line

test/e2e/utils.go
```diff
@@ -328,22 +328,45 @@ func toUnstructured(y string) interface{} {
...
+		bridgeVlans, err := bridgeVlansAtNode(node)
+		if err != nil {
+			return err
+		}
 
-		vlanByIdAndConection := fmt.Sprintf("%s.#(vlan==%d)", connection, expectedVlan)
-		ExpectWithOffset(1, parsedBridgeVlans.Get(vlanByIdAndConection).Exists()).To(BeTrue(), fmt.Sprintf("bridge connection %s has no vlan %d, obtainedVlans: \n %s", connection, expectedVlan, bridgeVlans))
+		parsedBridgeVlans := gjson.Parse(bridgeVlans)
+		for expectedVlan := minVlan; expectedVlan <= maxVlan; expectedVlan++ {
+
```
phoracek: please drop the empty line

test/e2e/default_bridged_network_test.go
qinqon: I think I merged changes to debug it, do I need to discard them ?

test/e2e/simple_bridge_and_bond.go
qinqon: It's inside "interfacesNameForNodes" it returns and AsyncAssertion

test/e2e/utils.go
qinqon: There is one issue, there is no concept of "EventualyWithOffset" so if we do the "Should" inside the function the the line of code that appears on error is the function not the test itself, returning AsyncAssertion is kind of idiom we can adopt here one option is to state it in the function with something like "eventuallyHasVlans" so we know what we are testing.

qinqon:
ci test please

test/e2e/default_bridged_network_test.go
phoracek: Meh, not a big deal.

PR #179, opened 2019-09-22
Build and deploy using travis on merge

SchSeba:
This PR add a deploy section to the travis yaml.
On every PR that is merge to master we are going to build and push
the nmstate-handler container to quay.io with laster tag.

when we create a git tag we will release a container with the same
tag on quay

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/kubernetes-nmstate/179)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,10 @@ go: "1.11"
 install: true
 script:
 - GINKGO_EXTRA_ARGS="--compilers=2" make all test/unit
+
+deploy:
+- provider: script
+  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io && make push-handler
+  skip_cleanup: true
+  on:
+    branch: master


SchSeba:
/hold

need to create a travis bot user and configure it in travis.

.travis.yml
```diff
@@ -4,3 +4,15 @@ go: "1.11"
 install: true
 script:
 - GINKGO_EXTRA_ARGS="--compilers=2" make all test/unit
+
+deploy:
+  - provider: script
+    script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io && make handler push-handler
```
phoracek: What is docker? Do you mean quay? ;)

Makefile
```diff
@@ -4,7 +4,8 @@ IMAGE_REPO ?= nmstate
 HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
 HANDLER_IMAGE_SUFFIX ?=
 HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME)$(HANDLER_IMAGE_SUFFIX)
-HANDLER_IMAGE ?= $(IMAGE_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)
+IMAGE_TAG ?= "latest"
```
phoracek: Why do we need this? We already have handler_image_full_name, just export the tag with : as its prefix.

phoracek: Could we explicitly pass latest from the Travis config for master? This makefile is do long it is starting to bother me.

.travis.yml
```diff
@@ -4,3 +4,15 @@ go: "1.11"
 install: true
 script:
 - GINKGO_EXTRA_ARGS="--compilers=2" make all test/unit
```
phoracek: Would you please add a scratch building here, while we are at it?

.travis.yml
SchSeba: we are login to quay I just reuse the variable names from our other projects

Makefile
SchSeba: I am not sure if we can pass that we can just use the latest in the `HANDLER_IMAGE_SUFFIX`

.travis.yml
SchSeba: the `all` in the make already build a container here

SchSeba:
/hold cancel

I added a new bot to the quay org and create the environment variables for travis.

Makefile
```diff
@@ -2,8 +2,8 @@ IMAGE_REGISTRY ?= quay.io
 IMAGE_REPO ?= nmstate
 
 HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
-HANDLER_IMAGE_SUFFIX ?=
-HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME)$(HANDLER_IMAGE_SUFFIX)
+HANDLER_IMAGE_SUFFIX ?= "latest"
```
qinqon: 'latest' default value was removed on purpose, since using ":latest" is not good practices, the idea is to use image un tagged, if you want to get the latest.

Some stuff related to "latest" https://blog.container-solutions.com/docker-latest-confusion

Makefile
```diff
@@ -2,8 +2,8 @@ IMAGE_REGISTRY ?= quay.io
...
+HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME):$(HANDLER_IMAGE_SUFFIX)
```
qinqon: This was done to be able to do use image without tags since ".../kubernetes-nmstate-handler:" would fail, again if we remove the latest concept from this we don't need this change.

.travis.yml
```diff
@@ -4,3 +4,15 @@ go: "1.11"
...
+
+deploy:
+- provider: script
+  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io && make handler push-handler
+  skip_cleanup: true
+  on:
+    branch: master
+- provider: script
+  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io && make handler push-handler HANDLER_IMAGE_SUFFIX=$TRAVIS_TAG
```
qinqon: We also need to make a github release, for that we were calling "make release" it was doing all, tagging, pushing and releasesing on github.
We can add "make github-release" it will just call https://github.com/nmstate/kubernetes-nmstate/blob/master/hack/release.sh without pushing or tagging and calling here.

Also the "script" sectoin is getting log better to using and script file or yaml multiline.

.travis.yml
```diff
@@ -4,3 +4,15 @@ go: "1.11"
 install: true
 script:
 - GINKGO_EXTRA_ARGS="--compilers=2" make all test/unit
+
+deploy:
+- provider: script
+  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io && make handler push-handler
```
qinqon: Humm do we need it for stdci so we don't build there ?

SchSeba: you want to do it in the stdci instead of travis?

.travis.yml
SchSeba: this provider section runs after there is a new tag in github.

Makefile
SchSeba: correct me if I am wrong here but it you build without a tag you will get latest by default no?

qinqon: Wait you are right, so then we don't need explicit "latest" here, since docker is going to do the "latest" tagging for us.

.travis.yml
qinqon: What I have in mind is having a release "PR" with version and release description and when we merge that PR a script will tag/release/push for us, that is what "make release" do

.travis.yml
qinqon: No no, I was meaning that now that we push "master" containers we may want to consume then at stdci so we don't need to build there, but if we consume there at stdci.

But thinking about using or not travis to release, we may want to stay away from github for this and do it with stdci or even prow in case we move to that.

.travis.yml
```diff
@@ -4,3 +4,15 @@ go: "1.11"
...
+  skip_cleanup: true
+  on:
+    branch: master
+- provider: script
+  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io && make handler push-handler HANDLER_IMAGE_SUFFIX=":"$TRAVIS_TAG
```
SchSeba: @qinqon This will work?

.travis.yml
phoracek: @qinqon with the `make github-release` script. Do we need to tag ourselves or is it done automatically? If automatically, we need to call it from the master branch.

Anyway, I think we could do it in a separate PR to keep things going and focused. Won't fight for it though :)

.travis.yml
phoracek: Do we need it with "latest is default"?

qinqon: You can just use "make push-handler" since "handler" is a dependency from "push-handler".
We don't need double quotes at any of the variables expansion (Maybe travis for us too)

PR #186, opened 2019-09-27
NodeSelector change does not trigger reconcile

### LINKED ISSUE #180: NodeSelector change does not trigger reconcile
When `nodeSelector` is changed and it matches a new node, this node ignores it.

HlinaCZ:
When `nodeSelector` is changed and it matches a new node, this node ignores it.

This issue was caused by reconcile predicate `UpdateFunc` callback depending on
matching the node with both the new and the old policy `nodeSelector`. This
caused `UpdateFunc` to not work properly.

Now we check our node only against the new (updated) policy `nodeSelector`.

Resolves #180 

Signed-off-by: HlinaCZ <djhlinacz@gmail.com>


phoracek:
Stuck on default network test, the rest looks fine.

PR #187, opened 2019-09-27
fix linux bridge cleanup

phoracek:
When bridge setup fails, nmstate sometimes leaves its netlink iface
behind although it removes the connection. This issue is tracked in
https://github.com/nmstate/nmstate/issues/516. Until that gets
resolved, we fix it with ugly sed workaround.


qinqon:
we have see some failures at the bridge still there even with this fix, but it can be other scenarios where we don't do a proper "absent" state instead of bad setup.

phoracek:
ci test please

PR #190, opened 2019-09-30
Clean up test logs

qinqon:


test/e2e/utils.go
```diff
@@ -223,7 +223,7 @@ func run(node string, command ...string) (string, error) {
 	cmd.Stderr = &stderr
 	cmd.Stdout = &stdout
 	err := cmd.Run()
-	GinkgoWriter.Write([]byte(stdout.String() + stderr.String() + "\n"))
+	GinkgoWriter.Write([]byte(fmt.Sprintf("stdout: %.500s...\n, stderr %s\n", stdout.String(), stderr.String())))
```
phoracek: do you think it would be worth setting this via an environment variable?

test/e2e/utils.go
```diff
@@ -57,12 +58,14 @@ func writePodsLogs(namespace string, sinceTime time.Time, writer io.Writer) erro
 			continue
 		}
 		defer podLogs.Close()
-		_, err = io.Copy(writer, podLogs)
+		//_, err = io.Copy(writer, podLogs)
```
phoracek: please drop this

test/e2e/utils.go
```diff
@@ -57,12 +58,14 @@ func writePodsLogs(namespace string, sinceTime time.Time, writer io.Writer) erro
...
+		rawLogs, err := ioutil.ReadAll(podLogs)
 		if err != nil {
 			io.WriteString(writer, fmt.Sprintf("error in copy information from podLogs to buf: %v\n", err))
```
phoracek: the error message needs to be changed

test/e2e/utils.go
qinqon: Let's do like this, in case of error vlans are show in full so we are good, another option is to dump to files the full output and print the truncated ones.

phoracek:
@qinqon 

```
test/e2e/utils.go:66:20: undefined: strings.ReplaceAll
```

qinqon:
@phoracek fixed, ReplaceAll is not at golang 1.11

PR #194, opened 2019-09-30
Rollback nmstate in case vlan-filtering fails

### LINKED ISSUE #184: Stabilize vlan script
The vlan script is a weak point in the configuration. However, there are few things we may do to make it more stable and predictable.

1. Wait for the bridge to become available before we start configuring.
2. Use nmstate rollback mechanism to revert configuration in case the VLAN script fails.

qinqon:
Closes #184

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -19,6 +19,7 @@ import (
 )
 
 const nmstateCommand = "nmstatectl"
+const vlanFilteringCommand = "vlan-filtering"
 
 var (
 	interfacesFilterGlob glob.Glob
@@ -47,12 +48,12 @@ func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
 	command := []string{bridgeName}
 	command = append(command, ports...)
 
-	cmd := exec.Command("vlan-filtering", command...)
+	cmd := exec.Command(vlanFilteringCommand, command...)
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
-		return "", fmt.Errorf("failed to execute %s: '%v', '%s', '%s'", "vlan-filtering", err, stdout.String(), stderr.String())
+		return "", fmt.Errorf("failed to execute %s: '%v', '%s', '%s'", vlanFilteringCommand, err, stdout.String(), stderr.String())
 	}
 	return stdout.String(), nil
 }
@@ -90,13 +91,22 @@ func set(state string) (string, error) {
 	//        https://nmstate.atlassian.net/browse/NMSTATE-247
 	retries := 2
 	for retries > 0 {
-		output, err = nmstatectl([]string{"set"}, state)
+		output, err = nmstatectl([]string{"set", "--no-commit"}, state)
 		if err == nil {
 			break
 		}
 		retries--
 	}
 	return output, err
+}
+
+func commit() (string, error) {
+	return nmstatectl([]string{"commit"}, "")
+}
+
+func rollback(cause error) error {
+	_, err := nmstatectl([]string{"rollback"}, "")
+	return fmt.Errorf("rollback cause: %v, rollback error: %v", cause, err)
 }
 
 func GetNodeNetworkState(client client.Client, nodeName string) (nmstatev1alpha1.NodeNetworkState, error) {
@@ -164,11 +174,9 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 	// we have to enforce it at the desiredState bridges and outbound ports
 	// they will be configured with vlan_filtering 1 and all the vlan id range
 	// set
-	// TODO: After implementing commit/rollack from nmstate we have to
-	//       rollback if vlanfiltering fails
 	bridgesUpWithPorts, err := getBridgesUp(nodeNetworkState.Spec.DesiredState)
 	if err != nil {
-		return "", fmt.Errorf("error retrieving up bridges from desired state: %v", err)
+		return "", rollback(fmt.Errorf("error retrieving up bridges from desired state"))
 	}
 
 	commandOutput := ""
@@ -176,10 +184,14 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 		outputVlanFiltering, err := applyVlanFiltering(bridge, ports)
 		commandOutput += fmt.Sprintf("bridge %s ports %v applyVlanFiltering command output: %s\n", bridge, ports, outputVlanFiltering)
 		if err != nil {
-			return commandOutput, err
+			return commandOutput, rollback(err)
 		}
 	}
 
+	_, err = commit()
+	if err != nil {
+		return commandOutput, rollback(err)
+	}
 	commandOutput += fmt.Sprintf("setOutput: %s \n", setOutput)
 	return commandOutput, nil
 }
--- a/test/e2e/rollback_test.go
+++ b/test/e2e/rollback_test.go
@@ -1,0 +1,36 @@
+package e2e
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("rollback", func() {
+	Context("when an error happens during state configuration", func() {
+		BeforeEach(func() {
+			By("Rename vlan-filtering to vlan-filtering.bak to force failure during state configuration")
+			runAtPods("sudo", "mv", "/usr/local/bin/vlan-filtering", "/usr/local/bin/vlan-filtering.bak")
+		})
+		AfterEach(func() {
+			By("Rename vlan-filtering.bak to vlan-filtering to leave it as it was")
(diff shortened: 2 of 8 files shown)


pkg/helper/client.go
```diff
@@ -80,6 +80,28 @@ func set(state string) (string, error) {
 	return stdout.String(), nil
 }
 
+func Rollback() (string, error) {
+	cmd := exec.Command(nmstateCommand, "rollback")
```
phoracek: Could you  create a helper function to run nmstatectl commands? This is being duplicated

pkg/helper/client.go
```diff
@@ -56,28 +56,42 @@ func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
...
+				fmt.Printf("failed to write input into stdin: %v\n", err)
+			}
+		}()
 
-	if err = cmd.Run(); err != nil {
-		return "", fmt.Errorf("failed to execute nmstate set: '%v' '%s' '%s'", err, stdout.String(), stderr.String())
 	}
-
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("failed to execute %s %s: '%v' '%v' '%s'", nmstateCommand, arguments, err, stdout.String(), stderr.String())
```
phoracek: this was originally '%v' "%s' %s', you changed it to '%v' **'%v'** '%s', which does not match passed stdin/out.

pkg/helper/client.go
```diff
@@ -56,28 +56,42 @@ func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
...
+	}
 	return stdout.String(), nil
+
+}
+
+func set(state string) (string, error) {
+	return nmstatectl([]string{"set", "--no-commit"}, state)
+}
+
+func Rollback() (string, error) {
```
phoracek: Why is Rollback and Commit public while set and show were private?

pkg/helper/client.go
```diff
@@ -56,28 +56,42 @@ func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
...
+	return nmstatectl([]string{"rollback"}, "")
+}
+
+func Commit() (string, error) {
+	return nmstatectl([]string{"commit"}, "")
```
phoracek: Don't we need to keep some rollback+commit ID to match the two? nmstatectl is stateless, there has to be a matching mechanism. Or do we just expect there is only a single user of the command?

pkg/helper/client.go
qinqon: We will use it later on for the lose connectivity to API I was expecting to do that outside of client.go, at controller, but we can do that check at ApplyDesiredState

pkg/helper/client.go
qinqon: I think it rollbacks the latest checkpoing, that's why checkpoin is optional in the command.

pkg/helper/client.go
phoracek: I see, still, please keep it private until we will need it elsewhere. Otherwise we may just forget about it

pkg/helper/client.go
phoracek: Gotcha. I think we should create named checkpoint. Makes less space for random issues IMO.

pkg/helper/client.go
qinqon: ack

pkg/helper/client.go
qinqon: Let make this work and pass, then we add checkpoints, what do you think ?

phoracek: Fair enough

SchSeba:
@qinqon can you please rebase

qinqon:
@SchSeba done, take a look now.

test/e2e/rollback_test.go
```diff
@@ -0,0 +1,54 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var _ = Describe("NodeNetworkState controller", func() {
```
phoracek: This description should be about the test module. I think "Rollback" alone is good.

test/e2e/rollback_test.go
```diff
@@ -0,0 +1,54 @@
...
+    type: linux-bridge
+    state: up
+    bridge:
+      options:
+        stp:
+          enabled: false
+      port: []
+`)
+	)
+	Context("when there is an error at vlan-filtering", func() {
```
phoracek: "when an error happens during state configuration"

We don't care what fails, do we?

test/e2e/rollback_test.go
```diff
@@ -0,0 +1,54 @@
...
+		BeforeEach(func() {
+			By("Rename vlan-filtering to force failure after nmstatectl set")
```
phoracek: We do not fail on `nmstatectl set` though. "to force failure during state configuration" may be more precise?

test/e2e/rollback_test.go
```diff
@@ -0,0 +1,54 @@
...
+			runAtPods("sudo", "mv", "/usr/local/bin/vlan-filtering", "/usr/local/bin/vlan-filtering.bak")
+		})
+		AfterEach(func() {
+			By("Correct the vlan-filtering script name")
```
phoracek: This tells me we are fixing something but you never said we are breaking the name. Would you either fix this `By` or the one in `BeforeEach`?

test/e2e/rollback_test.go
```diff
@@ -0,0 +1,54 @@
...
+			runAtPods("sudo", "mv", "/usr/local/bin/vlan-filtering.bak", "/usr/local/bin/vlan-filtering")
+			updateDesiredState(br1Absent)
+			for _, node := range nodes {
+				interfacesNameForNodeEventually(node).ShouldNot(ContainElement("br1"))
+			}
+		})
+		It("should rollback nmstate state to previous value", func() {
```
phoracek: "should rollback failed state configuration"

We should describe that we "fail the configuration" here, since it is not a part of the context description

test/e2e/utils.go
```diff
@@ -216,25 +216,57 @@ func deletePolicy(name string) {
...
 
+func kubectl(arguments ...string) (string, error) {
+	return run("./kubevirtci/cluster-up/kubectl.sh", arguments...)
+}
+
+func nmstatePods() ([]string, error) {
+	output, err := kubectl("get", "pod", "-n", namespace, "--no-headers=true", "-o", "custom-columns=:metadata.name")
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	names := strings.Split(output, "\n")
+	// Command outputs \n at the end of last element so new empty
```
phoracek: Could you use `TrimSpace` instead?

test/e2e/rollback_test.go
qinqon: Let's reworkd the By

qinqon:
@phoracek what need to get dropped, don't get it.

phoracek:
@qinqon there is a comment in client.go:176:
```
	// TODO: After implementing commit/rollack from nmstate we have to
	//       rollback if vlanfiltering fails
```

phoracek:
ci test please

qinqon:
ci test please

qinqon:
ci test please

pkg/helper/client.go
```diff
@@ -164,22 +173,27 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 	// we have to enforce it at the desiredState bridges and outbound ports
 	// they will be configured with vlan_filtering 1 and all the vlan id range
 	// set
-	// TODO: After implementing commit/rollack from nmstate we have to
-	//       rollback if vlanfiltering fails
 	bridgesUpWithPorts, err := getBridgesUp(nodeNetworkState.Spec.DesiredState)
 	if err != nil {
```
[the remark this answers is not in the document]
qinqon: Yep, let's wrap it.

qinqon:
This is how traces look now 

```json
{"level":"info","ts":1570104042.659638,"logger":"controller_nodenetworkstateconfiguration","msg":"Reconciling NodeNetworkStateConfiguration","Request.Namespace":"","Request.Name":"node01"}
{"level":"info","ts":1570104042.7493527,"logger":"controller_nodenetworkstatereport","msg":"Reconciling NodeNetworkState report","Request.Namespace":"","Request.Name":"node01"}
{"level":"info","ts":1570104043.1400118,"logger":"controller_nodenetworkstatereport","msg":"Reconciling NodeNetworkState report","Request.Namespace":"","Request.Name":"node01"}
{"level":"error","ts":1570104043.4481022,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"nodenetworkstateconfiguration-controller","request":"/node01","error":"error reconciling nodenetworkstate at desired state apply: rollback cause: failed to execute vlan-filtering: 'exec: \"vlan-filtering\": executable file not found in $PATH', '', '', rollback error: <nil>","stacktrace":"github.com/nmstate/kubernetes-nmstate/vendor/github.com/go-logr/zapr.(*zapLogger).Error
```

PR #196, opened 2019-10-01
Retry if nmstatectl fails

qinqon:
There is an issue at nmstate [1] that goes away if we just retry after failure

[1] https://nmstate.atlassian.net/browse/NMSTATE-247

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/helper/client.go
```diff
@@ -51,33 +52,60 @@ func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
...
+func set(state string) (string, error) {
+	output := ""
+	var err error = nil
+	// FIXME: Remove this retries after nmstate team fixes
+	//        https://nmstate.atlassian.net/browse/NMSTATE-247
+	retries := 2
+	for retries > 0 {
+		output, err = nmstatectl([]string{"set", "--no-commit"}, state)
+		if err == nil {
+			retries = 0
```
phoracek: use break instead

pkg/helper/client.go
```diff
@@ -51,33 +52,60 @@ func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
...
+		} else {
```
phoracek: with break, we won't need else, just add retries-- after the if block

qinqon:
ci test please

PR #198, opened 2019-10-02
Added waiting for cluster to be ready in makefile.

### LINKED ISSUE #195: Check ready condition at cluster-sync
After we do a "make cluster-sync" we have to check that nmstate is ready with kubectl wait.

HlinaCZ:
Resolves #195

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,11 +39,11 @@ export KUBEVIRT_NUM_SECONDARY_NICS ?= 2
 
 CLUSTER_DIR ?= kubevirtci/cluster-up/
 KUBECONFIG ?= kubevirtci/_ci-configs/$(KUBEVIRT_PROVIDER)/.kubeconfig
-KUBECTL ?= $(CLUSTER_DIR)/kubectl.sh
+export KUBECTL ?= $(CLUSTER_DIR)/kubectl.sh
 CLUSTER_UP ?= $(CLUSTER_DIR)/up.sh
 CLUSTER_DOWN ?= $(CLUSTER_DIR)/down.sh
 CLI ?= $(CLUSTER_DIR)/cli.sh
-SSH ?= $(CLUSTER_DIR)/ssh.sh
+export SSH ?= $(CLUSTER_DIR)/ssh.sh
 
 install_kubevirtci := hack/install-kubevirtci.sh
 local_handler_manifest = build/_output/handler.local.yaml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -146,6 +146,11 @@ cluster-sync-handler: cluster-sync-resources $(local_handler_manifest)
 	$(KUBECTL) apply -f deploy/crds/nmstate_v1alpha1_nodenetworkconfigurationpolicy_crd.yaml
 	$(KUBECTL) delete --ignore-not-found -f $(local_handler_manifest)
 	$(KUBECTL) create -f $(local_handler_manifest)
+	until [[ \
```
qinqon: Can't we use kubectl wait here ? 

Also desiredNumberScheduler can be retrieved out of the loop

Makefile
```diff
@@ -146,6 +146,11 @@ cluster-sync-handler: cluster-sync-resources $(local_handler_manifest)
...
+		$$($(KUBECTL) get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.desiredNumberScheduled}') \
+		-eq \
+		$$($(KUBECTL) get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.numberAvailable}') \
```
qinqon: Limite the number of checks we do by time or by cardinality, we don't want an infinite loop here.

Makefile
phoracek: We can't, `wait` can be used only for conditions. The DS readiness is not exposed in any condition (maybe if we digged into replica-set...).

We can retrieve it outside the loop, but we cannot store any variables in make target, can we.

qinqon: Ack about wait :-/

About variable, let's use this PR to move bash code of cluster-sync to hack/cluster-sync.sh, let's export KUBECTL as an environment variable.

hack/cluster-sync.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -ex
+
+${KUBECTL} apply -f deploy/crds/nmstate_v1alpha1_nodenetworkstate_crd.yaml
+${KUBECTL} apply -f deploy/crds/nmstate_v1alpha1_nodenetworkconfigurationpolicy_crd.yaml
+${KUBECTL} delete --ignore-not-found -f ${local_handler_manifest}
+${KUBECTL} create -f ${local_handler_manifest}
+
+desiredNumberScheduled="$(${KUBECTL} get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.desiredNumberScheduled}')"
+for i in {60..0}; do
```
phoracek: let's do 300..0 and set the sleep to 1 second, to save some calculation in code-readers head :)

qinqon:
Also for me I have to externally set KUBEVIRT_PROVIDER before calling make cluster-sync to make it work, can we fix that here ?

phoracek:
@qinqon how is it related to this PR?

qinqon:
@phoracek we can do at follow up.

hack/cluster-sync.sh
```diff
@@ -0,0 +1,21 @@
...
+		echo "nmstate-handler DS is ready"
+		break
+	fi
+
+	if [ $i -eq 0 ]; then
+		echo "nmstate-handler DS haven't turned ready within the given timeout"
+		exit 1
+	fi
+
+	sleep 5;
```
phoracek: drop ;

phoracek:
@qinqon I think I understand what you meant now. We have to export the PROVIDER to cluster-sync to make this PR work. I thought we are talking about the issue that we gotta export the k8s.1.14, even when working with the default.

PR #200, opened 2019-10-02
Reset dersired state after each test

qinqon:


test/e2e/simple_bridge_and_bond.go
```diff
@@ -215,6 +219,7 @@ var _ = Describe("NodeNetworkState", func() {
 				for _, node := range nodes {
 					interfacesNameForNode(node).ShouldNot(ContainElement("bond1"))
 				}
+				resetDesiredStateForNodes()
```
phoracek: Can't we do it in module-wide (or even better, package-wide) AfterEach?

qinqon: well we don't control what afterEach come last..., prefer to do this way to be sure that this is the last thing is called.

phoracek: Fine with me. Would you just add a TODO to resetDesiredStateForNodes mentioning that we should drop it once Policy cleans up the state?

phoracek:
ci test please

phoracek:
ci test please

PR #202, opened 2019-10-04
Ping external ip to verify connectivity

qinqon:

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -14,6 +14,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	yaml "sigs.k8s.io/yaml"
 
+	"github.com/tidwall/gjson"
+
 	"github.com/gobwas/glob"
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
@@ -159,6 +161,34 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
 	return nil
 }
 
+func ping(target string) (string, error) {
+	cmd := exec.Command("ping", "-c", "3", target)
+	var outputBuffer bytes.Buffer
+	cmd.Stdout = &outputBuffer
+	cmd.Stderr = &outputBuffer
+	return outputBuffer.String(), cmd.Run()
+}
+
+func defaultGw() (string, error) {
+	observedStateRaw, err := show()
+	if err != nil {
+		return "", fmt.Errorf("error running nmstatectl show: %v", err)
+	}
+
+	currentState, err := yaml.YAMLToJSON([]byte(observedStateRaw))
+	if err != nil {
+		return "", fmt.Errorf("Impossible to convert current state to JSON")
+	}
+
+	defaultGw := gjson.ParseBytes([]byte(currentState)).
+		Get("routes.running.#(destination==\"0.0.0.0/0\").next-hop-address").String()
+	if defaultGw == "" {
+		return "", fmt.Errorf("Impossible to retrieve default gw")
+	}
+
+	return defaultGw, nil
+}
+
 func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (string, error) {
 	desiredState := string(nodeNetworkState.Spec.DesiredState)
 	if len(desiredState) == 0 {
@@ -188,10 +218,21 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 		}
 	}
 
+	defaultGw, err := defaultGw()
+	if err != nil {
+		return commandOutput, rollback(err)
+	}
+
+	pingOutput, err := ping(defaultGw)
+	if err != nil {
+		return pingOutput, rollback(fmt.Errorf("error pinging external address after network reconfiguration: %v", err))
+	}
+
 	_, err = commit()
 	if err != nil {
 		return commandOutput, rollback(err)
 	}
+
 	commandOutput += fmt.Sprintf("setOutput: %s \n", setOutput)
 	return commandOutput, nil
 }
--- a/test/e2e/rollback_test.go
+++ b/test/e2e/rollback_test.go
@@ -5,7 +5,29 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
+
+func badDefaultGw(address string, nic string) nmstatev1alpha1.State {
+	return nmstatev1alpha1.State(fmt.Sprintf(`interfaces:
+  - name: %s
+    type: ethernet
+    state: up
+    ipv4:
+      dhcp: false
+      enabled: true
+      address:
+        - ip: %s
+          prefix-length: 24
+routes:
+  config:
+    - destination: 0.0.0.0/0
+      metric: 150
+      next-hop-address: 192.0.2.1
+      next-hop-interface: %s
+`, nic, address, nic))
+}
 
 var _ = Describe("rollback", func() {
 	Context("when an error happens during state configuration", func() {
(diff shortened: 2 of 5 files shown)


qinqon:
I am installing "iputils" but we may want to use golang tools like https://godoc.org/github.com/sparrc/go-ping both options are going to increase the size of Dockerfile

pkg/helper/client.go
```diff
@@ -188,10 +188,16 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
 		}
 	}
 
+	err = exec.Command("ping", "-c", "3", "8.8.8.8").Run()
```
phoracek: Please hide this in a helper function, pass only the IP address.

Also, the IP should be read from our config map.

qinqon: Yep that was the next step, pushed it to see how it goes at stdci.

deploy/operator.yaml
```diff
@@ -55,6 +55,11 @@ spec:
                 configMapKeyRef:
                   name: nmstate-config
                   key: interfaces_filter
+            - name: CONNECTIVITY_CHECK_ENDPOINT
```
phoracek: If I was a user I would have no idea what the endpoint means. Maybe connectivity_check_ip_target?

phoracek: But since it can be domain name, maybe just connectivity_check_target?

qinqon: ack, is good name too.

phoracek:
@SchSeba the API ping would come later. We want to start with this since we work on single node environment, so access to API server is not that hard. However y the idea with default gateway sounds great. Less configuration variables.

qinqon:
@SchSeba @phoracek I like usin the default gateway as the connectivity target, but want to keep the environment variable for testing purposes is that ok ?

qinqon:
@phoracek @SchSeba in case we can modify the default gw with a bad value, we don't need the config to test it.

SchSeba:
@qinqon for a test you can do it.

just run the `ip route add` command on the host to change the default one. (you need to remove the correct one and the create a new one)

Then just return the one you deleted on the `afterEach`

phoracek:
@qinqon @SchSeba we can configure the route via nmstate ;) https://nmstate.github.io/examples.html#route

qinqon:
@SchSeba my idea is to try to do first using nmstate, if this renders impossible we can hack it at node.

pkg/helper/client.go
```diff
@@ -159,12 +167,26 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
 	return nil
 }
 
+func ping(target string) error {
```
phoracek: It may be interesting to log the output of the ping too. Don't you think?

qinqon:
@SchSeba @phoracek looks like we can change default gw with nmstate but they have an issue and it's only possible changing eth0 to static ip address, but at least is working.

pkg/helper/client.go
qinqon: I want to see CI passing and then will cover this.

pkg/helper/client.go
```diff
@@ -159,6 +161,27 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
...
+	return exec.Command("ping", "-c", "3", target).Run()
+}
+
+func defaultGw() (string, error) {
+	observedStateRaw, err := show()
+	if err != nil {
+		return "", fmt.Errorf("error running nmstatectl show: %v", err)
+	}
+	currentState, err := yaml.YAMLToJSON([]byte(observedStateRaw))
```
phoracek: please add empty line before this

pkg/helper/client.go
```diff
@@ -159,6 +161,27 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
...
+	if err != nil {
+		return "", fmt.Errorf("Impossible to convert current state to JSON")
+	}
+	defaultGw := gjson.ParseBytes([]byte(currentState)).
```
phoracek: ditto

pkg/helper/client.go
```diff
@@ -159,6 +161,27 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
...
+		Get("routes.running.#(destination==\"0.0.0.0/0\").next-hop-address").String()
+	if defaultGw == "" {
+		return "", fmt.Errorf("Impossible to retrieve default gw")
+	}
+	return defaultGw, nil
```
phoracek: ditto

pkg/helper/client.go
```diff
@@ -188,10 +211,21 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+	err = ping(defaultGw)
+	if err != nil {
+		return commandOutput, rollback(fmt.Errorf("error pinging external address after network reconfiguration"))
+	}
+
 	_, err = commit()
 	if err != nil {
 		return commandOutput, rollback(err)
 	}
+
```
phoracek: you read my mind!

test/e2e/rollback_test.go
```diff
@@ -5,8 +5,31 @@ import (
...
+      address:
+        - ip: %s
+          prefix-length: 24
+routes:
+  config:
+    - destination: 0.0.0.0/0
+      metric: 150
+      next-hop-address: 192.0.2.1
+      next-hop-interface: %s
+      table-id: 254
```
phoracek: do we need this?

test/e2e/rollback_test.go
```diff
@@ -26,11 +49,43 @@ var _ = Describe("rollback", func() {
...
+				updateDesiredStateAtNode(node, badDefaultGw(address, "eth0"))
+			}
+		})
+		AfterEach(func() {
+			By("Clean up desired state")
+			resetDesiredStateForNodes()
+		})
+		It("should rollback to a good gw configuration", func() {
+			for _, node := range nodes {
+
```
phoracek: please drop this line

test/e2e/rollback_test.go
```diff
@@ -26,11 +49,43 @@ var _ = Describe("rollback", func() {
...
+				By("Check that desiredState is applied")
+				Eventually(func() bool {
+					return dhcpFlag(node, "eth0")
+				}, ReadTimeout, ReadInterval).Should(BeFalse())
+
+				By(fmt.Sprintf("Check that %s is rolled back", "eth0"))
+				Eventually(func() bool {
+					return dhcpFlag(node, "eth0")
+				}, ReadTimeout, ReadInterval).Should(BeTrue())
+
```
phoracek: please drop this line. it is already indented, we don't need to divide it even more

test/e2e/rollback_test.go
qinqon: Nope, removing table-id

test/e2e/rollback_test.go
```diff
@@ -26,11 +48,41 @@ var _ = Describe("rollback", func() {
...
+			resetDesiredStateForNodes()
+		})
+		It("should rollback to a good gw configuration", func() {
+			for _, node := range nodes {
+				By("Check that desiredState is applied")
+				Eventually(func() bool {
+					return dhcpFlag(node, "eth0")
+				}, ReadTimeout, ReadInterval).Should(BeFalse())
+
+				By(fmt.Sprintf("Check that %s is rolled back", "eth0"))
```
phoracek: Sprintf is not really needed, you duplicate the string anyway. I think it would improve readibility if you just wrote it directly

qinqon: Did it for the future okd4 provider that has parameterize nic names.

qinqon: Bah let's do it then.

pkg/helper/client.go
```diff
@@ -188,10 +219,22 @@ func ApplyDesiredState(nodeNetworkState *nmstatev1alpha1.NodeNetworkState) (stri
...
+	defaultGw, err := defaultGw()
+	if err != nil {
+		return commandOutput, rollback(err)
+	}
+
+	pingOutput, err := ping(defaultGw)
+	if err != nil {
+		return pingOutput, rollback(fmt.Errorf("error pinging external address after network reconfiguration: %v", err))
+	}
+	commandOutput += fmt.Sprintf("pingOutput: %s \n", pingOutput)
```
phoracek: We don't care in case it succeeded.

PR #214, opened 2019-10-10
Stop re-concile after rollback detected.

qinqon:
In case of rollback at re-concile, manual intervention is needed to fix the desiredState, also
a error log  is printed.


pkg/controller/nodenetworkstateconfiguration/nodenetworkstateconfiguration_controller.go
```diff
@@ -200,14 +201,19 @@ func (r *ReconcileNodeNetworkStateConfiguration) Reconcile(request reconcile.Req
...
-		errmsg := fmt.Errorf("error reconciling nodenetworkstate at desired state apply: %v", err)
+		errmsg := fmt.Errorf("error reconciling nodenetworkstate at desired state apply: %s, %v", nmstateOutput, err)
 
 		retryErr := r.setCondition(false, errmsg.Error(), request.NamespacedName)
 		if retryErr != nil {
 			reqLogger.Error(retryErr, "Failing condition update failed while reporting error: %v", errmsg)
 		}
 
-		return reconcile.Result{}, errmsg
+		if strings.Contains(errmsg.Error(), "rollback cause:") {
```
phoracek: Why only on rollback? We don't want to re-reconcile even if it failed validation of the scheme. Or even if it succeeded with rollback, we should not get into a loop of rollbacking and reconfigurations. Please tell me if I'm missing something. Is there a case where we would want to retry the configuration before user manually edits the Policy?

qinqon: Let's fail always, I don't find a failure scenario where a re-concile would fix it.

phoracek: Then please also add the "manual intervention" message to the condition.

pkg/helper/client.go
```diff
@@ -161,12 +163,20 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
...
 
-func ping(target string) (string, error) {
-	cmd := exec.Command("ping", "-c", "3", target)
-	var outputBuffer bytes.Buffer
-	cmd.Stdout = &outputBuffer
-	cmd.Stderr = &outputBuffer
-	return outputBuffer.String(), cmd.Run()
+func ping(target string, timeout int) (string, error) {
+	output := ""
+	return output, wait.PollImmediate(1*time.Second, 50*time.Second, func() (bool, error) {
```
phoracek: You should use the timeout here. Also, please make the timeout of type `time.Duration`.

pkg/helper/client.go
```diff
@@ -161,12 +163,20 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
...
+		cmd := exec.Command("ping", "-c", "1", target)
```
phoracek: You already implemented Eventually, maybe we can extend it with Consistently here inside?

pkg/controller/nodenetworkstateconfiguration/nodenetworkstateconfiguration_controller.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
```
phoracek: Please update the commit message and PR description to reflect all the changes you are doing here.

pkg/helper/client.go
qinqon: Let's check if we really needed, adding consistently is adding complexity that you have to debug in case of error, let me add a TODO

pkg/controller/nodenetworkstateconfiguration/nodenetworkstateconfiguration_controller.go
qinqon: Ack, I am going to stop re-concile always, as you suggested.

qinqon:
ci test please

test/e2e/rollback_test.go
```diff
@@ -30,6 +34,13 @@ routes:
 }
 
 var _ = Describe("rollback", func() {
+	BeforeEach(func() {
+		By("Check NodeNetworkState is not at failing condition")
```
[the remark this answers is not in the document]
qinqon: We can refactor e2e tests with conditions check at a follow up PR, what do you think ?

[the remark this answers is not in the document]
phoracek: fair enough

phoracek:
Rebase is not affecting the tests.

PR #216, opened 2019-10-11
Call CLI ports from withing target to read KUBEVIRT_PROVIDER

qinqon:
If make cluster-sync is call without KUBEVIRT_PROVIDER sync is faling
at getting local registry ports, looks like Makefile $(shell) does not
export environment variables to not pollute the sell.

It also moves cluster-sync.sh to cluster-sync-handler.sh since is what it does.


qinqon:
@phoracek is implicit, if you environment variables exported before $(shell  foo.sh) are not seen inside foo.sh, so we just do $$(./foo.sh) to run directly at bash.

PR #217, opened 2019-10-14
Retry on conflict at network configuration report

qinqon:
We have two controllers that update Status of NodeNetworkState, conflicts
where resolved re-conconcile, but that takes too much time, also it produces
a confusing error, with this commit in case of a conflict on NodeNetworKState it
retries.


pkg/controller/nodenetworkstatereport/nodenetworkstatereport_controller.go
```diff
@@ -104,25 +105,28 @@ type ReconcileNodeNetworkStateReport struct {
 func (r *ReconcileNodeNetworkStateReport) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.V(1).Info("Reconciling NodeNetworkState report")
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
phoracek: This retries only when we have API write conflict, but returns IsNotFound error as soon as it happens? Neat.

PR #218, opened 2019-10-14
Add sleep between nmstaectl set retries

qinqon:
Sometimes NM takes some time to rollback previous failure.


qinqon:
ci test please

PR #220, opened 2019-10-14
Add simple tests for openvswitch bridge

fedepaol:
This PR:
- installs openvswitch on the cluster
- extends the installation of nm to all the nodes instead of only the first one
- adds simple tests for openvswitch bridge
- adds docs describing how to configure openvswitch


ovirt-infra:

test/e2e/simple_ovs_bridge.go
```diff
@@ -0,0 +1,50 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("NodeNetworkState", func() {
+	var ()
```
phoracek: This should be dropped (and from the original test too).

test/e2e/simple_ovs_bridge.go
```diff
@@ -0,0 +1,50 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("NodeNetworkState", func() {
```
phoracek: Please describe it as "Simple OVS bridge"

test/e2e/simple_ovs_bridge.go
```diff
@@ -0,0 +1,50 @@
...
+	var ()
+	Context("when desiredState is configured", func() {
+		Context("with an ovs bridge up with no ports", func() {
```
phoracek: It is questionable whether this makes sense at all. OVS bridge without cannot be used for anything. Please drop the test.

test/e2e/simple_ovs_bridge.go
```diff
@@ -0,0 +1,50 @@
...
+			It("should have the ovs bridge at currentState", func() {
+				for _, node := range nodes {
+					interfacesForNode(node).Should(ContainElement(SatisfyAll(
+						HaveKeyWithValue("name", bridge1),
+						HaveKeyWithValue("type", "ovs-bridge"),
+						HaveKeyWithValue("state", "up"),
+					)))
+				}
+			})
+		})
```
phoracek: Please also run a test with internal OVS interface. That is needed if someone wants to set an IP address on the bridge. I would say it is a basic scenario. https://github.com/nmstate/nmstate/blob/master/examples/ovsbridge_create.yml

test/e2e/simple_ovs_bridge.go
```diff
@@ -0,0 +1,50 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("NodeNetworkState", func() {
+	var ()
+	Context("when desiredState is configured", func() {
+		Context("with an ovs bridge up with no ports", func() {
+			BeforeEach(func() {
+				updateDesiredState(ovsBrUpNoPorts(bridge1))
```
phoracek: We should specify desired config only via NodeNetworkConfigurationPolicy. Setting those via NodeNetworkState will be likely deprecated (it was never supported for users, but we may drop it even from our internal implementation). See https://github.com/nmstate/kubernetes-nmstate/blob/master/test/e2e/default_bridged_network_test.go#L68

docs/user-guide-policy-configure-ovs-bridge.md
```diff
@@ -0,0 +1,86 @@
+# Tutorial: Create a Ovs Bridge and Connect it to a Node Interface
```
phoracek: s/Ovs/Open vSwitch/

hack/install-ovs.sh
```diff
@@ -11,8 +11,11 @@ function install_ovs_on_node() {
...
+if [[ "$KUBEVIRT_PROVIDER" =~  k8s- ]]; then
+    echo 'Installing Open vSwitch on nodes'
 
-for node in $($kubectl get nodes --no-headers | awk '{print $1}'); do
-    install_ovs_on_node "$node"
-done
+    for node in $($kubectl get nodes --no-headers | awk '{print $1}'); do
+        install_ovs_on_node "$node"
+    done
+fi
```
phoracek: please add a newline

docs/user-guide-policy-configure-ovs-bridge.md
phoracek: Please add reference to https://github.com/nmstate/kubernetes-nmstate#deployment-and-usage and https://github.com/nmstate/kubernetes-nmstate/blob/master/docs/user-guide.md

phoracek: Attach host NIC to it

phoracek: Eeeeh, then let it be and we can fix it later everywhere. There is already some overlap between those two guides. I want to move the node-selecting out in the future. And provide more use-cases info and background on these guides.

hack/install-nm.sh
```diff
@@ -1,13 +1,18 @@
 #!/bin/bash -e
 
-# TODO: Iterate all the nodes
+ssh=kubevirtci/cluster-up/ssh.sh
```
qinqon: use $SSH and $KUBECTL they are exported at env variables in the Makefile

hack/install-nm.sh
```diff
@@ -1,13 +1,18 @@
...
+    node=$1
+    $ssh $node sudo -- yum install -y NetworkManager NetworkManager-ovs
+    $ssh $node sudo -- systemctl daemon-reload
+    $ssh $node sudo -- systemctl restart NetworkManager
+    echo "Check NetworkManager is working fine on node $node"
+    $ssh $node -- nmcli device show > /dev/null
+}
 
-echo 'Install NetworkManager on the node'
-docker exec $node01_id ssh.sh sudo yum install -y NetworkManager
```
qinqon: I am happy we can remove this, there were a kubecirtci cli bug that was not blocking until command end.

hack/install-ovs.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
+
+ssh=kubevirtci/cluster-up/ssh.sh
```
qinqon: ditto

test/e2e/simple_ovs_bridge.go
qinqon: I would suggest leave this like linux-bridge and do a follow up PR remove NodeNetworkState desiredState use all together from e2e tests.

fedepaol:
> First pass, also as extra mile can we rename https://github.com/nmstate/kubernetes-nmstate/blob/master/pkg/helper/bridges.go#L13 as "linuxBridgesUp" ? at golang getters don't
> have the "get" prefix.

To be honest, I'd be happy to split it as a separate package since

fedepaol:
Wrong button

test/e2e/simple_ovs_bridge.go
fedepaol: Ok, as per @qinqon comment I'll leave it with state so we can change all of them in a follow-up pr.

fedepaol:
@phoracek @qinqon I think this is good to go

Makefile
```diff
@@ -31,6 +31,10 @@ E2E_TEST_ARGS += -test.v -test.timeout=40m -ginkgo.v -ginkgo.slowSpecThreshold=6
 ifdef E2E_TEST_FOCUS
 	E2E_TEST_ARGS +=  -ginkgo.focus $(E2E_TEST_FOCUS)
 endif
+# skip ovs tests for non k8s providers
+ifeq (,$(findstring k8s-,$(KUBEVIRT_PROVIDER)))
+	E2E_TEST_SKIP += .*OVS.*
```
phoracek: You sure this works? On other places we extend the flag with `+=  FOO` (two spaces, one is splitting from `+=` another adding extra space to the flag.

docs/user-guide-policy-configure-ovs-bridge.md
```diff
@@ -0,0 +1,86 @@
...
+## Requirements
+
+Before we start, please make sure that you have your Kubernetes/OpenShift
+cluster ready. In order to do that, you can follow the guides of deployment on
+[local cluster](deployment-local-cluster.md) or your
+[arbitrary cluster](deployment-arbitrary-cluster.md).
+
+You must also be sure that [OpenVSwitch](https://www.openvswitch.org/) and [NetworkManager OpenVSwitch plugin](https://developer.gnome.org/NetworkManager/stable/nm-openvswitch.html) are installed in the nodes.
+
+Please also note that on OpenShift `openvswitch` is already installed as a daemon, so only `nm-openvswitch` is required.
```
phoracek: This is true only for OpenShift. On Kubernetes, we don't know.

docs/user-guide-policy-configure-ovs-bridge.md
```diff
@@ -0,0 +1,86 @@
...
+metadata:
+  name: br1-eth1-policy
+spec:
+  desiredState:
+    interfaces:
+      - name: br1
+        type: ovs-bridge
+        state: absent
+EOF
+```
```
phoracek: Please add section explaining that while the previous example may be enouh/useful to provide bridge for Pods via OVS CNI, it won't have any IP endpoint in the system. To do so, internal iface has to be created (and show how).

README.md
```diff
@@ -98,8 +98,9 @@ After that, you can follow one of the following guides that will guide you
 through node state reporting and interface configuration.
 
 - [Report node network state](docs/user-guide-state-reporting.md)
-- [Connect an Linux bridge to a node interface](docs/user-guide-policy-configure-linux-bridge.md)
-- [Configure Linux bonding interface](docs/user-guide-policy-configure-linux-bond.md)
+- [Connect a Linux bridge to a node interface](docs/user-guide-policy-configure-linux-bridge.md)
+- [Configure a Open vSwitch bridge to a node interface](docs/user-guide-policy-configure-ovs-bridge.md)
```
phoracek: an Open

hack/install-ovs.sh
```diff
@@ -0,0 +1,18 @@
...
+
+
+function install_ovs_on_node() {
+    node=$1
+    $SSH $node -- sudo yum install -y http://cbs.centos.org/kojifiles/packages/openvswitch/2.9.2/1.el7/x86_64/openvswitch-2.9.2-1.el7.x86_64.rpm http://cbs.centos.org/kojifiles/packages/openvswitch/2.9.2/1.el7/x86_64/openvswitch-devel-2.9.2-1.el7.x86_64.rpm http://cbs.centos.org/kojifiles/packages/dpdk/17.11/3.el7/x86_64/dpdk-17.11-3.el7.x86_64.rpm
+    $SSH $node -- sudo systemctl daemon-reload
+    $SSH $node -- sudo systemctl restart openvswitch
+}
+
+# we currently skip ovs for non k8s providers.
```
phoracek: Please say that we test it only on Kubernetes providers (stay positive Federico)

test/e2e/simple_ovs_bridge.go
```diff
@@ -0,0 +1,60 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Simple OVS bridge", func() {
+	Context("when desiredState is configured", func() {
```
phoracek: This context does not do anything. Could you remove it and add the description to underlying Contexts?

Makefile
fedepaol: I tried it, by replacing k8s with blah and it worked as expected.

docs/user-guide-policy-configure-ovs-bridge.md
fedepaol: Isn't what the comment says?

phoracek: Tired me. Nevermind.

docs/user-guide-policy-configure-ovs-bridge.md
fedepaol: I would not speak about the pod connectivity but keep talking about the host networking. After all, this may be used just to configure the host. Using the ovs with the pod is the next step.

phoracek: Fair enough, but please explain the difference between the two.

phoracek:
ci add to whitelist

phoracek:
ci test please

docs/user-guide-policy-configure-ovs-bridge.md
```diff
@@ -0,0 +1,123 @@
...
+          options:
+            stp: true
+          port:
+            - name: eth1
+              type: system
+            - name: ovs0
+              type: internal
+EOF
+```
+
```
phoracek: So many empty lines! Please use just a single one.

docs/user-guide-policy-configure-ovs-bridge.md
```diff
@@ -0,0 +1,123 @@
...
+        state: up
+        bridge:
+          options:
+            stp: false
+          port:
+            - name: eth1
+EOF
+```
+
+By doing this though, we will be able to create an Open vSwitch bridge on the host but the host may loose connectivity since its nic is now connected to the bridge.
```
phoracek: Let's not mention loss of connectivity. That is very specific issue. You may find some inspiration here https://github.com/nmstate/nmstate.github.io/blob/master/devel/api.md#ovsbridgeport_type (although it is not the best worded description. Maybe the blog post you shared with us would work better?

fedepaol: @phoracek tried to rephrase, let me know if this version sounds better.

fedepaol:
> okd4.1 is not skipping OVS tests
> 
> [Fail] Simple OVS bridge when desiredState is configured with an ovs bridge up [It] should have the ovs bridge at currentState
> /home/jenkins/workspace/nmstate_kubernetes-nmstate_standard-check-pr/go/src/github.com/nmstate/kubernetes-nmstate/test/e2e/simple_ovs_bridge.go:22

Ouch, will fix it

docs/user-guide-policy-configure-ovs-bridge.md
phoracek: It does not. I would not mention connectivity at all. It is just about setting an IP. If it is not for the default interface, connectivity is only distracting.

fedepaol:
I _think_ the failures are not related to the pr?

phoracek:
@fedepaol no they are not

PR #228, opened 2019-10-17
configure VLAN on the bridge device

phoracek:
Without explicit configuration of bridge's VID, it is not possible
to use it for IP connectivity of the host. This patch sets default
VID on the bridge.

### DIFF
--- a/build/bin/vlan-filtering
+++ b/build/bin/vlan-filtering
@@ -8,6 +8,8 @@ vlan_range=2-4094
 
 ip link set dev $bridge type bridge vlan_filtering 1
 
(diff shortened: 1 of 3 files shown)


phoracek:
@qinqon @fedepaol please take a look

phoracek:
@qinqon looks like we are good here. The default bridge test on OKD has to be changed to use the `eno*` interface instead of `eth0`, but that should be done in a separate PR.

qinqon:
@phoracek thanks thanks thanks !!!

qinqon:
Why did I close It instead of merge it ?

PR #239, opened 2019-10-26
Move set to nncp

phoracek:


phoracek:
@qinqon @fedepaol would you please take a look? Sorry about the size, but moving of the conditions did it.

PR #246, opened 2019-11-08
move to go modules

qinqon:


cmd/manager/main.go
```diff
@@ -107,10 +114,32 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Create Service object to expose the metrics port.
-	_, err = metrics.ExposeMetricsPort(ctx, metricsPort)
+	if err = serveCRMetrics(cfg); err != nil {
```
phoracek: Why do we need these in this PR?

hack/install-github-release.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -xe
```
phoracek: How is this change related to the rest? Would you please split the PR into multiple commits (or multiple PRs, but won't fight for it). Moving to newer operator-sdk, go.mod, this little script.

cmd/manager/main.go
qinqon: operator-sdk 0.12.0 has this at generated code, I have updated it to make it compile.

hack/install-github-release.sh
qinqon: We cannot, proper support for operator-sdk is done at 0.12.0 so it have to be at the same PR, although I have to remove this script.

automation/check-patch.e2e-k8s.sh
```diff
@@ -15,7 +15,6 @@ main() {
     export KUBEVIRT_PROVIDER='k8s-1.14.6'
 
     source automation/check-patch.e2e.setup.sh
-    cd ${TMP_PROJECT_PATH}
```
phoracek: Why do you remove this? It is a part of the kubevirtci scripting. It makes sure that you can execute the automation/ test script on any environment and it will create a temporary project directory with everything working, even if you did not have a working GOPATH before.

automation/check-patch.e2e.setup.sh
```diff
@@ -1,27 +1,9 @@
-# Prepare environment for kubernetes-nmstate end to end testing. This includes temporary Go paths and binaries.
+# Prepare environment for kubernetes-nmstate end to end testing.
 #
-# source automation/check-patch.e2e.setup.sh
-# cd ${TMP_PROJECT_PATH}
 
-echo 'Setup Go paths'
-export GOROOT=/tmp/kubernetes-nmstate/go/root
```
phoracek: With this the script would kill my Go environment if executed locally. On top of that, it would not work in an environment where Go is not set up yet. The point of these scripts was to allow it to run anywhere without dependencies.

automation/check-patch.e2e-k8s.sh
qinqon: Adding them it back.

automation/check-patch.e2e.setup.sh
qinqon: going back to them

hack/install-github-release.sh
phoracek: You promised to remove this :(

Makefile
```diff
@@ -78,7 +81,7 @@ test/unit: $(GINKGO)
 	INTERFACES_FILTER="" NODE_NAME=node01 $(GINKGO) $(unit_test_args) ./pkg/
 
 test/e2e: $(OPERATOR_SDK)
-	$(OPERATOR_SDK) test local ./test/e2e \
+	unset GOFLAGS && $(OPERATOR_SDK) test local ./test/e2e \
```
phoracek: What is the unset good for?

qinqon: operator-sdk is adding another mod=vendor so it fails.

phoracek: Please explain it in an inline comment for our future selves.

phoracek:
ci test please

cmd/manager/main.go
phoracek: Feels bad to have metrics that we never even tried. But it is not a regression compared to the original code, just complaining.

hack/install-github-release.sh
qinqon: Broken promises, let's remove it :-)

cmd/manager/main.go
qinqon: Let's try to simplify not used stuff at follow-up PRs

PR #247, opened 2019-11-09
add list of conditions to policy

phoracek:
This list will be used to keep configuration status per each node in Policy's State.

That will be a temporary solution until we introduce Enactment object.

The benefit of this is, that we keep the status of configuration in the object that is used to trigger the configuration. In other words, we decouple reporting and configuration.

Also add condition list helpers. These helpers will be used to A) set configuration condition for a specific node and B) to obtain condition of certain type for a given node.

Finally, this PR finally drops Priority from the Policy object Spec, since there will be no ordering employed.


phoracek:
@qinqon @oshoval could you please review this?

This is pushed to a side branch "decouple_configuration_and_reporting". Let's use this branch to move configuration from NNS to NNCP. Once that's done, we can merge this whole branch into master and continue on the rest of the features in parallel.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -30,36 +48,65 @@ type NodeNetworkConfigurationPolicySpec struct {
...
+type NodeInfoList []NodeInfo
 
-// NodeNetworkConfigurationPolicy is the Schema for the nodenetworkconfigurationpolicies API
-// +k8s:openapi-gen=true
-// +kubebuilder:subresource:status
-// +kubebuilder:resource:path=nodenetworkconfigurationpolicies,shortName=nncp
-type NodeNetworkConfigurationPolicy struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+type NodeInfo struct {
```
qinqon: What about start to call this NodeNetworkConfigurationEnactment so we start to introduce the Enactment concept ? then we move it to it's own CRD.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ = Describe("Node info list", func() {
+	Context("is empty", func() {
+		originalNodeInfoList := NodeInfoList{}
```
qinqon: For "is empty" having an originalNodeInfoList is overkill

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+		var newNodeInfoList NodeInfoList
+
+		BeforeEach(func() {
+			newNodeInfoList = originalNodeInfoList.DeepCopy()
```
qinqon: Just initizlize it.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+			addedConditionStatus := corev1.ConditionTrue
+			addedConditionReason := ConditionReason("bar")
+			addedConditionMessage := "baz"
+
+			BeforeEach(func() {
+				newNodeInfoList.SetCondition(addedNodeName, addedConditionType, addedConditionStatus, addedConditionReason, addedConditionMessage)
+			})
+
+			It("should extend the list", func() {
+				Expect(newNodeInfoList).To(HaveLen(len(originalNodeInfoList) + 1))
```
qinqon: len(1) originalNodeInfoList is overkill here.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+			})
+
+			It("should add expected node entry to the list", func() {
+				Expect(newNodeInfoList[0].Name).To(Equal(addedNodeName))
+			})
+
+			It("should be able to find the added condition", func() {
+				addedCondition := newNodeInfoList.FindCondition(addedNodeName, addedConditionType)
+				Expect(addedCondition).NotTo(BeNil())
```
qinqon: Check that it's the correct one.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+				newNodeInfoList.SetCondition(existingNodeName, addedConditionType, addedConditionStatus, addedConditionReason, addedConditionMessage)
+			})
+
+			It("should not add a new node entry to the list", func() {
+				Expect(newNodeInfoList).To(HaveLen(len(originalNodeInfoList)))
+			})
+
+			It("should contain both the old and new conditions", func() {
+				existingCondition := newNodeInfoList.FindCondition(existingNodeName, existingConditionType)
+				Expect(existingCondition).NotTo(BeNil())
```
qinqon: The rest of fields ?

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+				addedCondition := newNodeInfoList.FindCondition(existingNodeName, addedConditionType)
```
qinqon: ditto

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+				newNodeInfoList.SetCondition(addedNodeName, addedConditionType, addedConditionStatus, addedConditionReason, addedConditionMessage)
+			})
+
+			It("should extend the list", func() {
+				Expect(newNodeInfoList).To(HaveLen(len(originalNodeInfoList) + 1))
+			})
+
+			It("should contain both the old and the new conditions", func() {
+				existingCondition := newNodeInfoList.FindCondition(existingNodeName, existingConditionType)
+				Expect(existingCondition).NotTo(BeNil())
```
qinqon: ditto

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+				addedCondition := newNodeInfoList.FindCondition(addedNodeName, addedConditionType)
+				Expect(addedCondition).NotTo(BeNil())
```
qinqon: ditto

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
```diff
@@ -0,0 +1,135 @@
...
+			It("should not add a new node entry to the list", func() {
+				Expect(newNodeInfoList).To(HaveLen(len(originalNodeInfoList)))
+			})
+
+			It("should contain both the old and new conditions", func() {
+				existingCondition := newNodeInfoList.FindCondition(existingNodeName, existingConditionType)
+				Expect(existingCondition).NotTo(BeNil())
+
+				addedCondition := newNodeInfoList.FindCondition(existingNodeName, addedConditionType)
+				Expect(addedCondition).NotTo(BeNil())
```
qinqon: Compare with expected condition

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
phoracek: Done

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
phoracek: Done

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
phoracek: Done

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
phoracek: Read about this on some BDD rant/tutorial. We should not assume that the original length was 0, that is not what we are testing anyway. We are testing that the list will be extended, no matter the original length.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
phoracek: Done

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
phoracek: I checked it on the previous case, I don't think we need to do it everywhere though. We use the condition module for that and that is already extensively tested.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types_test.go
phoracek: Already tested by tests of the underlying module.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -30,36 +48,65 @@ type NodeNetworkConfigurationPolicySpec struct {
...
+type EnactmentList []Enactment
 
-// NodeNetworkConfigurationPolicy is the Schema for the nodenetworkconfigurationpolicies API
-// +k8s:openapi-gen=true
-// +kubebuilder:subresource:status
-// +kubebuilder:resource:path=nodenetworkconfigurationpolicies,shortName=nncp
-type NodeNetworkConfigurationPolicy struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+type Enactment struct {
```
qinqon: NodeNetworkConfigurationEnactment

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -30,36 +48,65 @@ type NodeNetworkConfigurationPolicySpec struct {
...
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+func (enactments *EnactmentList) SetCondition(nodeName string, conditionType ConditionType, status corev1.ConditionStatus, reason ConditionReason, message string) {
+	nodeInfo := enactments.find(nodeName)
 
-// NodeNetworkConfigurationPolicyList contains a list of NodeNetworkConfigurationPolicy
-type NodeNetworkConfigurationPolicyList struct {
-	metav1.TypeMeta `json:",inline"`
-	metav1.ListMeta `json:"metadata,omitempty"`
-	Items           []NodeNetworkConfigurationPolicy `json:"items"`
+	if nodeInfo == nil {
```
qinqon: nodeNetworkConfigurationEnactment

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
phoracek: NodeNetworkConfigurationEnactment will be the name of the new object. It will be a full object with Metadata, State etc... I don't want it to be confused with this list. The NodeNetworkCofigurationPolicy will have a list of all Enactments (having key .nodeNetworkConfigurationEnactment is a little too much). This list will contain link to enactment or simply matched=False and name of respective node.

This list will remain here, just its contents will be moved farther to NodeNetworkConfigurationEnactment obejcts. Do we want to gave `policy.enactments` or `policy.nodes`? I'm kinda in favor of the former.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
phoracek: Fixed

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
qinqon: ok, fair enough, I am for policy.enactments.

PR #249, opened 2019-11-09
pass raw State to helper functions

phoracek:
Make set functions more generic by changing their input type to plain State (thus not coupled with NNS).


phoracek:
@qinqon @SchSeba could you please take a look?

PR #251, opened 2019-11-09
remove Spec from NodeNetworkState

phoracek:
The NNS will become read-only and desired state will be moved to NNCP.


phoracek:
@qinqon @SchSeba would you please take a look?

phoracek:
It sure does :D goodbye

PR #252, opened 2019-11-09
drop NodeNetworkState config controller

phoracek:
With configuration being moved to the NNCP controller, there is no need to keep the NNS config controller anymore. Therefore, this patch is removing it.

This patch also renames NNS report controller to simply NNS controller since there is no need to differentiate between two anymore


phoracek:
@qinqon could you take a look please?

PR #256, opened 2019-11-12
add configuration code to policy controller

phoracek:
Run `nmstatectl set` directly from the NNCP controller. Tests will be adjusted
in follow-up PR since the scale is just too big to easily review in a single PR.

### DIFF
--- a/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -20,6 +21,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	nmstate "github.com/nmstate/kubernetes-nmstate/pkg/helper"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -73,13 +75,19 @@ func forThisNodePredicate(cl client.Client) predicate.Funcs {
...
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
-			return nodeSelectorMatchesThisNode(cl, deleteEvent.Object)
+			return false
 		},
 		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
-			return nodeSelectorMatchesThisNode(cl, updateEvent.ObjectNew)
-		},
-		GenericFunc: func(genericEvent event.GenericEvent) bool {
-			return nodeSelectorMatchesThisNode(cl, genericEvent.Object)
+			if !nodeSelectorMatchesThisNode(cl, updateEvent.ObjectNew) {
```
qinqon: We are missing a lot of stuff from https://github.com/nmstate/kubernetes-nmstate/blob/master/pkg/controller/nodenetworkstateconfiguration/nodenetworkstateconfiguration_controller.go#L70-L94

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -135,27 +143,91 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
...
-	nodeNetworkStateKey := types.NamespacedName{Name: nodeName}
-	err = r.client.Get(context.TODO(), nodeNetworkStateKey, nodeNetworkState)
+	nmstateOutput, err := nmstate.ApplyDesiredState(instance.Spec.DesiredState)
 	if err != nil {
-		if errors.IsNotFound(err) {
-			reqLogger.Info(fmt.Sprintf("the NodeNetworkState for %s is not there yet, let's requeue", nodeName))
-			// If there is no NodeNetworkState let's requeue could be that
-			// we are in the middle of the creation
-			return reconcile.Result{Requeue: true}, nil
+		errmsg := fmt.Errorf("error reconciling nodenetworkstate at desired state apply: %s, %v", nmstateOutput, err)
```
qinqon: s/nodenetworkstate/nodenetworkconfigurationpolicy/

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: The reason is that with the new design we reapply re-reconcile the configuration periodically, the check whether we need to Update or not should haven via the dry-run we discussed. We need to compare the desired configuration with the state of host, not with the previous desired configuration.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: Fixed

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Ahh that's right, ok.

PR #261, opened 2019-11-13
Use NNCP to set desiredState at e2e tests

qinqon:
Now that we have move desiredState to NNCP we have to change
e2e to use it instead of NNS to fix them.

### DIFF
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -31,8 +31,9 @@ import (
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-const ReadTimeout = 120 * time.Second
+const ReadTimeout = 180 * time.Second
 const ReadInterval = 1 * time.Second
+const TestPolicy = "test-policy"
 
 var (
 	bridgeCounter = 0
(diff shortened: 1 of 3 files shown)


test/e2e/utils.go
```diff
@@ -170,31 +171,19 @@ func setDesiredStateWithPolicy(name string, desiredState nmstatev1alpha1.State)
...
-		updateDesiredStateAtNode(node, desiredState)
-	}
+func updateDesiredStateAtNode(node string, desiredState nmstatev1alpha1.State) {
+	nodeSelector := map[string]string{"kubernetes.io/hostname": node}
+	setDesiredStateWithPolicyAndNodeSelector(DefaultPolicy, desiredState, nodeSelector)
 }
 
 // TODO: After we implement policy delete (it will cleanUp desiredState) we have
 //       to remove this
 func resetDesiredStateForNodes() {
```
HlinaCZ: Is this comment still valid?

test/e2e/utils.go
```diff
@@ -31,8 +31,9 @@ import (
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-const ReadTimeout = 120 * time.Second
+const ReadTimeout = 180 * time.Second
 const ReadInterval = 1 * time.Second
+const DefaultPolicy = "default-policy"
```
phoracek: It is not really a default of anything. Maybe TestPolicy would be better?

test/e2e/utils.go
```diff
@@ -31,8 +31,9 @@ import (
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-const ReadTimeout = 120 * time.Second
+const ReadTimeout = 180 * time.Second
```
phoracek: Why is this change needed? There should be no such difference compared to the original code with configuration in NNS. (I was doing shotgun debugging in my PRs, if you saw it there)

qinqon: I have reuse the condition check and it was 180 there so I have put it for all 

https://github.com/nmstate/kubernetes-nmstate/pull/261/files#diff-27d7aee9b3a12c999081eedcb5a054aeL124

test/e2e/utils.go
```diff
@@ -315,24 +312,28 @@ func currentState(namespace string, node string, currentStateYaml *nmstatev1alph
...
 	return Eventually(func() corev1.ConditionStatus {
-		state := nodeNetworkState(key)
-		condition := state.Status.Conditions.Find(conditionType)
+		policy := nodeNetworkConfigurationPolicy(key)
+		condition := policy.Status.Enactments.FindCondition(node, conditionType)
 		if condition == nil {
 			return corev1.ConditionUnknown
 		}
 		return condition.Status
 	}, ReadTimeout, ReadInterval)
```
HlinaCZ: In function `checkEnacmentConditionConsistently()` has time outs explicitly set. In this function it is set by variables. I would like a consistent style, please.

qinqon: Thanks, let's define per nature the timeout and internval.

test/e2e/utils.go
qinqon: Nop, what we do is to ensure desiredState is cleanup but setting primary nic up

phoracek: AFAIK the comment was made when we though we would de-configure desiredState when it is removed. So the bridge would be rolled back. We decided that that won't be the case, so we will use resetDesiredStateForNodes forever. Thus we could drop it.

PR #263, opened 2019-11-14
Remove logs from metrics after operator-sdk upgrade

qinqon:
After the upgrade we get a massive number of [1], there is a workaround for it
[2] it's suppose to not ve needed at k8s >= 1.15

[1] reflector.go:95: Failed to list *unstructured.Unstructured: the server could not find the requested resource
[2] https://github.com/operator-framework/operator-sdk/issues/1858#issuecomment-548323725


cmd/manager/main.go
```diff
@@ -165,6 +165,10 @@ func serveCRMetrics(cfg *rest.Config) error {
 	if err != nil {
 		return err
 	}
+	// To bypass ugly metrics logs after operator-sdk 0.12.0 upgrade
+	// https://github.com/operator-framework/operator-sdk/issues/1858#issuecomment-548323725
+	operatorNs = ""
```
phoracek: I don't get why we get the namespace few lines above.

qinqon: You right, let's remove it all together.

qinqon:
@phoracek haven't check about metrics, let's do that at a follow up.

PR #271, opened 2019-11-19
Add Progressing state to enactment

qinqon:
As first step to get policy conditions that will represent the
overall cluster state we need to add the missing state of Progressing
to the enactment so policy condition can be calculated

### DIFF
--- a/pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
+++ b/pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
@@ -49,18 +49,6 @@ type NodeNetworkConfigurationPolicyStatus struct {
 	Enactments EnactmentList `json:"enactments,omitempty" optional:"true"`
 }
 
-const (
-	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
-	NodeNetworkConfigurationPolicyConditionFailing     ConditionType = "Failing"
-	NodeNetworkConfigurationPolicyConditionProgressing ConditionType = "Progressing"
-)
-
-const (
-	NodeNetworkConfigurationPolicyConditionFailedToConfigure        ConditionReason = "FailedToConfigure"
-	NodeNetworkConfigurationPolicyConditionSuccessfullyConfigured   ConditionReason = "SuccessfullyConfigured"
-	NodeNetworkConfigurationPolicyConditionConfigurationProgressing ConditionReason = "ConfigurationProgressing"
-)
-
 func init() {
 	SchemeBuilder.Register(&NodeNetworkConfigurationPolicy{}, &NodeNetworkConfigurationPolicyList{})
 }
--- a/pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
+++ b/pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
@@ -50,13 +50,15 @@ type NodeNetworkConfigurationPolicyStatus struct {
 }
 
 const (
-	NodeNetworkConfigurationPolicyConditionAvailable ConditionType = "Available"
-	NodeNetworkConfigurationPolicyConditionFailing   ConditionType = "Failing"
+	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
+	NodeNetworkConfigurationPolicyConditionFailing     ConditionType = "Failing"
+	NodeNetworkConfigurationPolicyConditionProgressing ConditionType = "Progressing"
 )
 
 const (
-	NodeNetworkConfigurationPolicyConditionFailedToConfigure      ConditionReason = "FailedToConfigure"
-	NodeNetworkConfigurationPolicyConditionSuccessfullyConfigured ConditionReason = "SuccessfullyConfigured"
+	NodeNetworkConfigurationPolicyConditionFailedToConfigure        ConditionReason = "FailedToConfigure"
+	NodeNetworkConfigurationPolicyConditionSuccessfullyConfigured   ConditionReason = "SuccessfullyConfigured"
+	NodeNetworkConfigurationPolicyConditionConfigurationProgressing ConditionReason = "ConfigurationProgressing"
 )
 
 func init() {
--- a/pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
+++ b/pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
@@ -1,7 +1,6 @@
 package v1alpha1
 
 import (
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -48,52 +47,6 @@ type NodeNetworkConfigurationPolicySpec struct {
 // +k8s:openapi-gen=true
 type NodeNetworkConfigurationPolicyStatus struct {
 	Enactments EnactmentList `json:"enactments,omitempty" optional:"true"`
-}
-
-// TODO: This is a temporary solution. This list will be replaced by a dedicated
-// NodeNetworkConfigurationEnactment object.
-type EnactmentList []Enactment
-
-type Enactment struct {
-	NodeName   string        `json:"nodeName"`
-	Conditions ConditionList `json:"conditions,omitempty"`
-}
-
-func NewEnactment(nodeName string) Enactment {
-	return Enactment{
-		NodeName:   nodeName,
-		Conditions: ConditionList{},
-	}
-}
-
-func (enactments *EnactmentList) SetCondition(nodeName string, conditionType ConditionType, status corev1.ConditionStatus, reason ConditionReason, message string) {
-	enactment := enactments.find(nodeName)
-
-	if enactment == nil {
-		enactment := NewEnactment(nodeName)
-		enactment.Conditions.Set(conditionType, status, reason, message)
-		*enactments = append(*enactments, enactment)
-		return
-	}
-
-	enactment.Conditions.Set(conditionType, status, reason, message)
-}
-
-func (enactments EnactmentList) FindCondition(nodeName string, conditionType ConditionType) *Condition {
-	enactment := enactments.find(nodeName)
-	if enactment == nil {
-		return nil
-	}
-	return enactment.Conditions.Find(conditionType)
-}
-
-func (enactments EnactmentList) find(nodeName string) *Enactment {
-	for i, enactment := range enactments {
-		if enactment.NodeName == nodeName {
-			return &enactments[i]
-		}
-	}
-	return nil
 }
 
 const (
--- a/pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
@@ -10,10 +10,6 @@ func setEnactmentFailedToConfigure(enactments *nmstatev1alpha1.EnactmentList, no
 	setEnactmentFailed(enactments, nodeName, nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionFailedToConfigure, message)
 }
 
-func setEnactmentFailedToFindPolicy(enactments *nmstatev1alpha1.EnactmentList, nodeName string, message string) {
-	setEnactmentFailed(enactments, nodeName, nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionFailedToFindPolicy, message)
-}
-
 func setEnactmentFailed(enactments *nmstatev1alpha1.EnactmentList, nodeName string, reason nmstatev1alpha1.ConditionReason, message string) {
 	enactments.SetCondition(
 		nodeName,
--- a/pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
@@ -1,0 +1,87 @@
+package conditions
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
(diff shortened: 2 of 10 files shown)


pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -202,6 +203,13 @@ func setConditionFailed(enactments *nmstatev1alpha1.EnactmentList, message strin
 		nmstatev1alpha1.NodeNetworkConfigurationPolicyConditionFailedToConfigure,
 		"",
 	)
+	enactments.SetCondition(
```
phoracek: shouldn't we do this in defer?

qinqon:
@phoracek this is a draft PR is not ready for review

qinqon:
@oshoval @phoracek this is ready now.

qinqon:
ci test please

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
```diff
@@ -0,0 +1,87 @@
+package conditions
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+func setEnactmentFailedToConfigure(enactments *nmstatev1alpha1.EnactmentList, nodeName string, message string) {
+	setEnactmentFailed(enactments, nodeName, nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionFailedToConfigure, message)
```
oshoval: maybe names shouldn't have version as part of them ?

unless version isnt going to change

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func setEnactmentFailedToFindPolicy(enactments *nmstatev1alpha1.EnactmentList, nodeName string, message string) {
+	setEnactmentFailed(enactments, nodeName, nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionFailedToFindPolicy, message)
+}
+
+func setEnactmentFailed(enactments *nmstatev1alpha1.EnactmentList, nodeName string, reason nmstatev1alpha1.ConditionReason, message string) {
+	enactments.SetCondition(
```
oshoval: is it ok that this 3 commands aren't part of one critical section? 
or there isnt parallel execution that might be corrupted due to this ?

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
qinqon: This is a must in golang, so you are able to have part of the code using different api versions.

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
qinqon: we just do a retryOnConflict with this, there is no parallel code running.

oshoval:
lgtm

qinqon:
rollback tests are failing, at pre-flight checks for failed conditions, have to refactor this enactment conditions check so we can check multiple condition statuses for same type (failed can be False or Unknown depending on when you run the test and they both would be correct).

qinqon:
ci test please

qinqon:
ci test please

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		m.logger.Error(err, "Success condition update failed while reporting success: %v", err)
+	}
+}
+
+func (m *Manager) updateEnactmentCondition(
+	condition func(*nmstatev1alpha1.EnactmentList, string, string),
+	message string,
+) error {
+	// Set enactment condition
```
phoracek: Redundant, the function name is good enough.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
+package conditions
```
phoracek: I like the manager a lot.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -139,88 +141,28 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 			// Return and don't requeue
 			return reconcile.Result{}, nil
 		}
+		conditionsManager.FailedToFindPolicy(err)
```
phoracek: This is wrong. The previous if is ran if it is not found (and it not being found is totally valid, someone was just faster in removing it). If we reach to this point, it is not that the policy does not exist (and cannot be found), it means we have some unexpected bad error. We should make it obvious. The log is good, but for the condition, we should be explicit that it is an unexpected error. (Also, how do we set the condition if there is no policy)

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+func NewManager(client client.Client, nodeName string, policyName types.NamespacedName) Manager {
+	return Manager{
+		client:     client,
+		nodeName:   nodeName,
+		policyName: policyName,
+		logger:     log.WithValues("node", nodeName, "policy", policyName),
+	}
+}
+
+func (m *Manager) Progressing() {
```
phoracek: SetProgressing. As you said, Progressing would be a getter for the property. I think using Set in these methods would make the code little more readable. No questions to be asked.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (m *Manager) Success() {
+	err := m.updateEnactmentCondition(setEnactmentSuccess, "successfully reconciled")
+	if err != nil {
+		m.logger.Error(err, "Success condition update failed while reporting success: %v", err)
+	}
+}
+
+func (m *Manager) updateEnactmentCondition(
+	condition func(*nmstatev1alpha1.EnactmentList, string, string),
```
phoracek: conditionSetter?

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	message string,
+) error {
+	// Set enactment condition
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
phoracek: I think we can get rid of this helper. Due to that we do 3 times as many updates of conditions. We should update all 3 conditions at once and be atomic.

test/e2e/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func conditionsToYaml(conditions nmstatev1alpha1.ConditionList) string {
+	manifest, err := yaml.Marshal(conditions)
+	if err != nil {
+		panic(err)
+	}
+	return string(manifest)
+}
+func enactmentConditionsStatus(node string) nmstatev1alpha1.ConditionList {
```
phoracek: Please insert an empty line between functions. Is not that handled by gofmt?

test/e2e/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"k8s.io/apimachinery/pkg/types"
+	yaml "sigs.k8s.io/yaml"
+
+	corev1 "k8s.io/api/core/v1"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+type ExpectedConditionsStatus struct {
```
phoracek: Do we need this to be public?

BTW, consider creating a separate package for condition helpers, we should start making test utils more structured.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -50,60 +49,16 @@ type NodeNetworkConfigurationPolicyStatus struct {
...
-	}
-	return nil
-}
-
 const (
-	NodeNetworkConfigurationPolicyConditionAvailable ConditionType = "Available"
-	NodeNetworkConfigurationPolicyConditionFailing   ConditionType = "Failing"
+	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
+	NodeNetworkConfigurationPolicyConditionFailing     ConditionType = "Failing"
+	NodeNetworkConfigurationPolicyConditionProgressing ConditionType = "Progressing"
```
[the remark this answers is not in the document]
qinqon: No, we don't, let's remove them.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: You are right, conditions are at policy we cannot state at policy that it cannot be found at conditions :-)

Let's just remove this since there is nothing except for the log that we can do.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: conditionsManager.EnactProgressing() ?

qinqon: I would prefer use "Notify" or "Enact" to notify events to the manager (is like a little fsm)

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: the conditionSetter changes all three at the same time so we are already atomic.

test/e2e/conditions.go
qinqon: Is not handle by it.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
phoracek: ah, gotcha

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,67 @@
...
+func (m *Manager) NotifyProgressing() {
+	err := m.updateEnactmentCondition(setEnactmentProgressing, "Applying desired state")
+	if err != nil {
+		m.logger.Error(err, "Error change state to progressing")
+	}
+}
+func (m *Manager) NotifyFailedToConfigure(failedErr error) {
+	err := m.updateEnactmentCondition(setEnactmentFailedToConfigure, failedErr.Error())
+	if err != nil {
+		m.logger.Error(err, "Error chaing state to failing at configure with error: %v", failedErr)
```
phoracek: Is not the failedErr already logged somewhere else? The output will be large, we don't wan't to see it twice.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func (m *Manager) NotifySuccess() {
+	err := m.updateEnactmentCondition(setEnactmentSuccess, "successfully reconciled")
+	if err != nil {
+		m.logger.Error(err, "Success condition update failed while reporting success: %v", err)
```
phoracek: No need to have the err twice in the log. Error already takes it as the first arg

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
phoracek: Then please make it conditionsSetter so I don't get confused twice

test/e2e/nncp_conditions_test.go
```diff
@@ -55,14 +90,22 @@ var _ = Describe("NodeNetworkStateCondition", func() {
 			resetDesiredStateForNodes()
 		})
 
-		It("should have Failing ConditionType set to true", func() {
+		FIt("should have Failing ConditionType set to true", func() {
```
phoracek: Focus?!

test/e2e/nncp_conditions_test.go
```diff
@@ -19,27 +19,62 @@ func invalidConfig(bridgeName string) nmstatev1alpha1.State {
 `, bridgeName))
 }
 
-var _ = Describe("NodeNetworkStateCondition", func() {
+var _ = Describe("EnactmentCondition", func() {
```
phoracek: Could we chnage the module name?

test/e2e/nncp_conditions_test.go
```diff
@@ -19,27 +19,62 @@ func invalidConfig(bridgeName string) nmstatev1alpha1.State {
...
+			runAtPods("mv", "/usr/local/bin/vlan-filtering.bak", "/usr/local/bin/vlan-filtering")
 			updateDesiredState(linuxBrAbsent(bridge1))
 			for _, node := range nodes {
 				interfacesNameForNodeEventually(node).ShouldNot(ContainElement(bridge1))
 			}
 			By("Reset desired state at all nodes")
 			resetDesiredStateForNodes()
 		})
-		It("should have Available ConditionType set to true", func() {
+		FIt("should go from Progressing to Available", func() {
```
phoracek: F!

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: failedErr is at nmstate log and condition.message, they are quite different places is good to print it at both I think

qinqon: I get you know, you right, is confusing to see two errors.

phoracek:
Great work, thanks!

PR #273, opened 2019-11-25
Move enactments nns

qinqon:


qinqon:
@phoracek I am enternaining the idea of moving enactments to NodeNetworkState, It feels right, we can talk about it with this PR

PR #279, opened 2019-11-28
Enactment crd

qinqon:


deploy/crds/nmstate.io_v1alpha1_nodenetworkconfigurationenactment_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: nmstate.io/v1alpha1
```
phoracek: Please make this into a proper example

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
```diff
@@ -51,38 +51,106 @@ spec:
...
-                      - type
-                      type: object
-                    type: array
-                  nodeName:
-                    type: string
-                required:
-                - nodeName
-                type: object
-              type: array
+              description: NodeNetworkConfigurationPolicyList contains a list of NodeNetworkConfigurationPolicy
```
phoracek: Are you nesting NNCE into NNCP? We only want to reference the NNCE from here.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -0,0 +1,61 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status NodeNetworkConfigurationEnactmentStatus `json:"status,omitempty"`
+}
+
+// NodeNetworkConfigurationEnactmentStatus defines the observed state of NodeNetworkConfigurationEnactment
+// +k8s:openapi-gen=true
+type NodeNetworkConfigurationEnactmentStatus struct {
+	Conditions ConditionList  `json:"conditions,omitempty" optional:"true"`
```
phoracek: Add a TODO to include rendered spec here.

deploy/crds/nmstate.io_v1alpha1_nodenetworkconfigurationenactment_cr.yaml
qinqon: Since it's read only does it make sense to have examples ?

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
qinqon: What do you mean ?

cmd/manager/main.go
```diff
@@ -161,10 +161,11 @@ func serveCRMetrics(cfg *rest.Config) error {
 		return err
 	}
 	// Get the namespace the operator is currently deployed in.
-	operatorNs, err := k8sutil.GetOperatorNamespace()
-	if err != nil {
-		return err
-	}
+	// operatorNs, err := k8sutil.GetOperatorNamespace()
```
phoracek: No commented out code please

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
phoracek: But neither to include the whole object in it, right? This is a CRD for NNCP, with your change it contains list of enactments in their full format, why? We already have standalone NNCE in Kubernetes API, we don't want to see all NNCE when getting a single policy.

deploy/crds/nmstate.io_v1alpha1_nodenetworkconfigurationenactment_cr.yaml
phoracek: Examples are interesting for newcomers, but hard to maintain and keep up-to-date. Up to you, either make it a proper example or drop it.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -0,0 +1,50 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status NodeNetworkConfigurationEnactmentStatus `json:"status,omitempty"`
+}
+
+// NodeNetworkConfigurationEnactmentStatus defines the observed state of NodeNetworkConfigurationEnactment
+// +k8s:openapi-gen=true
+type NodeNetworkConfigurationEnactmentStatus struct {
+	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
```
phoracek: Please don't make it optional or omit empty. Make it clear that it will be always there. After all, that's why we have this object in the first place.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -0,0 +1,50 @@
...
+type NodeNetworkConfigurationEnactment struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status NodeNetworkConfigurationEnactmentStatus `json:"status,omitempty"`
+}
+
+// NodeNetworkConfigurationEnactmentStatus defines the observed state of NodeNetworkConfigurationEnactment
+// +k8s:openapi-gen=true
+type NodeNetworkConfigurationEnactmentStatus struct {
```
phoracek: NNCE is not completed until we have rendered state as part of the status. It can be done in a follow-up.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -46,7 +46,7 @@ type NodeNetworkConfigurationPolicySpec struct {
 // NodeNetworkConfigurationPolicyStatus defines the observed state of NodeNetworkConfigurationPolicy
 // +k8s:openapi-gen=true
 type NodeNetworkConfigurationPolicyStatus struct {
-	Enactments EnactmentList `json:"enactments,omitempty" optional:"true"`
+	Enactments NodeNetworkConfigurationPolicyList `json:"enactments,omitempty" optional:"true"`
```
phoracek: This is not the final state, is it?

cmd/manager/main.go
qinqon: This is not from this PR this is from the k8s-1.16 one, let me remove this change.

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
qinqon: Ahh wait didn't show it, let me remove it.

deploy/crds/nmstate.io_v1alpha1_nodenetworkconfigurationenactment_cr.yaml
qinqon: Let's play nice with operator-sdk and keep them.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
qinqon: This is wrong that's why we have enactments in the CRD, removed it already.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
qinqon: Let's do it here, so we have a complete CRD, the content is just about copy desiredState right now.

test/e2e/conditions.go
```diff
@@ -49,14 +63,21 @@ func enactmentConditionsStatus(node string) nmstatev1alpha1.ConditionList {
...
 }
 
-func enactmentConditionsStatusConsistently(node string) AsyncAssertion {
+func enactmentConditionsStatusForPolicyConsistently(node string, policy string) AsyncAssertion {
 	return Consistently(func() nmstatev1alpha1.ConditionList {
-		return enactmentConditionsStatus(node)
+		return enactmentConditionsStatus(node, policy)
 	}, 5*time.Second, 1*time.Second)
 }
+func enactmentConditionsStatusEventually(node string) AsyncAssertion {
```
phoracek: Please insert an empty line

test/e2e/conditions.go
```diff
@@ -26,17 +29,28 @@ func conditionsToYaml(conditions nmstatev1alpha1.ConditionList) string {
...
+func nodeNetworkConfigurationEnactment(key types.NamespacedName) nmstatev1alpha1.NodeNetworkConfigurationEnactment {
+	state := nmstatev1alpha1.NodeNetworkConfigurationEnactment{}
+	Eventually(func() error {
+		return framework.Global.Client.Get(context.TODO(), key, &state)
+	}, ReadTimeout, ReadInterval).ShouldNot(HaveOccurred())
+	return state
+}
+
+func enactmentKey(node string, policy string) types.NamespacedName {
+	return types.NamespacedName{Name: node + "-" + policy}
```
phoracek: please try to use a different separator as we talked about

test/e2e/nncp_cleanup_test.go
```diff
@@ -0,0 +1,53 @@
...
+			deletePolicy(bridge1)
+		})
+		It("should also delete nodes enactments", func() {
+			for _, node := range nodes {
+				Eventually(func() bool {
+					key := enactmentKey(node, bridge1)
+					state := nmstatev1alpha1.NodeNetworkState{}
+					err := framework.Global.Client.Get(context.TODO(), key, &state)
+					return errors.IsNotFound(err)
+				}, 5*time.Second, 1*time.Second).Should(BeTrue())
```
phoracek: Please add a message describing what are we checking in this Should. BeTrue() is terribly vague.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -48,20 +54,48 @@ func (m *Manager) NotifySuccess() {
 	}
 }
 
+func (m *Manager) initializeEnactment() (*nmstatev1alpha1.NodeNetworkConfigurationEnactment, error) {
+	//TODO: Don't harcode this take it from m.policy meta
```
phoracek: Will you resolve this as part of the PR?

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -48,20 +54,48 @@ func (m *Manager) NotifySuccess() {
...
+	ownerRefList := []metav1.OwnerReference{{Name: m.policy.Name, Kind: "NodeNetworkConfigurationPolicy", APIVersion: "v1alpha1", UID: m.policy.UID}}
+
+	enactment := nmstatev1alpha1.NodeNetworkConfigurationEnactment{
+		// Create NodeNetworkState for this node
```
phoracek: I don't understand this comment and the place where it is

test/e2e/conditions.go
qinqon: It's done

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -48,20 +57,56 @@ func (m *Manager) NotifySuccess() {
 	}
 }
 
+func (m *Manager) initializeEnactment() (*nmstatev1alpha1.NodeNetworkConfigurationEnactment, error) {
+
+	node := corev1.Node{}
+	err := m.client.Get(context.TODO(), types.NamespacedName{Name: m.nodeName}, &node)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("cannot find node %s", m.nodeName))
```
phoracek: It is not that we cannot find it, but we cannot get it. "Find" suggest that we looked for it but it does not exist, which should not be the case (we run on the node, so it probably exists). Could you change the message?

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -48,20 +57,56 @@ func (m *Manager) NotifySuccess() {
...
+			}
 		}
-
-		conditionsSetter(&instance.Status.Enactments, m.nodeName, message)
+		conditionsSetter(instance, message)
 
 		err = m.client.Status().Update(context.TODO(), instance)
-		return err
+		if err != nil {
+			return errors.Wrap(err, "updaging enactment state failed")
```
phoracek: updating

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -48,20 +57,56 @@ func (m *Manager) NotifySuccess() {
...
-		instance := &nmstatev1alpha1.NodeNetworkConfigurationPolicy{}
-		err := m.client.Get(context.TODO(), m.policyName, instance)
+		instance := &nmstatev1alpha1.NodeNetworkConfigurationEnactment{}
+		err := m.client.Get(context.TODO(), types.NamespacedName{Name: m.enactmentName}, instance)
 		if err != nil {
-			return err
+			if !k8s_errors.IsNotFound(err) {
+				return errors.Wrap(err, "enactment read failed")
+			}
+			instance, err = m.initializeEnactment()
```
phoracek: I think we should not initialize a new Enactment from here, that should happen on a higher level.

In this state, the condition manager is responsible of creating the enactment. That would confict with the upper layer where we want to really initialize the enactment with Initial conditions and most important - rendered state. Having initialization on two places makes this confusing.

I suggest we initialize it only on a single place, making sure it exists once at the beginning of the reconciliation (note that it should have all conditions already set to something there). The manager would only update (as the name updateEnactmentCondition suggest), if the enactment it manages does not exist, it would fail.

qinqon: Ack will move enactment creating outside of manager, but at the next PR we will have to do it pre-Reconcile to set the Matching condition.

phoracek: Or just in general when the enactment does not exist (it should when we reconcile a policy that is not being deleted). It is like "Now we initialize our metadata field, this is the initial state".

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -46,7 +49,12 @@ type NodeNetworkConfigurationPolicySpec struct {
 // NodeNetworkConfigurationPolicyStatus defines the observed state of NodeNetworkConfigurationPolicy
 // +k8s:openapi-gen=true
 type NodeNetworkConfigurationPolicyStatus struct {
-	Enactments EnactmentList `json:"enactments,omitempty" optional:"true"`
+	// TODO: Add overall condition
+	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
+}
+
+func enactmentKey(node string, policy string) types.NamespacedName {
```
phoracek: Why is this duplicated? It is already present in the ...enactment_types.go

deploy/crds/nmstate.io_v1alpha1_nodenetworkconfigurationenactment_cr.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: NodeNetworkConfigurationEnactment
+metadata:
+  creationTimestamp: "2019-12-03T08:32:47Z"
+  generation: 1
+  name: node01-test-policy
+  ownerReferences:
```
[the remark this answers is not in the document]
phoracek: It is still missing in the example.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -49,19 +51,22 @@ func (m *Manager) NotifySuccess() {
 }
 
 func (m *Manager) updateEnactmentCondition(
```
phoracek: updateEnactmentCondition**s**

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -49,19 +51,22 @@ func (m *Manager) NotifySuccess() {
...
-	conditionsSetter func(*nmstatev1alpha1.EnactmentList, string, string),
+	conditionsSetter func(*nmstatev1alpha1.NodeNetworkConfigurationEnactment, string),
```
phoracek: Could we give the callback only list of Conditions as a parameter? Make sure they don't touch what they should not and the interface is slimmer?

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -49,19 +51,22 @@ func (m *Manager) NotifySuccess() {
...
+			return errors.Wrap(err, "getting enactment failed")
 		}
 
-		conditionsSetter(&instance.Status.Enactments, m.nodeName, message)
+		conditionsSetter(instance, message)
 
 		err = m.client.Status().Update(context.TODO(), instance)
-		return err
+		if err != nil {
+			return errors.Wrap(err, "updating enactment state failed")
```
phoracek: If you wrap the error, how does the `RetryOnConfict` knows that the callback failed on conflict?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -69,10 +72,46 @@ func nodeSelectorMatchesThisNode(cl client.Client, eventObject runtime.Object) b
...
 
+func initializeEnactment(client client.Client, policy *nmstatev1alpha1.NodeNetworkConfigurationPolicy) (*nmstatev1alpha1.NodeNetworkConfigurationEnactment, error) {
+
+	node := corev1.Node{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, &node)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("cannot get node %s", nodeName))
+	}
+
+	ownerRefList := []metav1.OwnerReference{
```
phoracek: Can we initialize the object in the Enactment constructor (under types.go) and handle the Get and Create here?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -69,10 +72,46 @@ func nodeSelectorMatchesThisNode(cl client.Client, eventObject runtime.Object) b
...
+	return &enactment, nil
+}
+
 func forThisNodePredicate(cl client.Client) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(createEvent event.CreateEvent) bool {
-			return nodeSelectorMatchesThisNode(cl, createEvent.Object)
+			if nodeSelectorMatchesThisNode(cl, createEvent.Object) {
+				policy := createEvent.Object.(*nmstatev1alpha1.NodeNetworkConfigurationPolicy)
+				_, err := initializeEnactment(cl, policy)
```
phoracek: We should not do any side effects in predicate. The initialization have to happen inside the reconcile function. If the NNCE does not exist yet, create it.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: Depends on this:
```
// ReasonForError returns the HTTP status for a particular error.               
func ReasonForError(err error) metav1.StatusReason {                            
    switch t := err.(type) {                                                    
    case APIStatus:                                                             
        return t.Status().Reason                                                
    }                                                                           
    return metav1.StatusReasonUnknown                                           
}   
```

phoracek: So by wrapping you lose the granularity, don't you?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Then how we will state Matching state to False ?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: no one else is going to create and enactment with owner refs here is good enough let's not add to much stuff at _types.go

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: We need to move the Matching check to Reconcile function if we want to act (read "create objects on API") based on it.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: Fair enought. Would you mind splitting this function in here? So it is easier to test, one function creates an object with no side effects, the other calls API.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: We may want to implement some counter or the like or even if we go for phase passing the whole enactment is a good thing also, we are not a library the "they" is us.

phoracek: And we want to protect ourselves.

When reading a code, having smaller arguments is better since you know what can be changed or consider inside the function without looking into it and reading it.

If we will be refactoring the code, it is easier to move such a function around and make it generic if needed.

Smaller signature makes it easier to provide data when testing it. When we change the big structure (Enactment) we may need to change tests of this small function even though it would no be directly affected.

The function is mean to update a condition. Would you please give an example of when we may use the whole function?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: At the end I have create a NewEnactment at _types.go it feels ok.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Moved the initialize to Reconcile, well move the node check at Reconcile at Matching state PR.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: I was thinking about the missing enactment state "Counter of reconfigurations" but since this is different from conditions, we are ok I am going to pass a ConditionList

deploy/crds/nmstate.io_v1alpha1_nodenetworkconfigurationenactment_cr.yaml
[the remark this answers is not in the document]
phoracek: @qinqon please update the example.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
phoracek: @qinqon Please drop this function from this module.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -72,7 +74,10 @@ func nodeSelectorMatchesThisNode(cl client.Client, eventObject runtime.Object) b
 func forThisNodePredicate(cl client.Client) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(createEvent event.CreateEvent) bool {
-			return nodeSelectorMatchesThisNode(cl, createEvent.Object)
+			if nodeSelectorMatchesThisNode(cl, createEvent.Object) {
+				return true
+			}
+			return false
```
phoracek: Why did you change this? Isn't it exactly what the original did?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -120,6 +125,30 @@ type ReconcileNodeNetworkConfigurationPolicy struct {
...
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, &node)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("cannot get node %s", nodeName))
+	}
+
+	enactment := nmstatev1alpha1.NewEnactment(node, policy)
+
+	err = r.client.Create(context.TODO(), &enactment)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("error creating NodeNetworkConfigurationEnactment: %+v", enactment))
```
phoracek: What does +v do?

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -0,0 +1,69 @@
...
+	NodeNetworkConfigurationEnactmentConditionSuccessfullyConfigured   ConditionReason = "SuccessfullyConfigured"
+	NodeNetworkConfigurationEnactmentConditionConfigurationProgressing ConditionReason = "ConfigurationProgressing"
+)
+
+func EnactmentKey(node string, policy string) types.NamespacedName {
+	return types.NamespacedName{Name: fmt.Sprintf("%s.%s", node, policy)}
+}
+
+func NewEnactment(node corev1.Node, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
+	return NodeNetworkConfigurationEnactment{
```
phoracek: Would it make sense to initialize with all our conditions set to Unknown? In the API, a missing condition is considered to be Unknown implicitly, but I feel it may be better for the user to always give them all conditions.

test/e2e/conditions.go
```diff
@@ -26,17 +29,24 @@ func conditionsToYaml(conditions nmstatev1alpha1.ConditionList) string {
 	return string(manifest)
 }
 
-func enactmentConditionsStatus(node string) nmstatev1alpha1.ConditionList {
-	key := types.NamespacedName{Name: TestPolicy}
-	policy := nodeNetworkConfigurationPolicy(key)
+func nodeNetworkConfigurationEnactment(key types.NamespacedName) nmstatev1alpha1.NodeNetworkConfigurationEnactment {
+	state := nmstatev1alpha1.NodeNetworkConfigurationEnactment{}
```
phoracek: s/state/enactment/

test/e2e/nncp_cleanup_test.go
```diff
@@ -0,0 +1,54 @@
...
+	AfterEach(func() {
+		updateDesiredState(linuxBrAbsent(bridge1))
+		resetDesiredStateForNodes()
+	})
+
+	Context("when a policy is deleted", func() {
+		BeforeEach(func() {
+			deletePolicy(bridge1)
+		})
+		FIt("should also delete nodes enactments", func() {
```
phoracek: F!

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: Leftover from some changes, will go back to old version.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
qinqon: It shows the name of the struct fields.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
qinqon: Yep, I prefer explicit.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -13,55 +17,77 @@ import (
...
 
-func (m *Manager) updateEnactmentCondition(
-	conditionsSetter func(*nmstatev1alpha1.EnactmentList, string, string),
+func (m *Manager) nodeNetworkConfigurationEnactment() (nmstatev1alpha1.NodeNetworkConfigurationEnactment, error) {
+	var enactment nmstatev1alpha1.NodeNetworkConfigurationEnactment
+	pollErr := wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		err := m.client.Get(context.TODO(), nmstatev1alpha1.EnactmentKey(m.nodeName, m.policy.Name), &enactment)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				// Let's retry after a while, sometimes it takes some time
```
phoracek: Wait for the enactment to get created should not be considered in the manager. If we need to wait for it explicilty (and Create is not enough), it should happen after we call Create. Adding polling to be "safer" around the codebase makes the application less predictable. We should try to make it as sequential as possible.

qinqon: I always forget that knmstate is distributed we will read the anactments at the same node we create, except for overall policy, let me move it.

PR #285, opened 2019-12-02
Matching state

qinqon:


pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -37,12 +37,15 @@ const (
...
 
 const (
-	NodeNetworkConfigurationEnactmentConditionFailedToConfigure        ConditionReason = "FailedToConfigure"
-	NodeNetworkConfigurationEnactmentConditionSuccessfullyConfigured   ConditionReason = "SuccessfullyConfigured"
-	NodeNetworkConfigurationEnactmentConditionConfigurationProgressing ConditionReason = "ConfigurationProgressing"
+	NodeNetworkConfigurationEnactmentConditionFailedToConfigure                 ConditionReason = "FailedToConfigure"
+	NodeNetworkConfigurationEnactmentConditionSuccessfullyConfigured            ConditionReason = "SuccessfullyConfigured"
+	NodeNetworkConfigurationEnactmentConditionConfigurationProgressing          ConditionReason = "ConfigurationProgressing"
+	NodeNetworkConfigurationEnactmentConditionNodeSelectorNotMatching           ConditionReason = "NodeSelectorNotMatching"
+	NodeNetworkConfigurationEnactmentConditionNodeSelectorAllConditionsMatching ConditionReason = "AllConditionsMatching"
```
phoracek: What do you mean by AllConditions? We should make it AllSelectors in case any selectors are used or just Matching. Conditions may be confused with status conditions.

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
```diff
@@ -72,3 +72,60 @@ func setEnactmentProgressing(enactment *nmstatev1alpha1.NodeNetworkConfiguration
 		"",
 	)
 }
+
+func setEnactmentNodeSelectorNotMatching(enactment *nmstatev1alpha1.NodeNetworkConfigurationEnactment, message string) {
+	setEnactmentNotMatching(enactment, nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionNodeSelectorNotMatching, message)
+}
+func setEnactmentNotMatching(enactment *nmstatev1alpha1.NodeNetworkConfigurationEnactment, reason nmstatev1alpha1.ConditionReason, message string) {
```
phoracek: Empty line please

test/e2e/conditions.go
```diff
@@ -29,6 +29,10 @@ func conditionsToYaml(conditions nmstatev1alpha1.ConditionList) string {
 	return string(manifest)
 }
 
+func enactmentKey(node string, policy string) types.NamespacedName {
```
phoracek: This can cause collisions in case someone is trying to hard to break us. Can we get a different separator? Maybe the escaped slash used in resource names?

qinqon: You mean something like this ?

https://github.com/intel/multus-cni/blob/master/examples/sample-pod.yml#L7

phoracek: Yes, but I'm not sure it is allowed in object names. I think slash would be way better for NNCE than dash.

qinqon: humm not working
```
"node01/test-policy\" is invalid: metadata.name: Invalid value: \"node01/test-policy\":
```

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -196,6 +186,12 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 	}
 
 	conditionsManager := conditions.NewManager(r.client, nodeName, *instance)
+	if !r.nodeSelectorMatchesThisNode(*instance) {
+		reqLogger.Info("Policy node selectors does not match node")
+		conditionsManager.NotifyNodeSelectorNotMatching("TODO")
```
phoracek: Do the TODO! :D

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -152,36 +111,14 @@ var _ = Describe("NodeNetworkConfigurationPolicy controller predicates", func()
...
-				ObjNodeName: "node01",
-				NodeLabels: map[string]string{
-					"label1": "foo",
-					"label2": "bar",
-				},
-				NodeSelector: map[string]string{
-					"label1": "foo",
-					"label2": "bar",
-				},
-				GenerationOld:   1,
```
phoracek: We still have generation check in the predicate, could you keep the test?

qinqon: ack.

phoracek:
```
pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go:38:13: Sprintf format %d has arg m.nodeName of wrong type string

Makefile:53: recipe for target 'vet' failed
```

qinqon:
@phoracek fixed the vet error can you recheck?

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -29,7 +30,24 @@ func NewManager(client client.Client, nodeName string, policy nmstatev1alpha1.No
 	manager.logger = logf.Log.WithName("policy/conditions/manager").WithValues("enactment", nmstatev1alpha1.EnactmentKey(nodeName, policy.Name).Name)
 	return manager
 }
-
+func (m *Manager) NotifyNodeSelectorNotMatching(unmatchingLabels map[string]string) {
+	message := ""
+	if len(unmatchingLabels) > 0 {
+		message = fmt.Sprintf("Unmatching labels: %v", unmatchingLabels)
+	} else {
+		message = fmt.Sprintf("Cannot retrieve node %s", m.nodeName)
```
phoracek: This error is completely implicit. Meaning of zero unmatchingLabels should not be "cannot retrieve node".

test/e2e/utils.go
```diff
@@ -215,10 +215,16 @@ func deleteNodeNeworkStates() {
 }
 
 func deletePolicy(name string) {
+	By(fmt.Sprintf("deleting policy %s", name))
 	policy := &nmstatev1alpha1.NodeNetworkConfigurationPolicy{}
 	policy.Name = name
 	err := framework.Global.Client.Delete(context.TODO(), policy)
 	Expect(err).ToNot(HaveOccurred())
+	policy = nil
```
phoracek: Please insert an empty line above.

test/e2e/utils.go
```diff
@@ -215,10 +215,16 @@ func deleteNodeNeworkStates() {
...
+	Eventually(func() *nmstatev1alpha1.NodeNetworkConfigurationPolicy {
+		framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, policy)
```
phoracek: You should check the return code and see if it is NotFound. The state of `policy` is undefined in case the function fails. Then you can drop the `policy = nil`

test/e2e/utils.go
```diff
@@ -215,10 +215,16 @@ func deleteNodeNeworkStates() {
 }
 
 func deletePolicy(name string) {
+	By(fmt.Sprintf("deleting policy %s", name))
```
phoracek: Capital D

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager_test.go
```diff
@@ -0,0 +1,140 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   c.ObjNodeName,
+					Labels: c.NodeLabels,
+				},
+			}
+
+			policy := nmstatev1alpha1.NodeNetworkConfigurationPolicy{
+				Spec: nmstatev1alpha1.NodeNetworkConfigurationPolicySpec{
+					NodeSelector: c.NodeSelector,
+					DesiredState: nmstatev1alpha1.State{Raw: nil}, // TODO
```
phoracek: What is the TODO action?

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package selectors
```
phoracek: The name of the file should be selectors_suite_test since it is per package

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestUnit(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Policy Node Selectors Test Suite")
```
phoracek: Let's keep this generic (since the package name is generic) and make it Policy Selectors Test Suite

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type Manager struct {
+	client   client.Client
+	policy   nmstatev1alpha1.NodeNetworkConfigurationPolicy
+	nodeName string
+	logger   logr.Logger
+}
+
+func NewManager(client client.Client, nodeName string, policy nmstatev1alpha1.NodeNetworkConfigurationPolicy) Manager {
```
phoracek: Could you please elaborate why did you chose to have a stateful manager for this over a simple input-output function? That would be easier to test, all we need is `func Matches(policy, nodeName) (bool, string, error) {}`. Bool for whether it matches, string for the message `"All matches"`/`"Following does not match:\n..."` and error for unexpected failures. Maybe I'm missing something and there is a good reason to have a manager abstratction. Otherwise, I prefer simplicity.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: Let's refactor `MatchesThisNode` so it returns an error instead of bool, so if unmatching labels is empty but we  have an error we print that on the condition.

test/e2e/utils.go
qinqon: Originally I was checking it the proper way with apierrors.IsNotFound, but looks like the client form operator-sdk e2e test framework is not correctly propagating the error so it does not match, we can try to debug it though, the only way I found for it to work is checking the policy, but I am not happy with it either.

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager_test.go
qinqon: To be honest I don't remember, maybe it's related to advanced policy matchers, let's just remove DesiredState altogether since it's not needed for this.

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager.go
qinqon: The approach I was taking is having a Matching method per type, for example:
- `UnmatchedLabels`
- `UnmatchedNodeAffinity`
- `UnmatchedNetworkAffinity`

They don't do string formating inside they just return what is not matching and the conditions.Manager is the one that know how to format that into a condition.

The controller use ther results from selectors.Manager and calls conditions.Manager with them.

So controller calls them and pass result to conditions.Manager functions to compose proper condition.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
phoracek: Smart

test/e2e/utils.go
phoracek: Could you try to use basic client-go instead to see if it is really an issue of the operator-sdk client? I find this disturbing

pkg/controller/nodenetworkconfigurationpolicy/selectors/manager.go
phoracek: It is good to look forward to the new features, but I have a question. What would those method share that would be a part of the context. Could they live under separated packages (to limit the scope, make modules shorter and easier to learn and grasp the logic).

I also don't get the relation to conditions.Manager you explained. This module and conditions should live without knowing about one another, they should be connected through the main module.

++ for the UnmatchedLabels function, I like it.

qinqon: > It is good to look forward to the new features, but I have a question. What would those method share that would be a part of the context. 

The three of them will need client/policy/nodeName and that's what Manager struct has more or less.

> Could they live under separated packages (to limit the scope, make modules shorter and easier to learn and grasp the logic).

If we move to modules we cannot create methods there for Manager since they are not local to module, we can just split manager.go at labels.go, node_affinity.go, network_affinity.go and leave at manager.go only the struct.

qinqon: Also I don't really like the work Manager for a struct, since at golang they don't own the methods, they are just structs, their name has to be associated with the content.

phoracek: > If we move to modules we cannot create methods there for Manager since they are not local to module

I know, I was more in favor of just dropping the manager and keep simple functions. However, if you believe the manager makes the code simpler to read, let's keep it. About the name, maybe express that you are working with selectors in it?

qinqon: @phoracek maybe this https://github.com/nmstate/kubernetes-nmstate/pull/285/commits/4bb973b6a93f69896ac82dc0861432c3b4e2e3c3 make more sense ?

pkg/controller/nodenetworkconfigurationpolicy/selectors/request.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/go-logr/logr"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+type Request struct {
```
phoracek: Oh this is good!

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -196,6 +166,19 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 	}
 
 	conditionsManager := conditions.NewManager(r.client, nodeName, *instance)
+	selectors := selectors.New(r.client, nodeName, *instance)
```
phoracek: The naming is confusing. We don't get a new selector here. Can we leverage the `Resource` struct here?

pkg/controller/nodenetworkconfigurationpolicy/selectors/labels.go
```diff
@@ -0,0 +1,29 @@
+package selectors
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func matches(nodeSelector map[string]string, labels map[string]string) map[string]string {
```
phoracek: please rename the function properly

pkg/controller/nodenetworkconfigurationpolicy/selectors/labels_test.go
```diff
@@ -0,0 +1,143 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var _ = Describe("NodeNetworkConfigurationPolicy controller selectors", func() {
+	var ExpectedNode = "node01"
```
phoracek: why capital E?

pkg/controller/nodenetworkconfigurationpolicy/selectors/request.go
```diff
@@ -0,0 +1,27 @@
...
+	client   client.Client
+	policy   nmstatev1alpha1.NodeNetworkConfigurationPolicy
+	nodeName string
+	logger   logr.Logger
+}
+
+func New(client client.Client, nodeName string, policy nmstatev1alpha1.NodeNetworkConfigurationPolicy) Request {
```
phoracek: NewRequest. `New` in the scope of `selectors` is too vague.

test/e2e/utils.go
qinqon: Ok checking it correctly again works, something was not good at previous code.

test/e2e/utils.go
```diff
@@ -215,10 +215,17 @@ func deleteNodeNeworkStates() {
 }
 
 func deletePolicy(name string) {
+	By(fmt.Sprintf("Deleting policy %s", name))
 	policy := &nmstatev1alpha1.NodeNetworkConfigurationPolicy{}
 	policy.Name = name
 	err := framework.Global.Client.Delete(context.TODO(), policy)
 	Expect(err).ToNot(HaveOccurred())
+
+	policy = nil
```
phoracek: we don't need this, you can `Get` directly into anonymous `&nmstatev1alpha1.NodeNetworkConfigurationPolicy{}`.

pkg/controller/nodenetworkconfigurationpolicy/selectors/selectors.go
```diff
@@ -0,0 +1,25 @@
...
+type Selectors struct {
+	client client.Client
+	policy nmstatev1alpha1.NodeNetworkConfigurationPolicy
+	logger logr.Logger
+}
+
+func NewFromPolicy(client client.Client, policy nmstatev1alpha1.NodeNetworkConfigurationPolicy) Selectors {
+	selectors := Selectors{
+		client: client,
+		policy: policy,
```
phoracek: If we introduce `NewFromSomethingElse`, this would not work again. The `Selectors` object should have an attribute `nodeSelectors` and this attribute should be populated in this constructor.

qinqon: We are not going to do NewFromSomethingElse

phoracek: I want to limit the structure to minimum that is needed. Small scopes are easier to read and debug. But I won't block this PR anymore, this is just a nit a maybe a matter of taste.

phoracek:
We don't have OWNERS in the repo. Gotta add it. Merging manually.

PR #290, opened 2019-12-05
check for trailing whitespaces

phoracek:
gofmt checks that there are no trailing whitespaces in .go modules,
we should do something similar for all files under the source tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,13 +44,19 @@ description = build/_output/description
 resources = deploy/namespace.yaml deploy/service_account.yaml deploy/role.yaml deploy/role_binding.yaml
 all: check handler
 
-check: format vet
+check: format vet whitespace-check
 
-format:
+format: whitespace-format
 	gofmt -d cmd/ pkg/
 
 vet:
 	go vet ./cmd/... ./pkg/...
+
+whitespace-format:
+	hack/whitespace.sh format
+
+whitespace-check:
+	hack/whitespace.sh check
 
 $(GINKGO): go.mod
 	GOBIN=$$(pwd)/$(BIN_DIR) go install ./vendor/github.com/onsi/ginkgo/ginkgo
@@ -177,4 +183,6 @@ tools-vendoring:
 	cluster-sync-handler \
 	cluster-sync \
 	cluster-clean \
-	release
+	release \
+	whitespace-check \
+	whitespace-format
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -44,14 +44,20 @@ description = build/_output/description
...
+check: format vet whitespace-check
 
-format:
+format: whitespace
 	gofmt -d cmd/ pkg/
 
 vet:
 	go vet ./cmd/... ./pkg/...
 
+whitespace:
```
qinqon: make whitespace-fix or make whitespace-format

hack/whitespace.sh
```diff
@@ -0,0 +1,28 @@
...
+# Check for trailing whitespaces in all tracked files.
+#
+# Usage:
+# hack/whitespace.sh # Check for trailing whitespaces
+# hack/whitespace.sh --fix # Drop trailing whitespaces
+
+set -e
+
+function fix() {
+    git ls-files -- ':!vendor/' | xargs sed --follow-symlinks -i 's/[[:space:]]*$//'
```
qinqon: can we put in a function `git ls-files -- ':!vendor/'` command we are using it twice.

hack/whitespace.sh
```diff
@@ -0,0 +1,28 @@
...
+        echo "$invalid_files"
+        return 1
+    fi
+}
+
+if [ "$1" == "--fix" ]; then
+    fix
+else
+    check
+fi
```
qinqon: what about just
$1
so we do `whitespace check` or `whitespace fix`

hack/whitespace.sh
phoracek: The rest of the line is different, I would rather keep it than have a function which would have a name practically saying the same.

hack/whitespace.sh
```diff
@@ -0,0 +1,31 @@
...
+        echo "$invalid_files"
+        return 1
+    fi
+}
+
+if [ "$1" == "format" ]; then
+    format
+elif [ "$1" == "check" ]; then
+    check
+else
```
qinqon: if [[ "$1" =~ (check|format) ]]; then
   $1
else

phoracek: That saves line but pushes API from one layer (user params) to another (internal naming of functions). I'd prefer keeping the mapping as a dividing line between the two.

qinqon: ack

PR #292, opened 2019-12-05
Add overall policy conditions

qinqon:


qinqon:
One important thing that it's missing on the wip, on conflict we have to recheck enactments not only re-update status, will do when back from PTO.

PR #299, opened 2019-12-11
bump used kubevirtci version to fix cert issues

phoracek:


phoracek:
/close

Although newer kubevirtci helps a little, it still does not pass. Let's try this again later.

PR #300, opened 2019-12-11
Policy overall conditions

### LINKED ISSUE #131: Add status conditions to NodeNetworkConfigurationPolicy
It has to contain applied per node at status.

qinqon:
Fixes #131


phoracek:
@qinqon \honk does not work :/

phoracek:
@qinqon also please rebase.

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
```diff
@@ -130,3 +130,30 @@ func setEnactmentMatching(conditions *nmstatev1alpha1.ConditionList, message str
 		message,
 	)
 }
+
+type CountByConditionType map[nmstatev1alpha1.ConditionType]int
+
+func enactmentsCount(enactments nmstatev1alpha1.NodeNetworkConfigurationEnactmentList) CountByConditionType {
+	trueConditionsFrequency := CountByConditionType{}
```
phoracek: Again, not a frequency IMO, what about just countByConfigurationType?

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -90,3 +97,59 @@ func (m *Manager) updateEnactmentConditions(
 		return nil
 	})
 }
+
+func (m *Manager) refreshPolicyConditions() error {
+
```
phoracek: Please drop the empty line

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -16,12 +17,12 @@ import (
...
 type Manager struct {
 	client   client.Client
-	policy   nmstatev1alpha1.NodeNetworkConfigurationPolicy
+	policy   *nmstatev1alpha1.NodeNetworkConfigurationPolicy
 	nodeName string
 	logger   logr.Logger
 }
 
-func NewManager(client client.Client, nodeName string, policy nmstatev1alpha1.NodeNetworkConfigurationPolicy) Manager {
+func NewManager(client client.Client, nodeName string, policy *nmstatev1alpha1.NodeNetworkConfigurationPolicy) Manager {
```
phoracek: Only now I see how vague it is. we should make clear that this manages conditions in the Enactment (therefore it takes nodeName as an argument). Could you rename it to NewEnactmentConditionsManager? And the strcuture as well? Or maybe to follow the schema from previous PR, let's call the structure `EnactmentConditions`.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -90,3 +97,59 @@ func (m *Manager) updateEnactmentConditions(
...
 }
+
+func (m *Manager) refreshPolicyConditions() error {
+
+	// On conflict we need to re-retrieve enactments since the
+	// conflict can denote that the calculated policy conditions
+	// are now not accurate.
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+
+		enactments := nmstatev1alpha1.NodeNetworkConfigurationEnactmentList{}
```
phoracek: Let's stop here for a second. We have retry on conflict, so the overall conditions we read are for sure latest. Do we need to fetch all the enactments here or can we do the calculation based only on what we have in the overall conditions and our conditions?

I'm too tired to tell for sure, but I feel there may be a way to do the calculation without fetching all the Enactments. Some fun programming finally!

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
```diff
@@ -90,3 +97,59 @@ func (m *Manager) updateEnactmentConditions(
...
+		policyLabelFilter := client.MatchingLabels{"policy": m.policy.Name}
```
phoracek: If we do use labels, please namespace them: nmstate.io/policy or nmstate.io/configurationPolicy (configurationEnactment and state)

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -79,6 +79,21 @@ func NewEnactment(node corev1.Node, policy NodeNetworkConfigurationPolicy) NodeN
 	return enactment
 }
 
+func (enactments NodeNetworkConfigurationEnactmentList) TrueConditionsFrequency() map[ConditionType]int {
```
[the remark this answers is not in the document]
qinqon: I was suppose to remove this, at the end the word "frequency" was not a good idea.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -46,10 +46,28 @@ type NodeNetworkConfigurationPolicySpec struct {
...
+)
+
+var NodeNetworkConfigurationPolicyConditionTypes = [...]ConditionType{
+	NodeNetworkConfigurationPolicyConditionAvailable,
+	NodeNetworkConfigurationPolicyConditionDegraded,
+	NodeNetworkConfigurationPolicyConditionProgressing,
+}
+
+const (
+	NodeNetworkConfigurationPolicyConditionFailedToConfigure            ConditionReason = "FailedToConfigure"
```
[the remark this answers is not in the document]
qinqon: If you have a cluster with failing partial or total you have to do same actions, I don't see any benefict on differenciate it, portant thing is to show stuff so users can know where to look (affected enactments and the like).

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactments.go
qinqon: Fixing.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: Well we are updating policy conditions too, but maybe that was not a good idea, let's talk about this one.

pkg/controller/nodenetworkconfigurationpolicy/conditions/manager.go
qinqon: This would be kind of the algorithm in case we don't wan to count always:

NotifyFailure:
- If overall is failure we change the counter
- If overall is progressing we notify failure
- if it's success we have do something wrong
NotifyProgress:
- if overall is failure we do nothing
- if overall is progressing we change the counter.
- if it's success we have do something wrong
NotifySuccess:
- We have to look at the whole cluster and change to success is everything is allright.
NotifyNotMatching:
- if overall is failure we do nothing
- if overall is progressing we decrease the counter 
- if overall is success we did something wrong
- else we we have to look at the whole cluster and set sucess to false.

I see this more complex than a refresh function at controller (selecting it with leader election) with a refresh timer.

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy.go
```diff
@@ -0,0 +1,166 @@
...
+		// Let's get conditions with true status frequency
+		enactmentsCount := enactment.Count(enactments)
+
+		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount.String()))
+
+		if enactmentsCount.Failed() > 0 {
+			message := fmt.Sprintf("%d/%d nodes failed to configure", enactmentsCount.Failed(), numberOfNodesAffected)
+			setPolicyFailedToConfigure(&policy.Status.Conditions, message)
+
+		} else if enactmentsCount.Progressing() > 0 {
```
phoracek: I think this is redundant. We don't care if any enactment is "Progressing". As long as not all Nodes have Enactment and not all Enactments are either Degraded or Available, the Policy is progressing. Plus, I think that the information on how many nodes it is Available, Degraded and YetToMatch is just as interesting for the user to know.

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy.go
```diff
@@ -0,0 +1,166 @@
...
+
+		enactments := nmstatev1alpha1.NodeNetworkConfigurationEnactmentList{}
+		policyLabelFilter := client.MatchingLabels{nmstatev1alpha1.ENACTMENT_POLICY_LABEL: policy.Name}
+		err := cli.List(context.TODO(), &enactments, policyLabelFilter)
+		if err != nil {
+			return errors.Wrap(err, "geting enactments failed")
+		}
+
+		nodes := corev1.NodeList{}
+		nodeSelectorFilter := client.MatchingLabels(policy.Spec.NodeSelector)
```
phoracek: This scares me (and at the same time is super cool for what we have now). It works now because of all selectors we need are NodeSelectors... However, let's roll with this. We can address improving this once we have a reason (other selectors).

Just one question. What happens with `nil` NodeSelector?

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy_test.go
```diff
@@ -0,0 +1,181 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	enactmentconditions "github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/conditions/enactment"
+)
+
+type Enactment struct {
```
phoracek: Please make this private. The module is public (it is not `package conditions_test`), we don't want to expose these structures.

Also, functions in this package work with real types. Could you just use those here? Then you can test the labels an everything. But maybe I'm missing something and this way is better for testing?

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -38,6 +38,7 @@ type NodeNetworkConfigurationEnactmentStatus struct {
 }
 
 const (
+	ENACTMENT_POLICY_LABEL                                              = "nmstate.io/policy"
```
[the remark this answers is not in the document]
qinqon: My bad.

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactment/notifier.go
```diff
@@ -0,0 +1,220 @@
...
+
+		err = n.client.Status().Update(context.TODO(), instance)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+}
+
+func SetFailedToConfigure(conditions *nmstatev1alpha1.ConditionList, message string) {
```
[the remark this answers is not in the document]
qinqon: We use them at at policyconditions tests and they are at different package.

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy.go
```diff
@@ -0,0 +1,166 @@
+package conditions
```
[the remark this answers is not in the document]
qinqon: First I started with packaging at policyconditions and enactmentconditions and having two `notifiers` then I realize we were allready behind `nodenetworkconfigurationpolicy` so we are under policy that's why a I changed to this structure, but I prefer to go back to `policyconditions`and `enactmentconditions`.

test/e2e/conditions.go
```diff
@@ -73,3 +73,40 @@ func enactmentConditionsStatusEventually(node string) AsyncAssertion {
...
+	return obtainedConditions
+}
+
+func policyConditionsStatusForPolicyEventually(policy string) AsyncAssertion {
+	return Eventually(func() nmstatev1alpha1.ConditionList {
+		return policyConditionsStatus(policy)
+	}, 180*time.Second, 1*time.Second)
+}
+
+func policyConditionsStatusForPolicyConsistently(policy string) AsyncAssertion {
```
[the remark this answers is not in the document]
qinqon: Because golagn does not support function override, we have two functions one that uses "test-policy" and the other that specify the policy you want to use.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -165,33 +166,39 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 		log.Error(err, "Error initializing enactment")
 	}
 
-	conditionsManager := conditions.NewManager(r.client, nodeName, *instance)
+	enactmentConditionsNotifier := enactmentconditions.NewNotifier(r.client, nodeName, instance)
```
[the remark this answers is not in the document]
qinqon: Agree with all, I am just going to pass the enactment name.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -165,33 +166,39 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
...
+
+	// Policy conditions will be updated at the end so updating it
+	// does not impact at applyting state, it will increase just
+	// reconcile time.
+	defer conditions.UpdatePolicyConditions(r.client, instance)
```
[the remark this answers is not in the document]
qinqon: Originally I started with `enactmentconditions` and `policyconditions` but since we are already under `nodenetworkconfigurationpolicy` package I did the current approach, but I don't like it either, gong back to original.

test/e2e/conditions.go
```diff
@@ -73,3 +73,40 @@ func enactmentConditionsStatusEventually(node string) AsyncAssertion {
...
+func policyConditionsStatus(policyName string) nmstatev1alpha1.ConditionList {
+	policy := nodeNetworkConfigurationPolicy(types.NamespacedName{Name: policyName})
+	obtainedConditions := nmstatev1alpha1.ConditionList{}
+	for _, policysConditionType := range nmstatev1alpha1.NodeNetworkConfigurationPolicyConditionTypes {
+		obtainedCondition := policy.Status.Conditions.Find(policysConditionType)
+		obtainedConditionStatus := corev1.ConditionUnknown
+		if obtainedCondition != nil {
+			obtainedConditionStatus = obtainedCondition.Status
+		}
+		obtainedConditions = append(obtainedConditions, nmstatev1alpha1.Condition{
```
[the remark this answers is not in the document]
qinqon: An alternative to this is create at ginkgo custom matcher that check that do the Unknown check logic, and that will add even more logic to test framework.

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy.go
qinqon: > A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic

From https://blog.golang.org/go-maps-in-action so we are good there if nil.

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy_test.go
qinqon: Those structs are not needed, I am going to just use nncp and nnce.

test/e2e/conditions.go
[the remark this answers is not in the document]
phoracek: I was just proposing change of the naming

pkg/controller/nodenetworkconfigurationpolicy/conditions/enactment/notifier.go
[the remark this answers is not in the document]
phoracek: I'm not a fan of making functions public just for the sake of testing. Can't we adjust tests so they respect these functions being private?

pkg/controller/nodenetworkconfigurationpolicy/conditions/policy.go
```diff
@@ -0,0 +1,166 @@
...
+	"k8s.io/client-go/util/retry"
+	client "sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	enactment "github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/conditions/enactment"
+)
+
+var (
+	log = logf.Log.WithName("policy")
```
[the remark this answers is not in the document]
qinqon: I think there is no easy way to retrieve file and line number with k8s logging system.

PR #302, opened 2019-12-13
Reconcile nodes 1 by 1 with leader lock

HlinaCZ:
Signed-off-by: HlinaCZ <djhlinacz@gmail.com>

### DIFF
--- a/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
@@ -28,6 +28,7 @@ import (
 	"github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/policyconditions"
 	"github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/selectors"
 	nmstate "github.com/nmstate/kubernetes-nmstate/pkg/helper"
+	"github.com/nmstate/kubernetes-nmstate/pkg/rollout"
 )
 
 var (
@@ -64,7 +65,16 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileNodeNetworkConfigurationPolicy{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	rollout, err := rollout.NewRollout(mgr.GetConfig(), mgr.GetScheme())
+	if err != nil {
+		panic("failed to create roullout manager " + err.Error())
+	}
+
+	return &ReconcileNodeNetworkStateConfiguration{
+		client:  mgr.GetClient(),
+		scheme:  mgr.GetScheme(),
+		rollout: rollout,
+	}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -91,8 +101,9 @@ var _ reconcile.Reconciler = &ReconcileNodeNetworkConfigurationPolicy{}
 type ReconcileNodeNetworkConfigurationPolicy struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
+	client  client.Client
+	scheme  *runtime.Scheme
+	rollout *rollout.Rollout
 }
 
 func (r *ReconcileNodeNetworkConfigurationPolicy) waitEnactmentCreated(policy nmstatev1alpha1.NodeNetworkConfigurationPolicy) error {
@@ -187,7 +198,10 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 
 	enactmentConditions.NotifyMatching()
 
-	enactmentConditions.NotifyProgressing()
+	unlock := r.rollout.Lock()
+	defer unlock()
+
+	conditionsManager.NotifyProgressing()
 	nmstateOutput, err := nmstate.ApplyDesiredState(instance.Spec.DesiredState)
 	if err != nil {
 		errmsg := fmt.Errorf("error reconciling NodeNetworkConfigurationPolicy at desired state apply: %s, %v", nmstateOutput, err)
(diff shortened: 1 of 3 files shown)


ovirt-infra:

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -180,6 +191,9 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 
 	conditionsManager.NotifyMatching()
 
+	unlock := r.rollout.Lock()
+	defer unlock()
+
 	conditionsManager.NotifyProgressing()
```
qinqon: I think we have to mark enactments as progressing while they are lock.

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,87 @@
...
+// NewRollout creates new Rollou
+func NewRollout(cfg *rest.Config, scheme *runtime.Scheme) (*Rollout, error) {
+	startedLeadingChannel := make(chan interface{})
+
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize apiserver client: %v", err)
+	}
+
+	recorderProvider, err := NewProvider(cfg, scheme, log.WithName("record_provider"))
```
qinqon: Why we don't pass directly clientset to NewRollout and NewProvider ?

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,87 @@
...
+		LeaseDuration: 3 * time.Second,
+		RenewDeadline: 2 * time.Second,
+		RetryPeriod:   1 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(_ context.Context) {
+				log.Info("Node is locked")
+				startedLeadingChannel <- true
+			},
+			OnStoppedLeading: func() {
+				log.Info("Node is unlocked")
```
qinqon: Do we have to close(startedLeadingChannel) here ?

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -63,7 +64,16 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileNodeNetworkConfigurationPolicy{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	rollout, err := rollout.NewRollout(mgr.GetConfig(), mgr.GetScheme())
```
qinqon: I would prefer to remove rollout as a member of ReconcileNodeNetworkStateConfiguration since we are calling it at `Reconcile` only and in case of error it will re-reconcile, so we don't have to panic

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,87 @@
...
+	id := uuid.New().String()
+	lock, err := resourcelock.New(resourcelock.ConfigMapsResourceLock, "nmstate", "leader-lock", clientset.CoreV1(), resourcelock.ResourceLockConfig{
+		Identity:      id,
+		EventRecorder: recorderProvider.GetEventRecorderFor(id),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create resource lock: %v", err)
+	}
+
+	// create new leader elector
```
qinqon: Create the `startedLeadingChannel` channel here so we don't have to close on the othe errors

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,87 @@
...
+			OnStartedLeading: func(_ context.Context) {
+				log.Info("Node is locked")
+				startedLeadingChannel <- true
+			},
+			OnStoppedLeading: func() {
+				log.Info("Node is unlocked")
+			},
+		},
+	})
+	if err != nil {
```
qinqon: close `startedLeadingChannel` here

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -180,6 +191,9 @@ func (r *ReconcileNodeNetworkConfigurationPolicy) Reconcile(request reconcile.Re
 
 	conditionsManager.NotifyMatching()
 
+	unlock := r.rollout.Lock()
+	defer unlock()
```
qinqon: We don't want to unlock at the end `Reconcile` we want to unlock at the end of apply so we update status in parallel.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: Do we want to mark as `Progressing` all nodes that are waiting to get the leadership or only the one that is the leader at the moment? I would vote for the latter for the nice effect...

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
phoracek: In other words, call the `unlock` right after `ApplyDesiredState` line.

pkg/rollout/rollout.go
phoracek: NewProvider is a copy paste, but you are right, we could make it so it takes the clientSet directly. @HlinaCZ could you adjust our copied module, so it takes the clientset we create here instead of cnv and scheme?

For NewRollout, the reason is, we don't need client set in layers above, only here, so I think it is better to just pass what we have and let NewRollout create the clientset. We can later think if we can get rid of it completely.

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,87 @@
...
+		EventRecorder: recorderProvider.GetEventRecorderFor(id),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create resource lock: %v", err)
+	}
+
+	// create new leader elector
+	leaderElector, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
+		Lock: lock,
+		// ReleaseOnCancel: true, // TODO: unknown field
```
phoracek: Seems like we don't need this, so let's drop it

pkg/rollout/rollout.go
phoracek: Why? The constructor failed, we return `nil`, there is no users of the channel, it should get garbage collected, no?

pkg/rollout/rollout.go
HlinaCZ: We can pass clientset directly but we still have to pass scheme because `NewProvider` uses scheme for func `GetEventRecorderFor()`.

PR #306, opened 2019-12-16
Use matching state at policy condition

qinqon:

Current policy conditions calculations ignores the Matching state at enactments and just apply the nodeSelector at policy to retrieve the expected nodes it has to run on, that is not correct since we are planning on use other filters different than nodeSelector and in that case calculation would be wrong.

This PR use the Matching state at enactments to do the policy conditions check, it also increase the interface for enacments counters so we have the false, negative and unknown counters of each condition type.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions/counter.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions/counter.go
@@ -8,33 +8,80 @@ import (
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-type CountByConditionType map[nmstatev1alpha1.ConditionType]int
+type CountByConditionStatus map[corev1.ConditionStatus]int
 
-func Count(enactments nmstatev1alpha1.NodeNetworkConfigurationEnactmentList) CountByConditionType {
-	trueConditionsCount := CountByConditionType{}
-	for _, enactment := range enactments.Items {
-		for _, conditionType := range nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionTypes {
+type ConditionCount map[nmstatev1alpha1.ConditionType]CountByConditionStatus
+
+func Count(enactments nmstatev1alpha1.NodeNetworkConfigurationEnactmentList) ConditionCount {
+	conditionCount := ConditionCount{}
+	for _, conditionType := range nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionTypes {
+		conditionCount[conditionType] = CountByConditionStatus{}
+		for _, enactment := range enactments.Items {
 			condition := enactment.Status.Conditions.Find(conditionType)
 			if condition != nil {
-				if condition.Status == corev1.ConditionTrue {
-					trueConditionsCount[conditionType] += 1
-				}
+				conditionCount[conditionType][condition.Status] += 1
+			} else {
+				conditionCount[conditionType][corev1.ConditionUnknown] += 1
 			}
 		}
 	}
-	return trueConditionsCount
+	return conditionCount
 }
 
-func (c CountByConditionType) Failed() int {
+func (c ConditionCount) failed() CountByConditionStatus {
 	return c[nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionFailing]
 }
-func (c CountByConditionType) Progressing() int {
+func (c ConditionCount) progressing() CountByConditionStatus {
 	return c[nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionProgressing]
 }
-func (c CountByConditionType) Available() int {
+func (c ConditionCount) available() CountByConditionStatus {
 	return c[nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionAvailable]
 }
+func (c ConditionCount) matching() CountByConditionStatus {
+	return c[nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionMatching]
+}
 
-func (c CountByConditionType) String() string {
-	return fmt.Sprintf("{failed: %d, progressing: %d, available: %d}", c.Failed(), c.Progressing(), c.Available())
+func (c CountByConditionStatus) true() int {
+	return c[corev1.ConditionTrue]
+}
+
+func (c CountByConditionStatus) false() int {
+	return c[corev1.ConditionFalse]
+}
+
+func (c CountByConditionStatus) unknown() int {
+	return c[corev1.ConditionUnknown]
+}
+
+func (c ConditionCount) Failed() int {
+	return c.failed().true()
+}
+func (c ConditionCount) NotFailed() int {
+	return c.failed().false()
+}
+func (c ConditionCount) Progressing() int {
+	return c.progressing().true()
+}
+func (c ConditionCount) NotProgressing() int {
+	return c.progressing().false()
+}
+func (c ConditionCount) Available() int {
+	return c.available().true()
+}
+func (c ConditionCount) NotAvailable() int {
+	return c.available().false()
+}
+func (c ConditionCount) Matching() int {
+	return c.matching().true()
+}
+func (c ConditionCount) NotMatching() int {
+	return c.matching().false()
+}
+
+func (c ConditionCount) String() string {
+	return fmt.Sprintf("{failed: %s, progressing: %s, available: %s, matching: %s}", c.failed(), c.progressing(), c.available(), c.matching())
+}
+
+func (c CountByConditionStatus) String() string {
+	return fmt.Sprintf("{true: %d, false: %d, unknown: %d}", c.true(), c.false(), c.unknown())
 }
--- a/pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
@@ -105,31 +105,39 @@ func Update(cli client.Client, policyKey types.NamespacedName) error {
 		}
 
 		nodes := corev1.NodeList{}
-		nodeSelectorFilter := client.MatchingLabels(policy.Spec.NodeSelector)
-		err = cli.List(context.TODO(), &nodes, nodeSelectorFilter)
+		err = cli.List(context.TODO(), &nodes)
 		if err != nil {
 			return errors.Wrap(err, "getting nodes failed")
 		}
-		numberOfNodesAffected := len(nodes.Items)
+		numberOfReadyNodes := 0
+		for _, node := range nodes.Items {
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == corev1.NodeReady &&
+					condition.Status == corev1.ConditionTrue {
+					numberOfReadyNodes += 1
+				}
+			}
+		}
 
 		// Let's get conditions with true status count
 		enactmentsCount := enactmentconditions.Count(enactments)
 
-		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount.String()))
+		numberOfFinishedEnactments := enactmentsCount.Available() + enactmentsCount.Failed() + enactmentsCount.NotMatching()
 
-		if numberOfNodesAffected == 0 {
-			message := "Policy does not match any node"
-			setPolicyNotMatching(&policy.Status.Conditions, message)
-		} else if enactmentsCount.Failed()+enactmentsCount.Available() == numberOfNodesAffected {
-			if enactmentsCount.Failed() > 0 {
-				message := fmt.Sprintf("%d/%d nodes failed to configure", enactmentsCount.Failed(), numberOfNodesAffected)
+		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount))
+		if numberOfFinishedEnactments < numberOfReadyNodes {
+			setPolicyProgressing(&policy.Status.Conditions, fmt.Sprintf("Policy is progressing %d/%d nodes finished", numberOfFinishedEnactments, numberOfReadyNodes))
+		} else {
+			if enactmentsCount.Matching() == 0 {
+				message := "Policy does not match any node"
+				setPolicyNotMatching(&policy.Status.Conditions, message)
+			} else if enactmentsCount.Failed() > 0 {
+				message := fmt.Sprintf("%d/%d nodes failed to configure", enactmentsCount.Failed(), enactmentsCount.Matching())
 				setPolicyFailedToConfigure(&policy.Status.Conditions, message)
 			} else {
-				message := fmt.Sprintf("%d/%d nodes successfully configured", enactmentsCount.Available(), numberOfNodesAffected)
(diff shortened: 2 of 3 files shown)


phoracek:
I need proper explanation in PR description and commit message before I start reviewing this.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -24,7 +24,7 @@ func setPolicyProgressing(conditions *nmstatev1alpha1.ConditionList, message str
 		nmstatev1alpha1.NodeNetworkConfigurationPolicyConditionDegraded,
 		corev1.ConditionUnknown,
 		nmstatev1alpha1.NodeNetworkConfigurationPolicyConditionConfigurationProgressing,
-		"",
+		message,
```
phoracek: I get why you would look for this message when "unknown", but I'd rather drop it to keep message only in a single condition. Plus it is not related to this PR. Once we have solid basics, we can optimize messages and stuff like that.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -93,31 +93,33 @@ func Update(cli client.Client, policy *nmstatev1alpha1.NodeNetworkConfigurationP
 		}
 
 		nodes := corev1.NodeList{}
-		nodeSelectorFilter := client.MatchingLabels(policy.Spec.NodeSelector)
-		err = cli.List(context.TODO(), &nodes, nodeSelectorFilter)
+		err = cli.List(context.TODO(), &nodes)
```
phoracek: I think we should list only Nodes that are available for scheduling. If we have one faulty node in the cluster, it would keep all Policies from completing.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -93,31 +93,33 @@ func Update(cli client.Client, policy *nmstatev1alpha1.NodeNetworkConfigurationP
...
-		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount.String()))
-
-		if numberOfNodesAffected == 0 {
-			message := "Policy does not match any node"
-			setPolicyNotMatching(&policy.Status.Conditions, message)
-		} else if enactmentsCount.Failed()+enactmentsCount.Available() == numberOfNodesAffected {
-			if enactmentsCount.Failed() > 0 {
-				message := fmt.Sprintf("%d/%d nodes failed to configure", enactmentsCount.Failed(), numberOfNodesAffected)
+		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount))
+		if numberOfEnactments < numberOfNodes {
```
phoracek: We should not forget to add the NNCE cleanup to Node controller. Otherwise, when someone removes a node, we would return here too quick.

Also, using only the number of NNCEs makes me unsure. Can we just check if `count.Available() + count.Degraded() + count.NotMatching() == numberOfNodes` here? Then you can get rid of the last `else` and move the nested `if else` block one layer higher.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -93,31 +93,33 @@ func Update(cli client.Client, policy *nmstatev1alpha1.NodeNetworkConfigurationP
...
+			setPolicyProgressing(&policy.Status.Conditions, fmt.Sprintf("Policy is progresssing %d/%d enactments not ready", numberOfNodes-numberOfEnactments, numberOfNodes))
```
phoracek: Typically, you wait until the progress bar reaches 10/10. Let's make this "X/Y enactments finished".

qinqon: is not finishe is created, I will change to that.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
qinqon: I like it we get the progressing at only one place, although is `numberOfFinished < numberOfNodes`

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
phoracek: If you follow the comment above about change of conditions, it becomes "Finished", aka all NNCEs are either Degraded or Available. No?

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
phoracek: Ah, you are right of course :)

qinqon:
One test has failed I think it's related to timing, since I see `enactmentconditions	enactment updated at the node: false	{"enactment": "node01.br4"}` we may need to increase those Eventuallies.
```
2019-12-18T12:06:41.136Z	DEBUG	controller_nodenetworkstate	Reconciling NodeNetworkState	{"Request.Namespace": "", "Request.Name": "node01"}
2019-12-18T12:06:41.179Z	DEBUG	controller_node	Reconciling Node	{"Request.Namespace": "", "Request.Name": "node01"}
2019-12-18T12:06:42.890Z	INFO	controller_nodenetworkconfigurationpolicy	Reconciling NodeNetworkConfigurationPolicy	{"Request.Namespace": "", "Request.Name": "br4"}
2019-12-18T12:06:42.891Z	INFO	enactmentconditions	NotifyMatching	{"enactment": "node01.br4"}
2019-12-18T12:06:42.902Z	INFO	enactmentconditions	enactment updated at the node: false	{"enactment": "node01.br4"}

• Failure [181.351 seconds]
NodeSelector
/tmp/kubernetes-nmstate/go/path/src/github.com/nmstate/kubernetes-nmstate/test/e2e/node_selector_test.go:12
  when policy is set with node selector not matching any nodes
  /tmp/kubernetes-nmstate/go/path/src/github.com/nmstate/kubernetes-nmstate/test/e2e/node_selector_test.go:15
    and we remove the node selector
    /tmp/kubernetes-nmstate/go/path/src/github.com/nmstate/kubernetes-nmstate/test/e2e/node_selector_test.go:45
      should update all nodes and have Matching enactment state [It]
      /tmp/kubernetes-nmstate/go/path/src/github.com/nmstate/kubernetes-nmstate/test/e2e/node_selector_test.go:50

      Timed out after 180.000s.
      Expected
          <[]string | len:7, cap:8>: ["cni0", "docker0", "eth0", "eth1", "eth2", "flannel.1", "lo"]
      to contain element matching
          <string>: br4


```

qinqon:
One thing we can do is to first check state then check interfaces.

phoracek:
/hold

Until we resolve the race

test/e2e/rollback_test.go
```diff
@@ -32,7 +34,7 @@ routes:
 `, nic, address, nic))
 }
 
-var _ = Describe("rollback", func() {
+var _ = FDescribe("rollback", func() {
```
phoracek: F?!

qinqon:
Something is not right with prow, it's failing at the first test because default gw is not there
```
econciling NodeNetworkConfigurationPolicy at desired state apply: , rollback cause: Impossible to retrieve default gw, state: ---
dns-resolver:
  config:
    search: []
    server: []
  running: {}
route-rules:
  config: []
routes:
  config: []
  running:
  - destination: 10.244.0.0/24
    metric: 0
    next-hop-address: ''
    next-hop-interface: cni0
    table-id: 254
  - destination: 172.17.0.0/16
    metric: 0
    next-hop-address: ''
    next-hop-interface: docker0
    table-id: 254
  - destination: fe80::/64
    metric: 256
    next-hop-address: ''
    next-hop-interface: cni0
    table-id: 254
  - destination: ff00::/8
    metric: 256
    next-hop-address: ''
    next-hop-interface: cni0
    table-id: 255
```

qinqon:
I am going to split this into other PR that stabilize e2e tests, then put this on top.

qinqon:
/override 
pull-bridge-maker-test

qinqon:
This is failing again at missing default gw:

```
, rollback cause: Impossible to retrieve default gw, state: ---
dns-resolver:
  config:
    search: []
    server: []
  running: {}
route-rules:
  config: []
routes:
  config: []
  running:
  - destination: 10.244.0.0/24
    metric: 0
    next-hop-address: ''
    next-hop-interface: cni0
    table-id: 254
  - destination: 172.17.0.0/16
    metric: 0
    next-hop-address: ''
    next-hop-interface: docker0
    table-id: 254
  - destination: fe80::/64
    metric: 256
    next-hop-address: ''
    next-hop-interface: cni0
    table-id: 254
  - destination: ff00::/8
    metric: 256
    next-hop-address: ''
    next-hop-interface: cni0
    table-id: 255
  - destination: ff00::/8
    metric: 256
    next-hop-address: ''
    next-hop-interface: eth0
    table-id: 255
```

qinqon:
/retest pull-kubernetes-nmstate-unit-test

network hiccup `Failed to download metadata for repo 'updates'`

pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions/counter.go
```diff
@@ -10,19 +10,25 @@ import (
 
 type CountByConditionType map[nmstatev1alpha1.ConditionType]int
 
-func Count(enactments nmstatev1alpha1.NodeNetworkConfigurationEnactmentList) CountByConditionType {
-	trueConditionsCount := CountByConditionType{}
+type ConditionCount map[corev1.ConditionStatus]CountByConditionType
+
+func Count(enactments nmstatev1alpha1.NodeNetworkConfigurationEnactmentList) ConditionCount {
+	conditionCount := ConditionCount{
```
phoracek: Could you please switch it so it is conditionsCount.Available.True instead of conditionsCount.True.Available. The structure seems odd.

pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions/counter.go
```diff
@@ -34,7 +40,51 @@ func (c CountByConditionType) Progressing() int {
 func (c CountByConditionType) Available() int {
 	return c[nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionAvailable]
 }
+func (c CountByConditionType) Matching() int {
+	return c[nmstatev1alpha1.NodeNetworkConfigurationEnactmentConditionMatching]
+}
+
+func (c ConditionCount) True() CountByConditionType {
```
phoracek: Please make sure not to set functions that are not used externally as public. I'd like to limit access to this module as much as possible. It is quite complex.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -105,31 +105,40 @@ func Update(cli client.Client, policyKey types.NamespacedName) error {
...
 
 		nodes := corev1.NodeList{}
-		nodeSelectorFilter := client.MatchingLabels(policy.Spec.NodeSelector)
-		err = cli.List(context.TODO(), &nodes, nodeSelectorFilter)
+		err = cli.List(context.TODO(), &nodes)
 		if err != nil {
 			return errors.Wrap(err, "getting nodes failed")
 		}
-		numberOfNodesAffected := len(nodes.Items)
+		readyNodes := []corev1.Node{}
```
phoracek: We don't need to keep list of nodes, it is not used anywhere. Would you mind just initializing numberOfReadyNodes here and then increment it in the following cycle?

Also, keeping the function that counts ready nodes in a separate function would improve readability.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -105,31 +105,40 @@ func Update(cli client.Client, policyKey types.NamespacedName) error {
...
+				}
+			}
+		}
 
 		// Let's get conditions with true status count
 		enactmentsCount := enactmentconditions.Count(enactments)
 
-		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount.String()))
+		numberOfReadyNodes := len(readyNodes)
+		numberOfFinished := enactmentsCount.Available() + enactmentsCount.Failed() + enactmentsCount.NotMatching()
```
phoracek: numberOfFinishedEnactments

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -105,31 +105,40 @@ func Update(cli client.Client, policyKey types.NamespacedName) error {
...
 
-		if numberOfNodesAffected == 0 {
-			message := "Policy does not match any node"
-			setPolicyNotMatching(&policy.Status.Conditions, message)
-		} else if enactmentsCount.Failed()+enactmentsCount.Available() == numberOfNodesAffected {
-			if enactmentsCount.Failed() > 0 {
-				message := fmt.Sprintf("%d/%d nodes failed to configure", enactmentsCount.Failed(), numberOfNodesAffected)
+		logger.Info(fmt.Sprintf("enactments count: %s", enactmentsCount))
```
phoracek: Can't we use Infof?

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions_test.go
```diff
@@ -36,22 +38,61 @@ func e(node string, policy string, conditionsSetter func(*nmstatev1alpha1.Condit
...
-		Spec: nmstatev1alpha1.NodeNetworkConfigurationPolicySpec{
-			NodeSelector: nodeSelector,
-		},
 		Status: nmstatev1alpha1.NodeNetworkConfigurationPolicyStatus{
 			Conditions: conditions,
 		},
 	}
 }
 
+func n(idx int, conditions []corev1.NodeCondition) corev1.Node {
```
phoracek: newNode

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions_test.go
```diff
@@ -36,22 +38,61 @@ func e(node string, policy string, conditionsSetter func(*nmstatev1alpha1.Condit
...
+func nodeNotReady() []corev1.NodeCondition {
+	return []corev1.NodeCondition{
+		corev1.NodeCondition{
+			Type:   corev1.NodeReady,
+			Status: corev1.ConditionFalse,
+		},
+	}
+}
+
+func nodesCardinality(cardinality int) []corev1.Node {
```
phoracek: newReadyNodes

pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions/counter.go
qinqon: I want to have it symetrical.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
qinqon: There is no Infof

Issue #274, opened 2019-11-26
Defining an IP address to a bond via NodeNetworkConfigurationPolicy will end in a collision


yossisegev:
According to the [bond configuration spec](https://github.com/nmstate/kubernetes-nmstate/blob/master/docs/user-guide-policy-configure-linux-bond.md#configure-bond), you can define an IP address to the bond interface, in spec.desiredState.interfaces[bond entry].ipv4.address.
Deploying the policy will end up in an IP collision, because it will create a bond on each node, with all these bonds having the same IP address.
Possible workaround until it is resolved:
1. Deploying different policy to each node using nodeSelector, with each policy setting a different IP address.
2. Manually setting the IP address of each bond after the policy is applied.

phoracek:
You are right, the example is valid only with a single node in a cluster. We should change the example to use DHCP.

PR #308, opened 2019-12-18
docs: use dhcp for cluster-wide policies

phoracek:

When showing a policy applying to multiple nodes, always use DHCP.
With static IPs, we would get collisions. Use static only with node
selectors per single node.

Resolves: #274 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/user-guide-policy-configure-linux-bond-with-vlans.md
+++ b/docs/user-guide-policy-configure-linux-bond-with-vlans.md
@@ -28,9 +28,7 @@ spec:
       type: bond
       state: up
       ipv4:
-        address:
-        - ip: 10.10.10.10
-          prefix-length: 24
+        dhcp: true
         enabled: true
       link-aggregation:
         mode: balance-rr
@@ -43,9 +41,7 @@ spec:
       type: vlan
       state: up
       ipv4:
-        address:
-        - ip: 10.102.10.10
-          prefix-length: 24
+        dhcp: true
         enabled: true
       vlan:
         base-iface: bond0
--- a/docs/user-guide-policy-configure-linux-bridge.md
+++ b/docs/user-guide-policy-configure-linux-bridge.md
@@ -28,6 +28,9 @@ spec:
         description: Linux bridge with eth1 as a port
         type: linux-bridge
         state: up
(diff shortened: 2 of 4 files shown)


docs/user-guide-policy-configure-linux-bond-with-vlans.md
```diff
@@ -27,11 +27,6 @@ spec:
     - name: bond0
       type: bond
       state: up
-      ipv4:
```
qinqon: What's ipv4 by default?  enable: false ? maybe is better to be explicit here.

docs/user-guide-policy-configure-linux-bridge.md
```diff
@@ -74,6 +77,10 @@ spec:
         description: Linux bridge with eth1 as a port
         type: linux-bridge
         state: up
+        ipv4:
+          address:
+          - ip: 10.10.10.10
```
qinqon: Use an ip address from https://tools.ietf.org/html/rfc5737

docs/user-guide-policy-configure-ovs-bridge.md
```diff
@@ -110,6 +108,10 @@ spec:
         description: Ovs bridge with eth1 as a port
         type: ovs-bridge
         state: up
+        ipv4:
+          address:
+          - ip: 10.10.10.10
```
qinqon: ditto

docs/user-guide-policy-configure-linux-bond-with-vlans.md
phoracek: It is disabled unless said otherwise https://github.com/nmstate/nmstate/blob/master/libnmstate/nm/ipv4.py#L52 (if I read it right).

I'd prefer to keep the docs small and attractive. I can make this use DHCP if you think it makes more sense. I don't have a strong opinion about this.

qinqon: We can have a chapter related to ip configuration and use at the other parts DHCP, so I am ok with using dhcp here also most of the people is going to use DHCP.

PR #323, opened 2020-01-06
report conditions with state report status

phoracek:

In case something really bad happens and we fail to obtain the current network state to update NNS, we should be able to report it. With this PR, we start reporting conditions in NNS.

This patch also piggy-backs fix of state_utils.go module name and +option
annotations.

Note that NNS attributes had to be changed into pointers. Without this change,
it was not possible to UPDATE NNS after unsuccessful reporting (the client would
complain about invalid type of attributes "null").

**Release note**:

```release-note
NNS conditions reporting success/failure of obtaining the current state of network
```

### DIFF
--- a/pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
+++ b/pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
@@ -8,7 +8,9 @@ import (
 	"time"
 
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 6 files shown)


pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -113,10 +115,20 @@ func (r *ReconcileNodeNetworkState) Reconcile(request reconcile.Request) (reconc
 			return err
 		}
 
-		err = nmstate.UpdateCurrentState(r.client, instance)
+		stateToReport, err := nmstate.CurrentState()
+		if err != nil {
+			setConditionDegraded(instance, err.Error())
```
qinqon: Don't we want to do similar stuff like we do for policies, moving this into s`statestatus` package and also do the retry on conflict there for status update (although I think there is no conflicts here).

phoracek: There should be no conflict. If we get a bug with a conflict, it is a red flag.

I want to keep this minimal, unlit the other conditions, here there is no concurency, we don't update while reconciling, we just use this helper function to set fields in the Status instance. Save it later together with updated spec. I'm happy with this whole controller being a single short module, which it is now.

phoracek:
@qinqon this should be ready now. Please take a look.

pkg/apis/nmstate/v1alpha1/zz_generated.openapi.go
```diff
@@ -11,15 +11,15 @@ import (
...
+		"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.NodeNetworkConfigurationPolicy":          schema_pkg_apis_nmstate_v1alpha1_NodeNetworkConfigurationPolicy(ref),
+		"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.NodeNetworkConfigurationPolicySpec":      schema_pkg_apis_nmstate_v1alpha1_NodeNetworkConfigurationPolicySpec(ref),
+		"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.NodeNetworkConfigurationPolicyStatus":    schema_pkg_apis_nmstate_v1alpha1_NodeNetworkConfigurationPolicyStatus(ref),
+		"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.NodeNetworkState":                        schema_pkg_apis_nmstate_v1alpha1_NodeNetworkState(ref),
+		"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.NodeNetworkStateStatus":                  schema_pkg_apis_nmstate_v1alpha1_NodeNetworkStateStatus(ref),
+		"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.State":                                   schema_pkg_apis_nmstate_v1alpha1_State(ref),
```
qinqon: this is weird, how did you run it ?

phoracek:
@qinqon it should be fine now.

PR #324, opened 2020-01-07
Add conditions reset mutating webhook

qinqon:
    To ensure that policy conditions are reset at updated
    policies, this wait a user can safely call kubectl apply
    and then kubectl wait, since previous conditions are
    cleaned up.

**Release note**:

```release-note
Now it's safe to call kubectl wait after kubectl apply on a policy.
```


build/Dockerfile
```diff
@@ -20,6 +20,10 @@ ENV OPERATOR=/usr/local/bin/kubernetes-nmstate \
 # install operator binary
 COPY build/_output/bin/kubernetes-nmstate ${OPERATOR}
 
+# Copy webhook ssl
+COPY build/_output/nmstate-webhook.pem /ssl/tls.crt
```
phoracek: Please make it so we load these from a secret. It does not need to be perfect, but we better build it with cert-manager in mind. Since you mount a secret as a file, it should remain the same for the app.

oshoval: https://github.com/morvencao/kube-mutating-webhook-tutorial/blob/master/deployment/webhook-create-signed-cert.sh 

this can help for that ?

oshoval:
nice 

would we have problems with the cert / ca rotations as the bugzilla that floating now ?

qinqon:
@oshoval this is wip final version will use cert-manager

build/Dockerfile
qinqon: I will do this as next step since, I see the hook callback working now, thanks for the pointer  @oshoval

qinqon: Ok done, @phoracek can you take a look ?

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,76 @@
...
+	return jsonpatchv2.Operation{
+		Path:      "/status/conditions",
+		Operation: "replace",
+		Value:     nmstatev1alpha1.ConditionList{},
+	}
+
+}
+
+func mutationAnnotationPatch(annotations map[string]string) jsonpatchv2.Operation {
+	key := "conditions_mutation"
```
phoracek: Please make it namespaced so it is obvious it was added by us. Also, AFAIK I'd make it lowerCamelCase. Also, if it cold be more specific, spelling that it is webhookMutationTimestamp maybe?

Makefile
```diff
@@ -8,6 +8,8 @@ HANDLER_IMAGE_SUFFIX ?=
 HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME)$(HANDLER_IMAGE_SUFFIX)
 HANDLER_IMAGE ?= $(IMAGE_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)
 
+WHAT ?= ./pkg
```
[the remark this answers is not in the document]
qinqon: The main purpose is to parameterize it, when you run test/unit you want sometimes to iterate fuster by just compiling/running unit test at package, for example for webhook I have iterate with `make WHAT=./pkg/webhook test/unit` using `-focus` re-runs all but just skip, at kubevirt they do same thing I have just copy from there.

cmd/manager/main.go
```diff
@@ -114,6 +115,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup webhook
+	if err := mgr.Add(webhook.Server()); err != nil {
```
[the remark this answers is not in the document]
qinqon: I have test it without "caBundle" and e2e webhook tests is failing as expected.

hack/install-tls-secrets.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
qinqon: I have modify it a little, the original project has apache license so we good [1]

[1] https://github.com/morvencao/kube-mutating-webhook-tutorial/blob/master/LICENSE

pkg/webhook/webhook.go
qinqon: Reading best practices from here [1] I am going to name it nmstate.io/webhook-mutation-timestamp

[1] https://www.replex.io/blog/9-best-practices-and-examples-for-working-with-kubernetes-labels

qinqon:
There is something fishy at default bridge e2e test
```golang
2020-01-09T09:19:50.563Z	DEBUG	controller_node	Reconciling Node	{"Request.Namespace": "", "Request.Name": "node01"}
2020-01-09T09:19:50.716Z	ERROR	enactmentconditions	Error notifying state Success	{"enactment": "node01.default-network", "error": "NodeNetworkConfigurationEnactment.nmstate.io \"node01.default-network\" not found"}
github.com/go-logr/zapr.(*zapLogger).Error
	kubernetes-nmstate/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions.(*EnactmentConditions).NotifySuccess
	kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/enactmentstatus/conditions/conditions.go:79
github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy.(*ReconcileNodeNetworkConfigurationPolicy).Reconcile
	kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go:209
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler

```

phoracek:
@qinqon please don't forget to add readiness probe to the webhook. It takes a while before the server is ready to serve.

qinqon:
/hold
TODO:
  - proper cert implementation
  - readiness probe

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+		return errors.Wrap(err, "failed updating CA bundle at webhook")
+	}
+	return s.webhookServer.Start(stop)
+}
+
+func (s *server) InjectFunc(f inject.Func) error {
+	return s.webhookServer.InjectFunc(f)
+}
+
+func (s *server) NeedLeaderElection() bool {
```
phoracek: What does this do? We don't need leader election

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return nil
+}
+
+func (s *server) Start(stop <-chan struct{}) error {
+	err := s.updateCABundle()
+	if err != nil {
+		return errors.Wrap(err, "failed updating CA bundle at webhook")
+	}
+	return s.webhookServer.Start(stop)
```
phoracek: Does this block?

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed retrieving mutationg webhook "+webhookName)
+	}
+	// Retrieve cluster CA bundle and encode to base 64
+	authenticationConfig := corev1.ConfigMap{}
```
phoracek: We should first check if the bundle is already there. It may have been created by another pod.

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+	return s
+}
+
+// add adds a new Webhook to mgr with r as the webhook.Server
+func add(mgr manager.Manager, s *server) error {
+	mgr.Add(s)
+	return nil
+}
+
+func (s *server) updateCABundle() error {
```
phoracek: This function (or at least a part of it) has to be run with collision avoidance in mind. Other controllers may be writting in the config.

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+func newServer(mgr manager.Manager) *server {
+	s := &server{
+		webhookServer: &webhook.Server{
+			Port:    8443,
+			CertDir: "/etc/webhook/certs/",
+		},
+		mgr: mgr,
+		log: logf.Log.WithName("webhook/mutating/server"),
+	}
+	s.webhookServer.Register("/mutating", resetConditionsHook())
```
phoracek: /mutate-policy Maybe?

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+const (
+	webhookName      = "nmstate"
+	RetrievalTimeout = 120
```
phoracek: I'd prefer not using a constant. But if we do, let's keep it with the unit (time.Second) and name it better (CARetrievalTimeout)?

pkg/webhook/mutating/server.go
```diff
@@ -0,0 +1,102 @@
...
+	mutatingWebHook.Webhooks[0].ClientConfig.CABundle = []byte(clientCaFile)
+	err = s.mgr.GetClient().Update(context.TODO(), &mutatingWebHook)
+	if err != nil {
+		return errors.Wrap(err, "failed to update mutating webhook CABundle")
+	}
+	return nil
+}
+
+func (s *server) Start(stop <-chan struct{}) error {
+	err := s.updateCABundle()
```
phoracek: Please keep in mind that we will need to handle expiration of CA. No need to handle it in this PR, just so we make it extendable.

hack/install-tls-secrets.sh
```diff
@@ -0,0 +1,141 @@
...
+detailed explantion and additional instructions.
+
+The server key/cert k8s CA cert are stored in a k8s secret.
+
+usage: ${0} [OPTIONS]
+
+The following flags are required.
+
+       --service          Service name of webhook.
+       --namespace        Namespace where webhook service and secret reside.
```
RamLavi: is order of parameter options relevant? use of function in cluster-sync-handler.sh seems different from this parameter order.

hack/install-tls-secrets.sh
```diff
@@ -0,0 +1,141 @@
...
+extendedKeyUsage = serverAuth
+subjectAltName = @alt_names
+[alt_names]
+DNS.1 = ${service}
+DNS.2 = ${service}.${namespace}
+DNS.3 = ${service}.${namespace}.svc
+EOF
+
+openssl genrsa -out ${tmpdir}/server-key.pem 2048
+openssl req -new -key ${tmpdir}/server-key.pem -subj "/CN=${service}.${namespace}.svc" -out ${tmpdir}/server.csr -config ${tmpdir}/csr.conf
```
RamLavi: If you don't want the key to be protected with password, I would suggest considering '-nodes' flag
For more information, see [link](https://stackoverflow.com/questions/5051655/what-is-the-purpose-of-the-nodes-argument-in-openssl).

pkg/webhook/mutating/server.go
qinqon: Since we are delegating on webhook.Server we have to honor https://github.com/kubernetes-sigs/controller-runtime/blob/40070e2a1958c3d974ba95da883a2bd088137789/pkg/manager/internal.go#L215

pkg/webhook/mutating/server.go
qinqon: Yep, we don't care controller-runtime, calls it for us.

pkg/webhook/mutating/server.go
qinqon: good enouogh, we are mutating conditions and labels, also do we need a full name ?
like "/api/nmstate.io-v1alpha1/mutate-policy" or the like.

pkg/webhook/mutating/server.go
qinqon: ack, we can take a look at webhook cert watcher to check if we can extend it to CA cert

hack/install-tls-secrets.sh
qinqon: Thanks for ACK, we are going to move this to service startup so we self sign stuff, so this script is going to disappear.

pkg/webhook/add_mutating.go
```diff
@@ -0,0 +1,10 @@
+package webhook
+
+import (
+	"github.com/nmstate/kubernetes-nmstate/pkg/webhook/mutating"
```
phoracek: I think the webhook should go under pkg/webhook/nodenetworkconfigurationpolicy, based on the type. It is more important than the fact we are changing it. But maybe it is just a matter of taste, up to you.

vendor/k8s.io/client-go/util/certificate/OWNERS
```diff
@@ -0,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
```
phoracek: This module is under util/, it may change drastically with newer version of client-go. Maybe we should clone it under under repo instead.

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	mutatingwebhook "github.com/nmstate/kubernetes-nmstate/pkg/webhook/mutating"
+)
+
+var _ = Describe("Mutating Admission Webhook", func() {
+	policyName := "webhook-test"
```
phoracek: Why do we have a dedicated name here, I thought we already have one used everywhere.

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"strconv"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	mutatingwebhook "github.com/nmstate/kubernetes-nmstate/pkg/webhook/mutating"
+)
+
+var _ = Describe("Mutating Admission Webhook", func() {
```
phoracek: Please add a comment explaining why we used the label instead of real testing of the set values.

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,54 @@
...
+			Expect(policy.ObjectMeta.Annotations).To(HaveKey(mutatingwebhook.TimestampLabelKey))
+		})
+		Context("and we updated it", func() {
+			var (
+				oldPolicy nmstatev1alpha1.NodeNetworkConfigurationPolicy
+			)
+			BeforeEach(func() {
+				oldPolicy = nodeNetworkConfigurationPolicy(policyName)
+				setDesiredStateWithPolicy(policyName, linuxBrAbsent(bridge1))
+				waitForAvailablePolicy(policyName)
```
phoracek: Now when this is really reliable, don't we have any sleeps or blocking in the rest of the tests we want to drop?

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,54 @@
...
+			})
+			It("should update annotation with newer mutation timestamp", func() {
+				newPolicy := nodeNetworkConfigurationPolicy(policyName)
+				Expect(newPolicy.ObjectMeta.Annotations).To(HaveKey(mutatingwebhook.TimestampLabelKey))
+
+				annotation := oldPolicy.ObjectMeta.Annotations[mutatingwebhook.TimestampLabelKey]
```
phoracek: Please name it oldAnnotation and newAnnotation. Reusing variable names is usually confusing.

pkg/webhook/tls/tls.go
```diff
@@ -0,0 +1,95 @@
+package tls
```
phoracek: Is this copied from somewhere or did you make it?

If just copied, please reference the source. Otherwise, please provide a good docs to this module. It may be reused and also, TLS is confusing.

pkg/webhook/tls/tls.go
```diff
@@ -0,0 +1,95 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+type webhookTLS struct {
+	mgr manager.Manager
+	log logr.Logger
+}
+
+func New(mgr manager.Manager) webhookTLS {
```
phoracek: This is confusing to me. Why is the `type` private and this method public? Due to the private `type`, this cannot be used outside the module anyway, right?

pkg/webhook/tls/store.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type filePairStore struct {
+	directory string
+	certFile  string
+	keyFile   string
+	fileStore certificate.FileStore
+}
+
+func NewFilePairStore(directory string, certFile string, keyFile string) (certificate.FileStore, error) {
```
phoracek: See my comments on tls.go. Both for the naming and docs.

hack/install-tls-secrets.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
+
+# Original script is here [1] we have just modify it to use ${KUBECTL} env var
```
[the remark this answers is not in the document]
qinqon: Removed

vendor/k8s.io/client-go/util/certificate/OWNERS
qinqon: Let's not cloning, if it changes we adapt, we can always go back to clone old version if needed.

test/e2e/webhook_test.go
qinqon: This test check that we catch CREATE and UPDATE policy events at webhook, so we have to be sure that policy does not exist that's why I create a new one, but we can ensure that TestPolicy is deleted at BeforeEach and use it.

test/e2e/webhook_test.go
qinqon: What do you mean ?

pkg/webhook/tls/tls.go
qinqon: Is new stuff I did, will add some docs, we may want to also put it ouside of knmstate in case is resued by kubemacpool.

pkg/webhook/tls/tls.go
qinqon: this way webhookTLS can only be instanciate with "New" function, so but people can still call it's public methods.

test/e2e/webhook_test.go
phoracek: I think that using the same name and ensuring that it is not present is more expressive.

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+		patch := client.ConstantPatch(types.JSONPatchType, obtainedResponse.Patch)
+		policy := c.Policy.DeepCopy()
+		err = cli.Patch(context.TODO(), policy, patch)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Retrieve the patched policy")
+		err = cli.Get(context.TODO(), types.NamespacedName{Name: ""}, &patchedPolicy)
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(obtainedResponse.Allowed).To(BeTrue(), "Mutation not allowed")
```
phoracek: This whole test is quite a wild west. It would help if there was something I could read to get some hints about the code. Could we change the explanation string to say "Mutation of the request should be allowed". Similar in the following checks "Status reason should match the expected". "There should be exactly two patches in the response"

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Expect(obtainedResponse.Result.Reason).To(Equal(metav1.StatusReason("Conditions reset")))
+
+		Expect(obtainedResponse.Patches).To(HaveLen(2), "Unexpected number of patches")
+
+		for _, patch := range obtainedResponse.Patches {
+			_, err := patch.MarshalJSON()
+			Expect(err).ToNot(HaveOccurred(), "Invalid patch")
```
phoracek: "The patches should contain valid JSON"

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		Expect(patchedPolicy.Status.Conditions).To(BeEmpty(), "Condition list not resetted")
```
phoracek: "The list of conditions should be reset" (but Petr would prefer if they were set to Unknown)

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+		if c.Policy.ObjectMeta.Annotations != nil {
+			Expect(patchedPolicy.ObjectMeta.Annotations).To(HaveLen(len(c.Policy.ObjectMeta.Annotations) + 1))
```
phoracek: Cool BDD!

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Expect(obtainedResponse.Patches).To(HaveLen(2), "Unexpected number of patches")
+
+		for _, patch := range obtainedResponse.Patches {
+			_, err := patch.MarshalJSON()
+			Expect(err).ToNot(HaveOccurred(), "Invalid patch")
+		}
+
+		Expect(patchedPolicy.Status.Conditions).To(BeEmpty(), "Condition list not resetted")
+
+		if c.Policy.ObjectMeta.Annotations != nil {
```
phoracek: This should never happen though. Shouldn't we check for it not to be nil via Expect?

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+			Expect(patchedPolicy.ObjectMeta.Annotations).To(HaveLen(len(c.Policy.ObjectMeta.Annotations) + 1))
+		}
+
+		Expect(patchedPolicy.ObjectMeta.Annotations).To(HaveKey(TimestampLabelKey))
+		annotation := patchedPolicy.ObjectMeta.Annotations[TimestampLabelKey]
+		mutationTimestamp, err := strconv.ParseInt(annotation, 10, 64)
+		Expect(err).ToNot(HaveOccurred(), "mutation timestamp has not a int64 value")
+		Expect(mutationTimestamp).To(BeNumerically(">", testStartTime), "mutation timestamp not updated")
```
phoracek: "mutation timestamp should be updated by the webhook"

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Expect(patchedPolicy.Status.Conditions).To(BeEmpty(), "Condition list not resetted")
+
+		if c.Policy.ObjectMeta.Annotations != nil {
+			Expect(patchedPolicy.ObjectMeta.Annotations).To(HaveLen(len(c.Policy.ObjectMeta.Annotations) + 1))
+		}
+
+		Expect(patchedPolicy.ObjectMeta.Annotations).To(HaveKey(TimestampLabelKey))
+		annotation := patchedPolicy.ObjectMeta.Annotations[TimestampLabelKey]
+		mutationTimestamp, err := strconv.ParseInt(annotation, 10, 64)
+		Expect(err).ToNot(HaveOccurred(), "mutation timestamp has not a int64 value")
```
phoracek: "mutation timestamp should have int64 value"

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+		ClientFn: func(current *tls.Certificate) (certificatesclientv1beta1.CertificateSigningRequestInterface, error) {
+			certClient, err := certificatesclientv1beta1.NewForConfig(crMgr.GetConfig())
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create cert client for webhook")
+			}
+			return newCSRApprover(certClient.CertificateSigningRequests()), nil
+		},
+		Template: &x509.CertificateRequest{
+			Subject: pkix.Name{
+				CommonName: fmt.Sprintf("%s", webhookName),
```
phoracek: webhookName is already string, you don't need to format it

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	certDir string, certFileName string, keyFileName string) (*manager, error) {
+
+	certStore, err := NewFilePairStore(
+		certDir,
+		certFileName,
+		keyFileName,
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to initialize new webhook cert/key file store")
+	}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	} else if webhookType == ValidatingWebhook {
+		validatingWebHookConfig, err := m.updateValidatingWebhookCABundle(webhookName)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to update CA bundle at webhook")
+		}
+
+		for _, webhook := range validatingWebHookConfig.Webhooks {
+			dnsNames = append(dnsNames, dnsNamesForService(*webhook.ClientConfig.Service)...)
+		}
+	}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+			},
+			DNSNames: dnsNames,
+		},
+		Usages: []certificatesv1beta1.KeyUsage{
+			certificatesv1beta1.UsageDigitalSignature,
+			certificatesv1beta1.UsageKeyEncipherment,
+			certificatesv1beta1.UsageServerAuth,
+		},
+		CertificateStore: certStore,
+	}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	certManager, err := certificate.NewManager(&certConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed initializing webhook cert manager")
+	}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	m.certManager = certManager
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+		keyExists, err := m.certStore.keyFileExists()
+		if err != nil {
+			return false, err
+		}
+		certExists, err := m.certStore.keyFileExists()
+		if err != nil {
+			return false, err
+		}
+		return keyExists && certExists, nil
+	})
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed creating webhook tls key/cert")
+	}
+	m.log.Info(fmt.Sprintf("TLS cert/key ready at %s", m.certStore.CurrentPath()))
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/manager.go
```diff
@@ -0,0 +1,166 @@
...
+	certificate, err := m.certStore.Current()
+	if err != nil {
+		return errors.Wrap(err, "failed retrieving webhook current certificate")
+	}
+	m.log.Info(fmt.Sprintf("Certificate expiration is %v-%v", certificate.Leaf.NotBefore, certificate.Leaf.NotAfter))
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/ca.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Retrieve cluster CA bundle and encode to base 64
+func (m manager) clientCAFile() ([]byte, error) {
+	authenticationConfig := corev1.ConfigMap{}
+	err := m.crMgr.GetClient().Get(context.TODO(), types.NamespacedName{Namespace: "kube-system", Name: "extension-apiserver-authentication"}, &authenticationConfig)
+
+	if err != nil {
+		return []byte{}, errors.Wrap(err, "failed to retrieve cluster authentication config")
+	}
+	clientCaFile := authenticationConfig.Data["client-ca-file"]
```
phoracek: Please try at least manually whether this works on OpenShift too. There are people looking in knmstate there and we should make sure we don't have some hard blocker there (I expect bugs to be there due to absent CI, but at least this is an obvious potential gap)

pkg/webhook/server/certificate/ca.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// Retrieve cluster CA bundle and encode to base 64
+func (m manager) clientCAFile() ([]byte, error) {
+	authenticationConfig := corev1.ConfigMap{}
+	err := m.crMgr.GetClient().Get(context.TODO(), types.NamespacedName{Namespace: "kube-system", Name: "extension-apiserver-authentication"}, &authenticationConfig)
+
+	if err != nil {
+		return []byte{}, errors.Wrap(err, "failed to retrieve cluster authentication config")
+	}
```
phoracek: please insert an empty line after this one

cmd/manager/main.go
```diff
@@ -48,6 +50,10 @@ func printVersion() {
 func main() {
 	var logType string
 
+	// Print V(2) logs from packages using klog
+	klog.InitFlags(nil)
```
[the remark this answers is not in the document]
qinqon: It's needed to have logs from pakcages that uses klog, in this case [1]

[1] https://github.com/kubernetes/client-go/blob/master/util/certificate/certificate_manager.go#L404

pkg/webhook/server/certificate/csr.go
```diff
@@ -0,0 +1,70 @@
...
+	return a.approve(csr)
+}
+
+func (a approver) Update(csr *v1beta1.CertificateSigningRequest) (*v1beta1.CertificateSigningRequest, error) {
+	return a.client.Update(csr)
+}
+
+func (a approver) UpdateStatus(csr *v1beta1.CertificateSigningRequest) (*v1beta1.CertificateSigningRequest, error) {
+	return a.client.UpdateStatus(csr)
+}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/csr.go
```diff
@@ -0,0 +1,70 @@
...
+func (a approver) Delete(name string, options *v1.DeleteOptions) error {
+	return a.client.Delete(name, options)
+}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/csr.go
```diff
@@ -0,0 +1,70 @@
...
+func (a approver) DeleteCollection(options *v1.DeleteOptions, listOptions v1.ListOptions) error {
+	return a.client.DeleteCollection(options, listOptions)
+}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/csr.go
```diff
@@ -0,0 +1,70 @@
...
+func (a approver) Get(name string, options v1.GetOptions) (*v1beta1.CertificateSigningRequest, error) {
+	csr, err := a.client.Get(name, options)
+	if err != nil {
+		return csr, err
+	}
+	return a.approve(csr)
+}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/csr.go
```diff
@@ -0,0 +1,70 @@
...
+func (a approver) List(opts v1.ListOptions) (*v1beta1.CertificateSigningRequestList, error) {
+	return a.client.List(opts)
+}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/csr.go
```diff
@@ -0,0 +1,70 @@
...
+func (a approver) Watch(opts v1.ListOptions) (watch.Interface, error) {
+	return a.client.Watch(opts)
+}
```
phoracek: please insert an empty line after this one

pkg/webhook/server/certificate/ca.go
```diff
@@ -0,0 +1,123 @@
...
+
+func mutatingWebhookConfig(webhook runtime.Object) *admissionregistrationv1beta1.MutatingWebhookConfiguration {
+	return webhook.(*admissionregistrationv1beta1.MutatingWebhookConfiguration)
+}
+
+func validatingWebhookConfig(webhook runtime.Object) *admissionregistrationv1beta1.ValidatingWebhookConfiguration {
+	return webhook.(*admissionregistrationv1beta1.ValidatingWebhookConfiguration)
+}
+
+func clientConfigList(webhook runtime.Object, webhookType WebhookType) []*admissionregistrationv1beta1.WebhookClientConfig {
```
phoracek: Could you please add a comment explaining what is clientConfig?

pkg/webhook/server/certificate/ca.go
```diff
@@ -0,0 +1,123 @@
...
+			err := m.crMgr.GetClient().Get(context.TODO(), webhookKey, webhook)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return false, nil
+				}
+				return false, err
+			}
+			return true, nil
+		})
+
```
phoracek: drop the line

pkg/webhook/server/server.go
```diff
@@ -0,0 +1,90 @@
...
+type ServerModifier func(w *webhook.Server)
+
+// Add creates a new Conditions Mutating Webhook and adds it to the Manager. The Manager will set fields on the Webhook
+// and Start it when the Manager is Started.
+func New(mgr manager.Manager, webhookName string, webhookType certificate.WebhookType, serverOpts ...ServerModifier) *server {
+	s := &server{
+		webhookName: webhookName,
+		webhookType: webhookType,
+		webhookServer: &webhook.Server{
+			Port:    8443,
```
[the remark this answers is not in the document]
qinqon: You can server multiple hooks for same server, at this implementation is a server per Mutating/ValidationWebhookConfiguration (It contains multiple webhooks)

pkg/webhook/server/server.go
```diff
@@ -0,0 +1,90 @@
...
+func WithCertDir(certDir string) ServerModifier {
+	return func(s *webhook.Server) {
+		s.CertDir = certDir
+	}
+}
+
+func (s *server) Start(stop <-chan struct{}) error {
+	s.log.Info("Starting nodenetworkconfigurationpolicy webhook server")
+
+	// We have only one webhook so we just take the first one
```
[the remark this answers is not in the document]
qinqon: not needed we update bundleCA at all the webhooks at the webhookConfig, this is old comment

pkg/webhook/server/server.go
```diff
@@ -0,0 +1,90 @@
...
+	defer certManager.Stop()
+
+	err = s.webhookServer.Start(stop)
+	if err != nil {
+		return errors.Wrap(err, "failed starting webhook server")
+	}
+	return nil
+}
+
+func (s *server) InjectFunc(f inject.Func) error {
```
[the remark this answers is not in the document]
qinqon: We have to implement SetFields.Injector to be able to running as another controller at controller-runtime manager.

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
qinqon: Well this is unit test I test the function and there it can happend.

pkg/webhook/server/certificate/manager.go
qinqon: It was something else before, now is just the hook name.

pkg/webhook/server/certificate/ca.go
qinqon: I did run webhook e2e tests at ocp-4.3 provider and it was passing, meaning that we have proper webhook setup I suppose this is ok then.

pkg/webhook/nodenetworkconfigurationpolicy/conditions_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+var _ = Describe("Mutating Admission Webhook", func() {
+	type WebhookCase struct {
```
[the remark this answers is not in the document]
qinqon: What do you mean ?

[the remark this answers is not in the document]
phoracek: If there is only a single attribute here, maybe we could change the signature to:

`DescibeTable("when reset condition is called on a Policy", func(p ...Policy) {`

I won't fight for it. Just that it may be more clear on the first sight.

pkg/webhook/server/server.go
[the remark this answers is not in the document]
phoracek: I still don't get why you set default `Port` and `CertDir` instead of using `WithPort` and `WithCertDir`.

phoracek:
/approve

Looks good, but we gotta fix tests.

pkg/apis/nmstate/v1alpha1/condition_types.go
```diff
@@ -38,14 +38,14 @@ func NewCondition(conditionType ConditionType, status corev1.ConditionStatus, re
 	return condition
 }
 
-func (conditions *ConditionList) Set(conditionType ConditionType, status corev1.ConditionStatus, reason ConditionReason, message string) {
+func (conditions *ConditionList) Set(conditionType ConditionType, status corev1.ConditionStatus, reason ConditionReason, message string) *ConditionList {
```
phoracek: I feel that returning the list makes the signature odd. The user then needs to ask whether the Set is mutating or not. I see you changed it to make table testing easier, but please, introduce a helper function for tests instead.

pkg/apis/nmstate/v1alpha1/condition_types_test.go
```diff
@@ -103,6 +103,28 @@ var _ = Describe("Conditions list", func() {
...
+			It("should change values and update LastTransitionTime and LastHearbeatTime", func() {
+				updatedCondition := newConditions.Find(originalConditions[0].Type)
+				Expect(updatedCondition.Status).To(Equal(updatedConditionStatus))
+				Expect(updatedCondition.Reason).To(Equal(updatedConditionReason))
+				Expect(updatedCondition.Message).To(Equal(updatedConditionMessage))
+				Expect(updatedCondition.LastTransitionTime.Time).To(BeTemporally(">", originalConditions[0].LastTransitionTime.Time))
+				Expect(updatedCondition.LastHeartbeatTime.Time).To(BeTemporally(">", originalConditions[0].LastHeartbeatTime.Time))
+			})
+		})
+		Context("and we reset it with unknown status and empty reason and message", func() {
```
phoracek: I don't see how is this Context needed. The `Set` does not care what you set, so the values in the previous context are as good as those from this Context. Am I wrong?

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
```diff
@@ -60,6 +60,7 @@ var NodeNetworkConfigurationPolicyConditionTypes = [...]ConditionType{
 }
 
 const (
+	NodeNetworkConfigurationPolicyConditionUnknownState                ConditionReason = "UnknownState"
```
phoracek: The value should be a reason (`ConditionReason`). So in this case, maybe `"DesiredStateHasChanged"`?

pkg/webhook/nodenetworkconfigurationpolicy/server.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/nmstate/kubernetes-nmstate/pkg/webhook/server/certificate"
+)
+
+const (
+	webhookName = "nmstate"
+)
+
+func Add(mgr manager.Manager) error {
+	server := webhookserver.New(mgr, webhookName, certificate.MutatingWebhook,
+		webhookserver.WithHook("/nodenetworkconfigurationpolicies-mutate", deleteConditionsHook()),
```
phoracek: I think this deserves some comments. Could you please describe why do we need two, one `delete` and one `set`? What is executed when, what do they do.

deploy/role.yaml
```diff
@@ -65,3 +78,9 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
phoracek: Why do we need this? Why is it introduced now?

pkg/webhook/server/server.go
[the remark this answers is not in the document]
phoracek: @qinqon ^^

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+	corev1 "k8s.io/api/core/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+)
+
+const (
+	TimestampLabelKey = "nmstate.io/webhook-mutating-timestamp"
```
phoracek: Could you move this to a separate module and under a separate hook? If we introduce a new hook (other than those for conditions.go), this label would be universal.

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+	return policy
+}
+
+func mutatePolicy(policy nmstatev1alpha1.NodeNetworkConfigurationPolicy, mutators ...mutator) nmstatev1alpha1.NodeNetworkConfigurationPolicy {
+	for _, mutator := range mutators {
+		policy = mutator(policy)
+	}
+	return policy
+}
+func mutatePolicyHandler(neededMutationFor func(nmstatev1alpha1.NodeNetworkConfigurationPolicy) bool, mutators ...mutator) admission.HandlerFunc {
```
phoracek: Empty line above

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+func deleteConditionsHook() *webhook.Admission {
+	return &webhook.Admission{
+		Handler: admission.HandlerFunc(
+			mutatePolicyHandler(
+				func(nmstatev1alpha1.NodeNetworkConfigurationPolicy) bool {
+					// Always delete conditions when nncp is changed
+					return true
+				},
+				deleteConditions,
+				resetTimestampAnnotation)),
```
phoracek: Please make those `))` so they are the same way as in the following function

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func setConditionsUnknownHook() *webhook.Admission {
+	return &webhook.Admission{
+		Handler: admission.HandlerFunc(
+			mutatePolicyHandler(
+				func(policy nmstatev1alpha1.NodeNetworkConfigurationPolicy) bool {
```
phoracek: So you expect the the hook above to be always executed first? Is there an ordering?

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
phoracek: I would use a separate hook for the timestamp anotation. Then you can make this helper Handler function simpler.

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+}
+func mutatePolicyHandler(neededMutationFor func(nmstatev1alpha1.NodeNetworkConfigurationPolicy) bool, mutators ...mutator) admission.HandlerFunc {
+	return func(ctx context.Context, req webhook.AdmissionRequest) webhook.AdmissionResponse {
+		original := req.Object.Raw
+		policy := nmstatev1alpha1.NodeNetworkConfigurationPolicy{}
+		err := json.Unmarshal(original, &policy)
+		if err != nil {
+			return admission.Errored(http.StatusInternalServerError, errors.Wrapf(err, "failed decoding policy: %s", string(original)))
+		}
+		if !neededMutationFor(policy) {
```
phoracek: empty line above

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+			return admission.Allowed("mutation not needed")
+		}
+		policy = mutatePolicy(policy, mutators...)
```
phoracek: empty line above

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
```diff
@@ -0,0 +1,107 @@
...
+		current, err := json.Marshal(policy)
+		if err != nil {
+			return admission.Errored(http.StatusInternalServerError, errors.Wrapf(err, "failed encoding policy: %+v", policy))
+		}
+		response := admission.PatchResponseFromRaw(original, current)
```
phoracek: empty line above

pkg/apis/nmstate/v1alpha1/condition_types_test.go
qinqon: Removing it, was introduced to check that it was working fine, but is not needed.

pkg/apis/nmstate/v1alpha1/nodenetworkconfigurationpolicy_types.go
qinqon: Not needed, also it could be that desiredState is not changed but we re-apply it.

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
qinqon: Well maybe we want two timestamps, one for deleteConditions and another for setConditionsUnknown.

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
qinqon: we don't we just one to have the timestamp.

pkg/webhook/nodenetworkconfigurationpolicy/conditions.go
qinqon: maybe is even better to use different annotation one for nncp and other nncp/status

pkg/webhook/server/server.go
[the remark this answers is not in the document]
qinqon: They will be override by serverOpts if needed, those are a good defaults.

phoracek:
@qinqon just gotta rebase it (I pushed the rebase button but it fails on DCO)

PR #328, opened 2020-01-09
Retry default gw retrieval on error

qinqon:

Sometimes it takes some time for DHCP to assign default gw, so
the default gw ping probe fails, here we add a 30 seconds timeout
with a 1 second interval.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -35,6 +35,7 @@ var (
 
 const nmstateCommand = "nmstatectl"
 const vlanFilteringCommand = "vlan-filtering"
+const defaultGwRetrieveTimeout = 120
 const defaultGwProbeTimeout = 120
 const apiServerProbeTimeout = 120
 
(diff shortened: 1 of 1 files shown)


pkg/helper/client.go
```diff
@@ -35,6 +35,7 @@ var (
 
 const nmstateCommand = "nmstatectl"
 const vlanFilteringCommand = "vlan-filtering"
+const defaultGwRetrieveTimeout = 30
```
phoracek: I can imagine the GW reachability in the next constant may be limited by this one. Let's make all of them 120 for now and later we can pass Context around with shared timeout. What do you think?

phoracek:
/approve
/lgtm
/hold

Let's wait for the CI.

phoracek:
We ping `""` now

pkg/helper/client.go
qinqon: 120 is good for me.

phoracek:
@qinqon please rebase

qinqon:
@phoracek let's wait for CI I will rebase

qinqon:
/hold cancel

ci is good.

PR #337, opened 2020-01-14
add OCP 4.3 CI lane

phoracek:
The OCP provider seems to be more stable in a moment, so it may allow us to unblock the CI for OpenShift.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/install-kubevirtci.sh
+++ b/hack/install-kubevirtci.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 organization=kubevirt
-commit="6132f2efcf2aa2d48f256a409fe90cc0b74c3563"
+commit="78dcc4f8aaec467bba8ae325646056128268a169"
 
 script_dir=$(dirname "$(readlink -f "$0")")
 kubevirtci_dir=kubevirtci
--- a/automation/check-patch.e2e-k8s.mounts
+++ b/automation/check-patch.e2e-k8s.mounts
@@ -1,1 +1,0 @@
-check-patch.e2e.mounts
--- a/automation/check-patch.e2e-k8s.packages
+++ b/automation/check-patch.e2e-k8s.packages
@@ -1,1 +1,0 @@
-check-patch.e2e.packages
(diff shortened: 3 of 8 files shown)


automation/check-patch.e2e-okd.sh
```diff
@@ -7,15 +7,12 @@
...
-# FIXME: Delete this when okd lane works
-exit 0
-
 teardown() {
     make cluster-down
 }
 
 main() {
-    export KUBEVIRT_PROVIDER='okd-4.1'
+    export KUBEVIRT_PROVIDER='ocp-4.3'
```
oshoval: hmm dont we need quay login here ?

hack/install-kubevirtci.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 organization=kubevirt
-commit="6132f2efcf2aa2d48f256a409fe90cc0b74c3563"
+commit="78dcc4f8aaec467bba8ae325646056128268a169"
```
oshoval: do we want to take bump-kubevirtci from kubevirt to k8s-nmstate ?

phoracek: It does not make that much sense here. We don't keep kubevirtci in the source tree, all needed here is to change this single hash. And I'd rather do it manually than keep a script for one-line change.

automation/check-patch.e2e-okd.sh
phoracek: You are right, will send a PR in a moment.

automation/check-patch.e2e-okd.sh
```diff
@@ -23,6 +23,9 @@ main() {
     # Let's fail fast if it's not compiling
     make handler
 
+    # The OCP provider requires jq
+    sudo yum install -y jq
```
oshoval: dnf or yum ? or both available i guess ?

oshoval: also do we have docker-ce latest build ? for big images
(i guess, lets see , or add docker --version print so we know if stuff happens)

oshoval:
LGTM

automation/check-patch.e2e-okd.sh
phoracek: actually, let me add it to the job. i don't want to install deps here. this script should just run tests on any environment (with deps in place)

phoracek:
/override pull-kubernetes-nmstate-unit-test

The CI issue is not related.

PR #340, opened 2020-01-16
Replace e2e test for nns cleanup with unit test

qinqon:
The e2e test for nns was removing nodes that's tricky operation
    for the cluster, let's just test this as unit test so we are good
    and we have an stable e2e test regression without skip/focus tricks.

**Release note**:

```release-note
NONE
```


automation/check-patch.setup.sh
```diff
@@ -18,6 +18,6 @@ gimme_dir=$tmp_dir/go/gimme
 mkdir -p $gimme_dir
 curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=${gimme_dir} bash >> ${gimme_dir}/gimme.sh
 source $gimme_dir/gimme.sh
-
+export E2E_TEST_ARGS='-ginkgo.noColor'
```
phoracek: Please keep this in the original place. We should keep our crazy Makefile parameters as close to the Makefile as possible, not to spread them everywhere.

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+		objs := []runtime.Object{&node, &nodenetworkstate}
+
+		// Create a fake client to mock API calls.
+		cl = fake.NewFakeClientWithScheme(s, objs...)
+
+		reconciler.client = cl
+	})
+	Context("when node is not found", func() {
+		BeforeEach(func() {
+			request.Name = "not-present-node"
```
phoracek: Would you please initialize the whole request in this BeforeEach? So it is clear what is a part of it and that we are not reusing previously popoulated object.

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+		})
+		It("should returns and don't queue", func() {
```
phoracek: should return empty Result (not requeue)

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+			result, err := reconciler.Reconcile(request)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result).To(Equal(reconcile.Result{}))
+		})
+	})
+	Context("when node is found", func() {
```
phoracek: when a node is found

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+		BeforeEach(func() {
+			request.Name = nodeName
```
phoracek: Again, please initialize the request here. Also, could we rename nodeName to existingNode?

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+		})
+		Context("and nodenetworkstate is there too", func() {
+			It("should returns and re-queue with refresh time", func() {
```
phoracek: should return a Result with RequeueAfter set (trigger re-reconciliation)

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+				result, err := reconciler.Reconcile(request)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(result).To(Equal(reconcile.Result{RequeueAfter: nodeRefresh}))
+			})
+		})
+		Context("and nodenetworkstate is not there", func() {
```
phoracek: and nodenetworkstate is not yet created

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+			BeforeEach(func() {
+				By("Delete the nodenetworkstate")
+				err := cl.Delete(context.TODO(), &nodenetworkstate)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			It("should create a new nodenetworkstate with node as owner ref", func() {
```
phoracek: as owner reference, making sure the nodenetworkstate will be removed when the node is deleted

(I want to make sure to record it, so we don't remove the test and unknowingly break the cleanup)

pkg/controller/node/node_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+				nnsKey := types.NamespacedName{Name: nodeName}
+				err = cl.Get(context.TODO(), types.NamespacedName{Name: nodeName}, &obtainedNNS)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(obtainedNNS.Name).To(Equal(nnsKey.Name))
+				Expect(obtainedNNS.ObjectMeta.OwnerReferences).To(HaveLen(1))
+				Expect(obtainedNNS.ObjectMeta.OwnerReferences[0]).To(Equal(
+					metav1.OwnerReference{Name: nodeName, Kind: "Node", APIVersion: "v1", UID: node.UID},
+				))
+			})
+			It("should returns and re-queue with refresh time", func() {
```
phoracek: should return a Result with RequeueAfter set (trigger re-reconciliation)

pkg/controller/node/node_controller_test.go
qinqon: There is nothing to initialize there.

phoracek: You are setting stuff in the shared global variable.l, that's not really transparent. I was asking to do request = Request{Name: ...} here

qinqon:
/retest

Transitory error 
```
Total download size: 32 k
Installed size: 40 k
Downloading packages:
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : yum-plugin-copr-1.1.31-52.el7.noarch                         1/1 
  Verifying  : yum-plugin-copr-1.1.31-52.el7.noarch                         1/1 

Installed:
  yum-plugin-copr.noarch 0:1.1.31-52.el7                                        

Complete!
2020/01/16 15:32:21 Waiting for host: 192.168.66.101:22
2020/01/16 15:32:21 Connected to tcp://192.168.66.101:22
Loaded plugins: copr, fastestmirror
Error: [Errno 14] HTTPS Error 302 - Found
Makefile:114: recipe for target 'cluster-up' failed
make: *** [cluster-up] Error 1
+ EXIT_VALUE=2

```

PR #342, opened 2020-01-17
WIP Add script to collect debug information

phoracek:
It is not much, but better than nothing for collecting debug info.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/collect.sh
+++ b/hack/collect.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/bash -x
+#
+# Collect the kubernetes-nmstate information from running cluster.
+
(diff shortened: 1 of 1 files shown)


phoracek:
/hold

I want to add more docs and try it out.

qinqon:
I am adapting the kubevirt k8s reporter, to get stuff, let's have it next week.

phoracek:
But that would be only for tests, right? This script is meant for users to be executed on their clusters before reporting an issue.

qinqon:
It will retrieve stuff between test so you can do per test forensics.

But you right, we need both.

PR #352, opened 2020-01-27
Pin and install go from go.mod version

qinqon:
This include a super low tech mechanism not docker multistage either
dockerized golang, it just download a golang tarball and install it under
repo.

There are missaligns between goalang dep from project and users environments, this way we always use the golang version needed by project.

We have skip multistage since we are based our build on Makefiles, so for example for make vet, gofmt etc we would need to include those as the multistage, and call them somehow from make running partial stage, let's just keep it simple, it may download tarball more than once, if we see it's brings more pain than help we can change to multistage.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,6 @@ export KUBEVIRT_PROVIDER ?= k8s-1.15.1
 export KUBEVIRT_NUM_NODES ?= 1
 export KUBEVIRT_NUM_SECONDARY_NICS ?= 2
 
-export GOFLAGS=-mod=vendor
-export GO111MODULE=on
 
 e2e_test_args = -singleNamespace=true -test.v -test.timeout=40m -ginkgo.v -ginkgo.slowSpecThreshold=60 $(E2E_TEST_ARGS)
 ifeq ($(findstring k8s,$(KUBEVIRT_PROVIDER)),k8s)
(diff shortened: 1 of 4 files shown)


eranco74:
LGTM, just note that one might experience changes in the behavior of gofmt [see here](https://github.com/golang/go/issues/26228).

phoracek:
/approve
/lgtm
/hold

Crazy. Holding until we resolve failing CI.

qinqon:
/retest

flaky ovs test

```
2020-01-28T09:46:52.243Z	INFO	enactmentstatus	status: {DesiredState:interfaces:
- bridge:
    options:
      stp: false
    port:
    - name: eth1
    - name: eth2
  name: br6
  state: up
  type: ovs-bridge
 Conditions:[{Type:Failing Status:True Reason:FailedToConfigure Message:error reconciling NodeNetworkConfigurationPolicy at desired state apply: cmd output: 'PING 192.168.66.2 (192.168.66.2) 56(84) bytes of data.

--- 192.168.66.2 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms

', rollback cause: error pinging external address after network reconfiguration -> error: timed out waiting for the condition, currentState: ---
dns-resolver:
  config:
    search: []
    server: []
  running:
    search: []
    server:
    - 192.168.66.2

```

PR #360, opened 2020-01-29
e2e tests baseline validation

eranco74:
First commit is validating that each test return the cluster with the same interface state as it started
Next commits remove some abounded configurations  

```release-note
NUNE
```

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -24,16 +24,18 @@ import (
 )
 
 var (
-	f                  = framework.Global
-	t                  *testing.T
-	namespace          string
-	nodes              []string
-	startTime          time.Time
-	bond1              string
-	bridge1            string
-	primaryNic         = flag.String("primaryNic", "eth0", "Primary network interface name")
-	firstSecondaryNic  = flag.String("firstSecondaryNic", "eth1", "First secondary network interface name")
-	secondSecondaryNic = flag.String("secondSecondaryNic", "eth2", "Second secondary network interface name")
+	f                    = framework.Global
+	t                    *testing.T
+	namespace            string
+	nodes                []string
+	startTime            time.Time
+	bond1                string
+	bridge1              string
+	primaryNic           = flag.String("primaryNic", "eth0", "Primary network interface name")
+	firstSecondaryNic    = flag.String("firstSecondaryNic", "eth1", "First secondary network interface name")
+	secondSecondaryNic   = flag.String("secondSecondaryNic", "eth2", "Second secondary network interface name")
+	nodesInterfacesState = make(map[string][]byte)
+	interfacesToIgnore   = []string{"flannel.1", "dummy0"}
 )
 
 var _ = BeforeSuite(func() {
@@ -46,7 +48,6 @@ var _ = BeforeSuite(func() {
 	nodeList := corev1.NodeList{}
 	err = framework.Global.Client.List(context.TODO(), &nodeList, &dynclient.ListOptions{})
 	Expect(err).ToNot(HaveOccurred())
-
 	for _, node := range nodeList.Items {
 		nodes = append(nodes, node.Name)
 	}
@@ -69,10 +70,21 @@ var _ = BeforeEach(func() {
 	bridge1 = nextBridge()
 	_, namespace = prepare(t)
 	startTime = time.Now()
+	By("Getting nodes initial state")
+	for _, node := range nodes {
+		nodeState := nodeInterfacesState(node, interfacesToIgnore)
+		nodesInterfacesState[node] = nodeState
+	}
+
 })
 
 var _ = AfterEach(func() {
-	writePodsLogs(namespace, startTime, GinkgoWriter)
+	By("Verifying initial state")
+	for _, node := range nodes {
+		nodeState := nodeInterfacesState(node, interfacesToIgnore)
+		Expect(nodesInterfacesState[node]).Should(MatchJSON(nodeState))
+		writePodsLogs(namespace, startTime, GinkgoWriter)
+	}
 })
 
 func getMaxFailsFromEnv() int {
--- a/test/e2e/nnce_conditions_test.go
+++ b/test/e2e/nnce_conditions_test.go
@@ -105,6 +105,9 @@ var _ = Describe("EnactmentCondition", func() {
 		})
 
 		AfterEach(func() {
+			By("Remove the bridge")
+			updateDesiredState(linuxBrAbsent(bridge1))
+			waitForAvailableTestPolicy()
 			By("Reset desired state at all nodes")
 			resetDesiredStateForNodes()
 		})
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -339,6 +339,12 @@ func deleteConnectionAtNodes(name string) []error {
 	return errs
 }
 
+func deleteDeviceAtNode(node string, name string) error {
+	By(fmt.Sprintf("Delete device %s  at node %s", name, node))
+	_, err := runAtNode(node , "sudo", "nmcli", "device", "delete", name)
+	return err
+}
+
 func interfaces(state nmstatev1alpha1.State) []interface{} {
 	var stateUnstructured map[string]interface{}
 	err := yaml.Unmarshal(state.Raw, &stateUnstructured)
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/nmstate/kubernetes-nmstate/build/_output/bin/go/src/encoding/json"
 	"io"
 	"io/ioutil"
 	"os/exec"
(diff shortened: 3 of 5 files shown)


test/e2e/main_test.go
```diff
@@ -69,10 +69,23 @@ var _ = BeforeEach(func() {
 	bridge1 = nextBridge()
 	_, namespace = prepare(t)
 	startTime = time.Now()
+	println("Getting nodes initial state")
```
phoracek: Please use `By` from the testing framework. The same for all the other `println`

test/e2e/utils.go
```diff
@@ -523,6 +524,24 @@ func dhcpFlag(node string, name string) bool {
 	return gjson.ParseBytes(currentStateJSON(node)).Get(path).Bool()
 }
 
+func nodeInterfacesState(node string) []byte {
```
phoracek: Could you please add a docstring describing the returned format? `[]byte` signature is not descriptive enough

test/e2e/main_test.go
eranco74: Sorry :man_facepalming:

test/e2e/nnce_conditions_test.go
```diff
@@ -105,6 +105,9 @@ var _ = Describe("EnactmentCondition", func() {
 		})
 
 		AfterEach(func() {
+			By("Remove the bridge")
+			updateDesiredState(linuxBrAbsent(bridge1))
+			waitForAvailableTestPolicy()
```
dankenigsberg: I'm not at all sure, but based on the /name/ of resetDesiredStateForNodes(), I would expect that cleaning the nodes should happen inside it.

phoracek: Resed desired state is removing the policy object. But removal of NNCP does not remove the interfaces it configured. Hence, we first need to remove what we created and only then remove the policy object.

dankenigsberg: isn't this order raceful? After the bridge is removed, the policy controller may wake up and re-create it. Or am I plain wrong?

phoracek: We changed the original policy with the `absent` desired state. So there should be no race.

eranco74:
@phoracek about the nodes_test.go failure:
i ran it a few times on my machine and it seems that sometimes the interface isn't deleted.
In the [logs](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/360/pull-kubernetes-nmstate-e2e-k8s/1222538314960080896/build-log.txt?Expires=1580314663&GoogleAccessId=prow-46%40kubevirt-240711.iam.gserviceaccount.com&Signature=CO5Ct3jVjcuuNGEYZfM%2B4XAkt6BV28BKvHxd6zPMv375bw2kLdm%2BcL26%2F4KijKHWPvkjJuDgpDeqLzO0r%2Ba0uOagRbjs9D%2Bgok7jfCDL%2F%2FmrgYkB27YSzZm76PyLrcOHw9WmpiZKJFh0pJ%2Fqlp6mzeeTSQxvf6mEq%2FmYrXxGLl6U5O5TqpuQCVQU9y42uxyM6ki9Xtx0Bj%2F6vSp8uvcTK7yD9hJvc%2BstJbn8A7ndJo6Utm7rHMbG22TAjRwhC%2FHd8qa3zG6DBAOhSWmRnqamPFW5S6yPSlp0qTrfURJz2Qe8yo91OBZLUVYVKUcmCzwEgDxLLlbiDdfcOYVE9Y3okg%3D%3D) I see:

> ./kubevirtci/cluster-up/ssh.sh node02 -- sudo nmcli con delete dummy0
stdout: 2020/01/29 15:52:03 Waiting for host: 192.168.66.102:22
2020/01/29 15:52:03 Connected to tcp://192.168.66.102:22
Connection 'dummy0' (6839202b-3562-4a8b-8844-603d8be99e0c) successfully deleted.
...
, stderr 
[1mSTEP[0m: Make sure the dummy interface was deleted
[1mSTEP[0m: Verifying initial state

[91m[1m• Failure in Spec Teardown (AfterEach) [192.620 seconds][0m
Nodes
[90m/tmp/knmstate/kubernetes-nmstate/test/e2e/nodes_test.go:8[0m
  when are up
  [90m/tmp/knmstate/kubernetes-nmstate/test/e2e/nodes_test.go:9[0m
    and new interface is configured
    [90m/tmp/knmstate/kubernetes-nmstate/test/e2e/nodes_test.go:25[0m
      [91m[1mshould update node network state with it [AfterEach][0m
      [90m/tmp/knmstate/kubernetes-nmstate/test/e2e/nodes_test.go:39[0m

      [91mTimed out after 180.000s.
      Expected
          <[]string | len:7, cap:8>: ["cni0", "docker0", "dummy0", "eth0", "eth1", "eth2", "lo"]
      not to contain element matching
          <string>: dummy0[0m

      /tmp/knmstate/kubernetes-nmstate/test/e2e/nodes_test.go:36

I can add an  `nmcli device delete dummy0`   command in case the interface is still there after a few seconds, thoughts?

phoracek:
@eranco74 I guess there may be a bug in NetworkManager. Please go for it, delete it

dankenigsberg:
> @eranco74 I guess there may be a bug in NetworkManager. Please go for it, delete it

If there is a bug in NM, do we want the test to pass? I'd feel more comfortable if we can have here an xfail as long as NM version has this specific bug.

eranco74:
> > @eranco74 I guess there may be a bug in NetworkManager. Please go for it, delete it
> 
> If there is a bug in NM, do we want the test to pass? I'd feel more comfortable if we can have here an xfail as long as NM version has this specific bug.

Still need to verify the actual root cause for this behavior, yet we do want the test to pass.
Flakey tests is not something we want to live with.

phoracek:
I personally don't care about failure to remove dummy interface netlink device enough to block any tests here. It fails due to test helpers, not on the subject of the test.

- [ ]

eranco74:
@tsorya any idea why there's s a "flannel.1" interface at the beginning of the "default IP address on top of the bond" test?
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/360/pull-kubernetes-nmstate-e2e-k8s/1222811236496838656

phoracek:
flannel.1 is bridge used by flannel to handle the primary network of Kubernetes. What is the issue?

eranco74:
The issue us that the interface exists at the end of the test but get's doesn't exists when the test starts.
So baseline validation fail.
I guess it happens since the node get rebooted during the test.

phoracek:
Sounds possible. Can we exclude flannel from the check that original ifaces are the same as current? We don't really care about that one.

test/e2e/main_test.go
```diff
@@ -24,16 +24,18 @@ import (
...
+	namespace            string
+	nodes                []string
+	startTime            time.Time
+	bond1                string
+	bridge1              string
+	primaryNic           = flag.String("primaryNic", "eth0", "Primary network interface name")
+	firstSecondaryNic    = flag.String("firstSecondaryNic", "eth1", "First secondary network interface name")
+	secondSecondaryNic   = flag.String("secondSecondaryNic", "eth2", "Second secondary network interface name")
+	nodesInterfacesState = make(map[string][]byte)
+	interfacesToIgnore   = []string{"flannel.1"}
```
qinqon: I think this is the same as veth we want to ignore them directly at knmstate [1]

[1] https://github.com/nmstate/kubernetes-nmstate/blob/master/deploy/operator.yaml#L78.

test/e2e/utils.go
```diff
@@ -523,6 +530,43 @@ func dhcpFlag(node string, name string) bool {
...
+			return true
+		}
+	}
+	return false
+}
+
+// return a json with all node interfaces and their state e.g.
+//{"cni0":"up","docker0":"up","eth0":"up","eth1":"down","eth2":"down","lo":"down"}
+// use exclude to filter out interfaces you don't care about
+func nodeInterfacesState(node string, exclude []string) []byte {
```
qinqon: ditto, we may not need exclude.

test/e2e/main_test.go
```diff
@@ -24,16 +24,18 @@ import (
...
+	t                    *testing.T
+	namespace            string
+	nodes                []string
+	startTime            time.Time
+	bond1                string
+	bridge1              string
+	primaryNic           = flag.String("primaryNic", "eth0", "Primary network interface name")
+	firstSecondaryNic    = flag.String("firstSecondaryNic", "eth1", "First secondary network interface name")
+	secondSecondaryNic   = flag.String("secondSecondaryNic", "eth2", "Second secondary network interface name")
+	nodesInterfacesState = make(map[string][]byte)
```
qinqon: What about adding a "previous" prefix or "expected" prefix to nodeInterfacesState ?

test/e2e/main_test.go
phoracek: I would not say so. There is only a couple of bridges on the host and not seeing a specific one is weird. On the other hand, with veths we are consistent by ignoring all of them.

qinqon: ack

PR #361, opened 2020-01-29
Run webhook only on master nodes

AlonaKaplan:
Currently all instances of the handler on all nodes serve as a webhook.
That includes both masters and workers.

Since workers are expected to be less stable than masters this commit
changes the behaviour to run webhook servers only on masters.

Since the webhooks are behind a Service we need to deploy two sets of
daemonsets, one will run only on workers, one only on masters.
The masters 'name' label would be selected in the Service selector.

Also, a RUN_WEBHOOK_SERVER="" env variable was added to the master pods.
In case the env var exists, the handler won't register+run the webhook.

Signed-off-by: Alona Kaplan <alkaplan@redhat.com>

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -120,10 +120,12 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Setup webhook
-	if err := webhook.AddToManager(mgr); err != nil {
-		log.Error(err, "Cannot initialize webhook")
-		os.Exit(1)
+	// Setup webhook on master only
+	if _, runWebhookServer := os.LookupEnv("RUN_WEBHOOK_SERVER"); runWebhookServer {
+		if err := webhook.AddToManager(mgr); err != nil {
+			log.Error(err, "Cannot initialize webhook")
+			os.Exit(1)
+		}
 	}
 
 	if err = serveCRMetrics(cfg); err != nil {
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -21,6 +21,7 @@ spec:
       serviceAccountName: nmstate-handler
       nodeSelector:
         beta.kubernetes.io/arch: amd64
+        node-role.kubernetes.io/master: ""
       tolerations:
         - key: node-role.kubernetes.io/master
           operator: Exists
@@ -36,6 +37,8 @@ spec:
           - kubernetes-nmstate
           env:
             - name: WATCH_NAMESPACE
+              value: ""
+            - name: RUN_WEBHOOK_SERVER
               value: ""
             - name: POD_NAME
               valueFrom:
@@ -68,6 +71,80 @@ spec:
           path: /run/dbus/system_bus_socket
           type: Socket
 ---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nmstate-handler-worker
+  namespace: nmstate
+spec:
+  selector:
+    matchLabels:
+      name: nmstate-handler-worker
+  template:
+    metadata:
+      labels:
+        app: kubernetes-nmstate
+        name: nmstate-handler-worker
+    spec:
+      # Needed to force vlan filtering config with iproute commands until
+      # future nmstate/NM is in place.
+      # https://github.com/nmstate/nmstate/pull/440
+      hostNetwork: true
+      serviceAccountName: nmstate-handler
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: node-role.kubernetes.io/master
+                    operator: NotIn
+                    values:
+                      - ""
+      nodeSelector:
+        beta.kubernetes.io/arch: amd64
+      containers:
+        - name: nmstate-handler
+          args:
+            - --v=production
+          # Replace this with the built image name
+          image: REPLACE_IMAGE
+          imagePullPolicy: Always
+          command:
+            - kubernetes-nmstate
+          env:
+            - name: WATCH_NAMESPACE
+              value: ""
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: "nmstate-handler"
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: NODE_NETWORK_STATE_REFRESH_INTERVAL
+              valueFrom:
+                configMapKeyRef:
+                  name: nmstate-config
+                  key: node_network_state_refresh_interval
+            - name: INTERFACES_FILTER
+              valueFrom:
+                configMapKeyRef:
+                  name: nmstate-config
+                  key: interfaces_filter
+          volumeMounts:
+            - name: dbus-socket
+              mountPath: /run/dbus/system_bus_socket
+          securityContext:
+            privileged: true
+      volumes:
+        - name: dbus-socket
+          hostPath:
+            path: /run/dbus/system_bus_socket
+            type: Socket
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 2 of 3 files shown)


cmd/manager/main.go
```diff
@@ -120,10 +121,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Setup webhook
-	if err := webhook.AddToManager(mgr); err != nil {
-		log.Error(err, "Cannot initialize webhook")
-		os.Exit(1)
+	// Setup webhook on master only
+	_, exist := os.LookupEnv("MASTER_NODE")
```
phoracek: Please make it imperative and explicit. "RUN_WEBHOOK_SERVER"

cmd/manager/main.go
```diff
@@ -26,6 +26,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/klog"
+	//client "github.com/nmstate/kubernetes-nmstate/pkg/helper"
```
phoracek: please drop this

cmd/manager/main.go
```diff
@@ -120,10 +121,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Setup webhook
-	if err := webhook.AddToManager(mgr); err != nil {
-		log.Error(err, "Cannot initialize webhook")
-		os.Exit(1)
+	// Setup webhook on master only
+	_, exist := os.LookupEnv("MASTER_NODE")
+	if exist {
```
phoracek: you can do this on a single line:

```
if _, runWebhookServer := os.LookupEnv(...); runWebhookServer {
```

phoracek:
Please fix the title and the commit message, we don't want "WIP" in our history.

Have you got a chance to test this on a multinoded OpenShift cluster?

AlonaKaplan:
> Have you got a chance to test this on a multinoded OpenShift cluster?

I tested it with multi node kubernetes cluster.

phoracek:
Thanks. Let's see if we can make CI happy. Thanks for the contribution!

hack/cluster-sync-handler.sh
```diff
@@ -10,19 +10,22 @@ sed "s#--v=production#--v=debug#" ${local_handler_manifest} | ${KUBECTL} create
 
 for i in {300..0}; do
     # We have to re-check desired number, sometimes takes some time to be filled in
-    desiredNumberScheduled=$(${KUBECTL} get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.desiredNumberScheduled}')
+    desiredMasterHandlersNumberScheduled=$(${KUBECTL} get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.desiredNumberScheduled}')
```
qinqon: Can we put this at a function passing the daemonset name and use it ?

hack/cluster-sync-handler.sh
```diff
@@ -10,19 +10,22 @@ sed "s#--v=production#--v=debug#" ${local_handler_manifest} | ${KUBECTL} create
...
+    desiredWorkerHandlersNumberScheduled=$(${KUBECTL} get daemonset -n nmstate nmstate-handler-worker -o=jsonpath='{.status.desiredNumberScheduled}')
 
-    numberAvailable=$(${KUBECTL} get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.numberAvailable}')
+    numberAvailableMasterHandlers=$(${KUBECTL} get daemonset -n nmstate nmstate-handler -o=jsonpath='{.status.numberAvailable}')
```
qinqon: A function would be nice too

hack/cluster-sync-handler.sh
```diff
@@ -10,19 +10,22 @@ sed "s#--v=production#--v=debug#" ${local_handler_manifest} | ${KUBECTL} create
...
+    numberAvailableWorkerHandlers=$(${KUBECTL} get daemonset -n nmstate nmstate-handler-worker -o=jsonpath='{.status.numberAvailable}')
 
-    if [ "$desiredNumberScheduled" == "$numberAvailable" ]; then
-        echo "nmstate-handler DS is ready"
+    if [ "$desiredMasterHandlersNumberScheduled" == "$numberAvailableMasterHandlers" ] && [ "$desiredWorkerHandlersNumberScheduled" == "$numberAvailableWorkerHandlers" ]; then
```
qinqon: You can replace ] && [ with " -a "

hack/cluster-sync-handler.sh
```diff
@@ -10,19 +10,22 @@ sed "s#--v=production#--v=debug#" ${local_handler_manifest} | ${KUBECTL} create
...
+        echo "nmstate-handler and nmstate-handler-worker DSs are ready"
         break
     fi
 
     if [ $i -eq 0 ]; then
```
qinqon: If we refactor this a little we can have something nice like 

```bash
if isOk "nmstate-handler-worker" && isOk "nmstate-handler"; then ...
```
So we will get the specific daemonset at logs.

deploy/operator.yaml
```diff
@@ -68,6 +71,80 @@ spec:
           path: /run/dbus/system_bus_socket
           type: Socket
 ---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nmstate-handler-worker
```
qinqon: Make sense to also change the other daemonset to nmstate-handler-master or it's problematic with upgrades and similar stuff ?

AlonaKaplan: It is problematic because of the upgrade. We will need to take care of deleting the old one.

phoracek: We have a mechanism for that in CNAO though :) and downtime of knmstate caused by the upgrade is not an issue. Let's keep the naming nice.

phoracek: We can rename it in a follow up though.

phoracek:
/lgtm
/approve
/hold

Waiting for @qinqon

hack/cluster-sync-handler.sh
```diff
@@ -8,21 +8,38 @@ ${KUBECTL} delete --ignore-not-found -f ${local_handler_manifest}
...
 
-    if [ "$desiredNumberScheduled" == "$numberAvailable" ]; then
-        echo "nmstate-handler DS is ready"
-        break
+        if [ "$desiredNumberScheduled" == "$numberAvailable" ]; then
+          echo "$1 DS is ready"
+          return 0
+        else
+          return 1
+        fi
```
qinqon: not needed the following is enough, since the lresult code of a function is from the last command in a function so just comparing stuff is enough.

```bash
[ "$desiredNumberScheduled" == "$numberAvailable" ]
```

hack/cluster-sync-handler.sh
```diff
@@ -8,21 +8,38 @@ ${KUBECTL} delete --ignore-not-found -f ${local_handler_manifest}
...
+}
+
+for i in {300..0}; do
+    # We have to re-check desired number, sometimes takes some time to be filled in
+    if isOk nmstate-handler && isOk nmstate-handler-worker ]; then
```
qinqon: typo: last bracket is not needed.

hack/cluster-sync-handler.sh
```diff
@@ -8,21 +8,38 @@ ${KUBECTL} delete --ignore-not-found -f ${local_handler_manifest}
...
+       break
     fi
 
     if [ $i -eq 0 ]; then
-        echo "nmstate-handler DS haven't turned ready within the given timeout"
+        echo "nmstate-handler or nmstate-handler-worker DS haven't turned ready within the given timeout"
```
qinqon: The failing "isOk" is going to show you wich one has fail, so we can put here
"kubernetes-nmstate hasn't turned ready within the given timeout"

hack/cluster-sync-handler.sh
AlonaKaplan: Hi @qinqon, maybe I don't understand your comment. But the function is not working without the return values. The result code of [ "$desiredNumberScheduled" == "$numberAvailable" ] is 0 whether the condition value is true or false.

hack/cluster-sync-handler.sh
AlonaKaplan: 'isOk' is printing just successful results not failures. It is possible to understand what handler caused the failure since it won't have a success message. But I would leave this messages as is, it is clearer to understand what happened (in case nmstate-handler-worker caused the failure, having the suggested message, the user will have to go to the scrtipt code to understand what exactly didn't turn ready).

PR #362, opened 2020-01-29
Improve log gathering:

tsorya:
Adding reporter that will dump all relevant logs to files that matches logtype
   On ci logs folder will be copied to artifacts

After every test we need to gather relevant logs.
Adding reporter that will run relevant commands and save output to files as <test>_<logtype>.log
It will add debug data such as device and connections statuses, networkmanager logs...
Hope it will help to debug failures.

It is very common to kubevirt reporters.
**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -36,6 +36,7 @@ var (
 	secondSecondaryNic   = flag.String("secondSecondaryNic", "eth2", "Second secondary network interface name")
 	nodesInterfacesState = make(map[string][]byte)
 	interfacesToIgnore   = []string{"flannel.1", "dummy0"}
+	reporter             = NewKubernetesNMStateReporter("test_logs/e2e")
 )
 
 var _ = BeforeSuite(func() {
@@ -48,6 +49,7 @@ var _ = BeforeSuite(func() {
 	nodeList := corev1.NodeList{}
 	err = framework.Global.Client.List(context.TODO(), &nodeList, &dynclient.ListOptions{})
 	Expect(err).ToNot(HaveOccurred())
+	reporter.BeforeSuiteDidRun()
 	for _, node := range nodeList.Items {
 		nodes = append(nodes, node.Name)
 	}
@@ -75,6 +77,7 @@ var _ = BeforeEach(func() {
 		nodeState := nodeInterfacesState(node, interfacesToIgnore)
 		nodesInterfacesState[node] = nodeState
 	}
+	reporter.dumpStateBeforeEach(getTestName())
 
 })
 
@@ -83,8 +86,8 @@ var _ = AfterEach(func() {
 	for _, node := range nodes {
 		nodeState := nodeInterfacesState(node, interfacesToIgnore)
 		Expect(nodesInterfacesState[node]).Should(MatchJSON(nodeState))
-		writePodsLogs(namespace, startTime, GinkgoWriter)
 	}
+	reporter.dumpStateAfterEach(getTestName(), namespace, startTime)
 })
 
 func getMaxFailsFromEnv() int {
--- a/test/e2e/reporter.go
+++ b/test/e2e/reporter.go
@@ -1,0 +1,143 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"io"
+	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"path"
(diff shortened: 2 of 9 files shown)


tsorya:
Logs are moved to artificats folder. Everything looks ok

test/e2e/main_test.go
```diff
@@ -46,7 +47,7 @@ var _ = BeforeSuite(func() {
 	nodeList := corev1.NodeList{}
 	err = framework.Global.Client.List(context.TODO(), &nodeList, &dynclient.ListOptions{})
 	Expect(err).ToNot(HaveOccurred())
-
+	reporter.BeforeSuiteDidRun()
```
qinqon: Instead of using it like this, I think we have to register the reporter at ginkgo with RunSpecsWithCustomReporters [1] for that we have to implement


```golang
type Reporter interface {
    SpecSuiteWillBegin(config config.GinkgoConfigType, summary *types.SuiteSummary)
    BeforeSuiteDidRun(setupSummary *types.SetupSummary)
    SpecWillRun(specSummary *types.SpecSummary)
    SpecDidComplete(specSummary *types.SpecSummary)
    AfterSuiteDidRun(setupSummary *types.SetupSummary)
    SpecSuiteDidEnd(summary *types.SuiteSummary)
}
```

[1]https://onsi.github.io/ginkgo/#writing-custom-reporters

test/e2e/reporter.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		for _, entery := range names {
+			os.RemoveAll(path.Join([]string{r.artifactsDir, entery.Name()}...))
+		}
+	}
+}
+
+
+func (r *KubernetesNMStateReporter) logNetworkManager(testName string, sinceTime time.Time){
+	r.OpenTestLogFile("netmanager", testName, func(writer io.ReadWriteCloser) {
```
qinqon: nm, network-manager, NetworkManager would be better.

qinqon:
This is awesome https://gcsweb.apps.ovirt.org/gcs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/362/pull-kubernetes-nmstate-e2e-k8s/1222823819492724736/artifacts/

test/e2e/main_test.go
qinqon: But we can do this in a follow up PR I think.

phoracek:
I'm sorry for raising this only now, but is there a way to tell which part of the log belongs to which test? Is it possible to at least label them or split to file-per-test? If not, it is still better than the current state.

test/e2e/main_test.go
```diff
@@ -34,6 +34,7 @@ var (
 	primaryNic         = flag.String("primaryNic", "eth0", "Primary network interface name")
 	firstSecondaryNic  = flag.String("firstSecondaryNic", "eth1", "First secondary network interface name")
 	secondSecondaryNic = flag.String("secondSecondaryNic", "eth2", "Second secondary network interface name")
+	reporter = NewKubernetesNMStateReporter("test_logs/e2e")
```
phoracek: seems like you did not run `go fmt` against the module

test/e2e/utils.go
```diff
@@ -534,7 +497,13 @@ func defaultRouteNextHopInterface(node string) AsyncAssertion {
...
 	}, 15*time.Second, 1*time.Second)
 }
+
 func vlan(node string, iface string) string {
 	vlanFilter := fmt.Sprintf("interfaces.#(name==\"%s\").vlan.id", iface)
 	return gjson.ParseBytes(currentStateJSON(node)).Get(vlanFilter).String()
 }
+
+func getTestName() string{
+	fileName := path.Base(CurrentGinkgoTestDescription().FileName)
```
phoracek: Can we get name of the specific test here?

test/e2e/main_test.go
tsorya: i would prefer :)

test/e2e/utils.go
tsorya: What do u mean? specific test case? not file?

phoracek: Yes, each `It` represented by a file. Now, when one of the tests fail, the log contains the whole module, so it may be harder to find the correct spot. But again, it all depends on the difficulty of such task.

tsorya: Not sure it so  easy. The easiest way is to run dump from AfterEach function of each test but then it will add run time for each case not sure it worth it.

test/e2e/main_test.go
tsorya: Done

test/e2e/main_test.go
qinqon: fair enough let's merge it ASAP is super nice addition!

test/e2e/utils.go
qinqon: This is the kind of thing that get improved with use, let's merge it and ideas will pop up.

PR #363, opened 2020-01-29
drop user guides root

phoracek:

The root is duplicate to the list we already keep in README.md. Since the
README is way more visible and tutorials should be as accessible as possible,
the root will be dropped and only README.md kept.

**Release note**:

```release-note
NONE
```


phoracek:

Unrelated for this documentation PR.

PR #364, opened 2020-01-30
Add the flakefinder link to the knmstate welcome page

qinqon:
We have integrate flakefinder to knmstate, let's just put a link here
so we can check this out.

Also would be nice to implement badges at flakefinder so we have some 
kind of CI semaphore.

**Release note**:

```release-note
NONE
```


README.md
```diff
@@ -1,3 +1,7 @@
+# CI
```
phoracek: Landing page is for people that see the project for the first time. CI results are irrelevant and confusing at that point. Please add this section to CONTRIBUTING.md. It should include explanation what is flakefinder good for.

PR #366, opened 2020-01-30
Count seconds instead of nano for default gw retrieval

qinqon:
A mechanism to poll default gw until it's there was implemented, but we forgot to specify the timeout as seconds, so it's using golang default wich is nanoseconds, that's why see still the "impossible to retrieve default gw" sometimes at e2e tests.

Also fixed PollImmediate at client.go there were a lot of places that we want to retry but an error was returned, at error PollImmediate exists with it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -226,7 +226,7 @@ func checkApiServerConnectivity(timeout time.Duration) error {
 
 func defaultGw() (string, error) {
 	defaultGw := ""
-	return defaultGw, wait.PollImmediate(1*time.Second, defaultGwRetrieveTimeout, func() (bool, error) {
+	return defaultGw, wait.PollImmediate(1*time.Second, defaultGwRetrieveTimeout*time.Second, func() (bool, error) {
 		observedStateRaw, err := show()
 		if err != nil {
 			return false, fmt.Errorf("error running nmstatectl show: %v", err)
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -35,9 +35,9 @@ var (
 
 const nmstateCommand = "nmstatectl"
 const vlanFilteringCommand = "vlan-filtering"
-const defaultGwRetrieveTimeout = 120
-const defaultGwProbeTimeout = 120
-const apiServerProbeTimeout = 120
+const defaultGwRetrieveTimeout = 120 * time.Second
+const defaultGwProbeTimeout = 120 * time.Second
+const apiServerProbeTimeout = 120 * time.Second
 
 var (
 	interfacesFilterGlob glob.Glob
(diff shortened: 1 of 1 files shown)


pkg/helper/client.go
```diff
@@ -214,33 +214,37 @@ func checkApiServerConnectivity(timeout time.Duration) error {
...
-			return false, fmt.Errorf("Impossible to convert current state to JSON: %v", err)
+			return false, errors.Wrap(err, "failed to convert current state to JSON")
 		}
 
-		defaultGw = gjson.ParseBytes([]byte(currentState)).
+		defaultGw = gjson.ParseBytes(currentState).
 			Get("routes.running.#(destination==\"0.0.0.0/0\").next-hop-address").String()
 		if defaultGw == "" {
-			return false, fmt.Errorf("Impossible to retrieve default gw, state: %s", string(observedStateRaw))
+			log.Info("default gw missing", map[string]string{"state": string(currentState)})
```
phoracek: why do you log a map? seems like an overkill. maybe i miss the motivation

pkg/helper/client.go
```diff
@@ -214,33 +214,37 @@ func checkApiServerConnectivity(timeout time.Duration) error {
...
 		if err != nil {
-			return false, fmt.Errorf("error running nmstatectl show: %v", err)
+			log.Error(err, fmt.Sprintf("failed retrieving current state"))
+			return false, nil
 		}
 
 		currentState, err := yaml.YAMLToJSON([]byte(observedStateRaw))
 		if err != nil {
-			return false, fmt.Errorf("Impossible to convert current state to JSON: %v", err)
+			return false, errors.Wrap(err, "failed to convert current state to JSON")
```
phoracek: So in this case we want to cancel the poll?

pkg/helper/client.go
```diff
@@ -214,33 +214,37 @@ func checkApiServerConnectivity(timeout time.Duration) error {
...
+			return false, nil
 		}
 		return true, nil
 	})
 }
 
 func defaultGw() (string, error) {
 	defaultGw := ""
-	return defaultGw, wait.PollImmediate(1*time.Second, defaultGwRetrieveTimeout, func() (bool, error) {
+	return defaultGw, wait.PollImmediate(1*time.Second, defaultGwRetrieveTimeout*time.Second, func() (bool, error) {
```
phoracek: maybe the timeout constant should be initialized with the unit as well

pkg/helper/client.go
qinqon: I am using this, but now that I see it it's wrong I have to past just the "state", value:
```golang
    // Info logs a non-error message with the given key/value pairs as context. 
    //                                                                          
    // The msg argument should be used to add some constant description to      
    // the log line.  The key/value pairs can then be used to add additional    
    // variable information.  The key/value pairs should alternate string       
    // keys and arbitrary values.                                               
    Info(msg string, keysAndValues ...interface{})                              
```

pkg/helper/client.go
qinqon: I suppose not being able to transform yaml to json is not transitory, but don't know what will happy with empty, since maybe empty can be transitory value.

pkg/helper/client.go
qinqon: Ahh right, let me do that thanks !

pkg/helper/client.go
```diff
@@ -112,7 +112,7 @@ func set(desiredState nmstatev1alpha1.State) (string, error) {
 		// commit timeout doubles the default gw ping probe timeout, to
 		// ensure the Checkpoint is alive before rolling it back
 		// https://nmstate.github.io/cli_guide#manual-transaction-control
-		output, err = nmstatectl([]string{"set", "--no-commit", "--timeout", strconv.Itoa(defaultGwProbeTimeout * 2)}, string(desiredState.Raw))
+		output, err = nmstatectl([]string{"set", "--no-commit", "--timeout", strconv.Itoa((defaultGwProbeTimeout / time.Second) * 2)}, string(desiredState.Raw))
```
phoracek: No need for this monkey business https://golang.org/pkg/time/#Duration.Seconds

PR #368, opened 2020-01-30
docs: 101 state

phoracek:

This patch replaces the former state tutorial with 101 introduction
article about state reporting. Follow-up patches will extend the 101
series and cover the whole API.

We miss proper documentation. Let's help users get into knmstate.

**Release note**:

```release-note
NONE
```


docs/user-guide-101-reporting-state.md
```diff
@@ -0,0 +1,192 @@
...
+```shell
+kubectl get nodenetworkstates
+```
+
+```
+NAME     AGE
+node01   22s
+```
+
+You can also use short name `nns` to reach the same effect:
```
qinqon: Advise with a `NOTE:` that first time you use shortname it fails.

PR #369, opened 2020-01-30
Changing the way format and check works

tsorya:
make format will generate format changes and not only print them
make check will fail if there are files that are not formatted
adding test/e2e to format list
fixing files with format issues

make format will generate format changes and not only print them
make check will fail if there are files that are not formatted
adding test/e2e to format list
I am sure that it is better to fail on build while the are files that needs to be formatted than to find out it on review.
As for me it is much better to have generated code when one runs make format than just to print changes.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,10 +56,9 @@ description = build/_output/description
 resources = deploy/namespace.yaml deploy/service_account.yaml deploy/role.yaml deploy/role_binding.yaml
 all: check handler
 
-check: format vet whitespace-check
+check: vet whitespace-check gofmt-check
 
-format: whitespace-format $(GO)
-	$(GOFMT) -d cmd/ pkg/
+format: whitespace-format gofmt
 
 vet: $(GO)
 	$(GO) vet ./cmd/... ./pkg/...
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -56,10 +56,11 @@ description = build/_output/description
 resources = deploy/namespace.yaml deploy/service_account.yaml deploy/role.yaml deploy/role_binding.yaml
 all: check handler
 
-check: format vet whitespace-check
+check: vet whitespace-check
+	test -z "`$(GOFMT) -l test/e2e/`"
```
qinqon: same as whitespace-check make a format-check Makefiel target, also add cmd/ and pkg/, also not that you have to add $(GO) as dependency so it install golang if it's not already there

In short:
```Makefile
gofmt-check: $(GO)
	test -z "`$(GOFMT) -l cmd/ pkg/ test/e2e/`"

check: vet whitespace-check gofmt-check
```

Makefile
```diff
@@ -56,10 +56,11 @@ description = build/_output/description
...
 
 format: whitespace-format $(GO)
```
qinqon: Same here:

```Makefile
gofmt: $(GO)
	$(GOFMT) -w cmd/ pkg/ test/e2e/

```

PR #370, opened 2020-01-31
We don't care about the list of file inside golang tarball

qinqon:
Inspecting the logs with at long list of files we don't care at the beginning make no sense, let's
be quiet unpacking golang tarball.

**Release note**:

```release-note
NONE
```


phoracek:
Half of the tests failed. What the heck

qinqon:
@phoracek checking, also the logs retrievial mess up the test logs, I t was nise to see the pod logs per test at prow.

PR #373, opened 2020-01-31
Gather e2e tests logs in parallel and improve some logs

qinqon:
Now e2e tests gather different logs before/after every test we have over 25 specs so it add
some time to the full test run, we can gather stuff in parallel to reduce those.

Also currentState is printed as string when retreving default gw, so is human compatible.

It moves all the runAt e2e tests functions to `runnner` package and the reporter to `report` package
under `test/e2e`

**Release note**:

```release-note
NONE
```


qinqon:
Does it make sense to gather logs on success ? maybe we can skip them.

qinqon:
/hold

device status retrieval is broken
```
    Expected
        <bool>: true
    to equal
        <bool>: false

    /home/ellorent/Documentos/cnv/upstream/knmstate/kubernetes-nmstate/test/e2e/node_selector_test.go:32
------------------------------
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x13b30ea]

goroutine 177 [running]:
git
```

tsorya:
@qinqon  i think it is better to start by gathering logs even if test succeeded and to change it further but not necessary.

tsorya:
Nice changes !!!

test/e2e/reporter.go
```diff
@@ -123,21 +184,32 @@ func (r *KubernetesNMStateReporter) logPods(testName string, namespace string, s
 			io.WriteString(writer, formattedLogs)
 		}
 	})
+
+	// Let's print the pods logs to the GinkgoWriter so
+	// we see the failure directly at prow junit output without opening files
+	if podsLogFile != "" {
```
tsorya: Why to read file if you already have those logs in formattedLogs?

test/e2e/reporter.go
```diff
@@ -123,21 +184,32 @@ func (r *KubernetesNMStateReporter) logPods(testName string, namespace string, s
...
+		podsLog, err := ioutil.ReadFile(podsLogFile)
+		if err != nil {
+			fmt.Println(err)
+			return nil
+		}
+		GinkgoWriter.Write(podsLog)
```
tsorya: I think that if test didn't fail there is no need to print logs to stdout for sure

eranco74: Yet, I think we want to have the logs collected to a file, so we will have the option to compare the success logs with failed ones

test/e2e/reporter.go
qinqon: Ahh you right I was just lazy, let's make func(io.Writer) be func(io.Writer) string and return it so it's used elsewhere

test/e2e/reporter.go
qinqon: We run this only on failure, so it will be write only then.

qinqon:
@tsorya after talking with @phoracek let's run only on failures, there is a lot of stuff going on on this reporter and for now is good enough.

qinqon:
/hold cancel

we are good at gathering device status now.

qinqon:
@tsorya @eranco74 are we good here ?

PR #380, opened 2020-02-03
Default to 0 if numberAvailable is not present

qinqon:
Now that there is two daemonsets one for master and one for worker
the one master node scenario we have by default at knmstate
does not work well, since numberAvailable value is not present so
it an infinite loop of "'0' == ''", to fix that we can just return 0
is the value is not set.

**Release note**:

```release-note
NONE
```


qinqon:
/cc @AlonaKaplan 

I miss this at the review, sorry :-/

PR #383, opened 2020-02-04
Move e2e libs to separate path

qinqon:
Looks like having then under test breaks ginkgo output and
does not shows until the end

Also add ./test to vet and put correct package name at modules

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/reporter/writers.go
+++ b/test/reporter/writers.go
@@ -1,4 +1,4 @@
-package e2e
+package reporter
 
 import (
 	"context"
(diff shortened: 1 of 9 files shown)


test/reporter/writers.go
```diff
@@ -1,4 +1,4 @@
-package e2e
```
phoracek: This may have been the cause? It was in reporter directory, but under the name e2e? In any case, I prefer it being at the top level of test

qinqon: Nah, tried that too, was still showing no logs, the real trick was moving the modules out of test/e2e.

PR #385, opened 2020-02-05
test/e2e: Set secondary nics down at teardown

qinqon:
After a check at e2e test was added to see if we have extra interfaces not cleaned up 
after the test, looks like sometimes secondary nics has up/down unexpected state, this
PR enforces them to down since is their state after cluster-up.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/states.go
+++ b/test/e2e/states.go
@@ -2,16 +2,36 @@ package e2e
 
 import (
 	"fmt"
+	"strings"
+	"text/template"
+
+	. "github.com/onsi/gomega"
 
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-func ethernetNicUp(nicName string) nmstatev1alpha1.State {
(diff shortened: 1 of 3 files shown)


test/e2e/states.go
```diff
@@ -2,16 +2,36 @@ package e2e
...
+
+	. "github.com/onsi/gomega"
 
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
 )
 
-func ethernetNicUp(nicName string) nmstatev1alpha1.State {
-	return nmstatev1alpha1.NewState(fmt.Sprintf(`interfaces:
-  - name: %s
+func ethernetNicsState(states map[string]string) nmstatev1alpha1.State {
```
phoracek: Could you please just create multiple policies, one per each NIC? So we decrease the number of helper function, plus reading the template is hard.

qinqon: Well we have around 24 test now we do 24 policy apply to do that if we do per nic that would be 72 so we open the door to something going bad network wise and is less atomic.

PR #386, opened 2020-02-05
Adding profiler:

tsorya:
Using pprof profiler and http listener to give and opportunity for code profiling on the runtime
	Profiler can be started by changing ENABLE_PROFILER to True
        Profiler port can be modified by changing PROFILER_PORT
	Usage:
	  1. Make port forwading to relevant profiler port by default 6060
	  2. Use go tool pprof to gather relevant metrics
	Examples of pprof usage can be found here: https://golang.org/pkg/net/http/pprof/
	Adding github.com/kelseyhightower/envconfig to read os environs to struct for clearness

Signed-off-by: Igal Tsoiref <itsoiref@redhat.com>

Adding option to profile kubernetes-nmstate

**Release note**:

```release-note
NONE
```


deploy/operator.yaml
```diff
@@ -60,6 +60,10 @@ spec:
                 configMapKeyRef:
                   name: nmstate-config
                   key: interfaces_filter
+            - name: ENABLE_PROFILER
+                value: "False"
+            - name: PROFILER_PORT
```
phoracek: Please add it to the port section of the pod: https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/

cmd/manager/main.go
```diff
@@ -186,3 +195,20 @@ func serveCRMetrics(cfg *rest.Config) error {
 	}
 	return nil
 }
+
+// Start profiler on given port if ENABLE_PROFILER is True
+func setProfiler() {
+	cfg := ProfilerConfig{}
+	_ = envconfig.Process("", &cfg)
```
phoracek: If we don't care about the return value, we can drop the left hand side

CONTRIBUTING.md
```diff
@@ -136,7 +135,18 @@ So the step would be:
...
 
-# CI infraestructure
+# Profiling
+
+There is a possibility to enable golang pprof profiler.
+ - Enable profiler in `operator.yaml` by changing value of 'ENABLE_PROFILER' to True
+ - You can change profiler port by editing 'PROFILER_PORT' - default is 6060
+ - Deploy new code to cluster - example:  "make cluster-sync"
+ - Find nmstate-handler pod name - "kubectl get pods -n nmstate"
+ - Create port forwarding to pod - example: "kubectl port-forward pod/<pod name> 6060:6060 -n nmstate"
```
phoracek: this renders as "pod/", <> are omitted.

Would you please keep all command examples as inline code (enclose them within `)

CONTRIBUTING.md
```diff
@@ -136,7 +135,18 @@ So the step would be:
...
+ - Use "go tool pprof" to gather relevant metrics.
```
phoracek: Could you should two examples, one for cpu and one for memory?

phoracek: I think you've skipped the step where you need to download the profile. Or at least all examples are found are using curl and only then `go tool pprof`

CONTRIBUTING.md
```diff
@@ -136,7 +135,18 @@ So the step would be:
...
+   Examples can be found here: https://golang.org/pkg/net/http/pprof/
```
phoracek: You can also mention that the pprof can be accessed via browser on http://localhost:6060/debug/pprof/

CONTRIBUTING.md
phoracek: I see you can...

go tool pprof -png http://localhost:6060/debug/pprof/heap > out.png

tsorya: u can just run go tool pprof -png http://localhost:6060/debug/pprof/heap
It will give u command line with many options nad from there u can create graphs and other stuff

PR #378, opened 2020-02-03
Fix ocp lane

qinqon:
Signed-off-by: Quique Llorente <ellorent@redhat.com>

**What this PR does / why we need it**:

Our OCP lane is broken, we fix it by:

- Use ocp-4.4 provider
- Use env to get nics names at e2e tests
- Fix 'failed to list *unstructured.Unstructured' log
- Continue loop if kubectl.sh fails at cluster-sync
- Fix core dump at e2e test when EOF
- Skip OVS e2e tests for okd/ocp lanes
- Parameterize e2e test timeout

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #387, opened 2020-02-06
use NM 1.22 on Kubernetes kubevirtci

phoracek:

In order to test the proper NetworkManager for nmstate 0.2, we will run
our CI against NetworkManager 1.22. This should prepare us for future on
RHEL 8.2.

We should keep another lane with OCP and NetworkManager 1.20 in order to
make sure that the current release is still fine.

**Release note**:

```release-note
NONE
```


phoracek:
This should go in only after https://github.com/nmstate/kubernetes-nmstate/pull/378

phoracek:
OCP job got stuck for hours.

PR #389, opened 2020-02-06
use specific commitID for commit/rollback

phoracek:

Limit space for races and bugs.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -124,13 +124,13 @@ func set(desiredState nmstatev1alpha1.State) (string, error) {
 	return output, err
 }
 
-func commit() (string, error) {
-	return nmstatectl([]string{"commit"}, "")
+func commit(commitID string) (string, error) {
+	return nmstatectl([]string{"commit", commitID}, "")
 }
 
-func rollback(cause error) error {
-	_, err := nmstatectl([]string{"rollback"}, "")
-	return fmt.Errorf("rollback cause: %v, rollback error: %v", cause, err)
+func rollback(commitID string, cause error) error {
+	_, err := nmstatectl([]string{"rollback", commitID}, "")
+	return errors.Wrapf(err, "rollback failed, cause: %s, commitID: %s", cause, commitID)
 }
 
 func GetNodeNetworkState(client client.Client, nodeName string) (nmstatev1alpha1.NodeNetworkState, error) {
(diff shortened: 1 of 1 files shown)


pkg/helper/client.go
```diff
@@ -124,12 +124,12 @@ func set(desiredState nmstatev1alpha1.State) (string, error) {
...
-	return nmstatectl([]string{"commit"}, "")
+func commit(commitID string) (string, error) {
+	return nmstatectl([]string{"commit", commitID}, "")
 }
 
-func rollback(cause error) error {
-	_, err := nmstatectl([]string{"rollback"}, "")
+func rollback(commitID string, cause error) error {
+	_, err := nmstatectl([]string{"rollback", commitID}, "")
 	return fmt.Errorf("rollback cause: %v, rollback error: %v", cause, err)
```
qinqon: Can you add commit id to to error and use errors.Wrapf ?

pkg/helper/client.go
```diff
@@ -251,6 +251,15 @@ func defaultGw() (string, error) {
 	})
 }
 
+// After running `nmstatectl set`, the last line contains the commit/rollback ID in format of:
+// Checkpoint: <ID>
+func readCommitID(setOutput string) string {
```
qinqon: setOutput feels like an action, let's call it outputFromSet or nmstateOutput.

pkg/helper/client.go
```diff
@@ -251,6 +251,15 @@ func defaultGw() (string, error) {
...
+	setOutputLines := strings.Split(strings.TrimSpace(setOutput), "\n")
+	commitIDLine := strings.TrimSpace(setOutputLines[len(setOutputLines)-1])
+	commitID := strings.Split(commitIDLine, " ")[1]
+	return commitID
```
qinqon: Can we do all this with a regex  and capturing group ? something like "^CheckPoint: (.*)$"

pkg/helper/client.go
```diff
@@ -261,46 +270,48 @@ func ApplyDesiredState(desiredState nmstatev1alpha1.State) (string, error) {
...
+	commitID := readCommitID(setOutput)
+
 	// Future versions of nmstate/NM will support vlan-filtering meanwhile
 	// we have to enforce it at the desiredState bridges and outbound ports
 	// they will be configured with vlan_filtering 1 and all the vlan id range
 	// set
 	bridgesUpWithPorts, err := getBridgesUp(desiredState)
 	if err != nil {
-		return "", rollback(fmt.Errorf("error retrieving up bridges from desired state"))
+		return "", rollback(commitID, fmt.Errorf("error retrieving up bridges from desired state"))
```
qinqon: Let's start to rall Wrapf little by little is the defacto stuff fro future 1.13 but we can start to use it at 1.12 with github.com/pkg/errors

pkg/helper/client.go
phoracek: I prefer basic tools if it is as simple as this. Regex is powerful but terrible to read.

pkg/helper/client.go
phoracek: That would deserve a refactoring of the module. I don't want to mess too much with this PR.

We should refactor the helper module completely

phoracek:
/override  pull-kubernetes-nmstate-e2e-ocp

The provider failed to be started.

phoracek:
/hold

Seems like this change broke rollback

PR #394, opened 2020-02-07
Replace tarball go vendoring with containerized vendoring

qinqon:

Current we vendor golang downloading the tarball and installing it under the project, that's
suboptimal since golang is share by a lot of project, this PR uses the golang container to compile
but compiles outside of it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/go-container.sh
+++ b/hack/go-container.sh
@@ -1,0 +1,94 @@
+#!/bin/sh
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


phoracek:
@qinqon can we avoid editing the script by exporting our custom variables to it? e.g. for the path to output. So later we can just copy a new version of the script without adjusting to our needs.

hack/go-container.sh
```diff
@@ -0,0 +1,94 @@
...
+# Usage:
+#  hack/go.sh go version
+#  hack/go.sh go build -o /out/kind .
+set -o nounset -o errexit
+
+# ============================ SCRIPT SETTINGS =================================
+# get the repo root for defaulting OUT_DIR and SOURCE_DIR
+REPO_ROOT="${REPO_ROOT:-$(cd "$(dirname "$0")/.." && pwd)}"
+# output directory, will be mounted to /out, defaults to /bin in REPO_ROOT
+OUT_DIR="${OUT_DIR:-${REPO_ROOT}/build/_output/bin}"
```
phoracek: Is this what was in the original file? If not and you are adjusting it for this repo, you better do it in hack/go by exporting variables.

PR #396, opened 2020-02-09
Adding config map controller

tsorya:
Making configuration map optional.
Adding config map controller that listen on events and run reconcile only if event is relevant to nmstate-config

@phoracek Just wanted to check if i got right the ticket, still no tests are added.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -53,11 +53,13 @@ spec:
             - name: NODE_NETWORK_STATE_REFRESH_INTERVAL
               valueFrom:
                 configMapKeyRef:
+                  optional: true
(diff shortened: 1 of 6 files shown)


phoracek:
@tsorya this was the option we did not want to do :D Additional controller is not needed, it only adds complexity with more communication going on between the handler and API server. We just merely want to mount the config map as a file (not an environment variable as done now) and run an fswatch on it. If it changes, we restart. I think it should be possible and quite simpler and less error prone. Sorry if I was not clear enough.

qinqon:
@tsorya can you mark this as WIP ?

PR #397, opened 2020-02-10
test/e2e: Check master/worker daemonsets

qinqon:
After splitting the daemonset intoo two, to have the webhook running only at master
we have to update also test/e2e so we check both are ok.

Extra mile: Move test/e2e prepare stuff to prepare.go

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/prepare.go
+++ b/test/e2e/prepare.go
@@ -1,0 +1,53 @@
+package e2e
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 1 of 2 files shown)


phoracek:
@qinqon since you move the function from one place to another, it is hard to spot the actual difference. Would you please do it in two commits? One moving the function, one fixing it in regards of multiple DaemonSets?

qinqon:
@phoracek done

PR #404, opened 2020-02-10
tests: fix cleanup function

phoracek:

We merged Skip into AfterEach. It should be in BeforeEach. In the current
state, it causes this cleanup to be skipped and therefore the test
leaves mess behind.

**Release note**:

```release-note
NONE
```


phoracek:
@qinqon

PR #405, opened 2020-02-11
Move nmstatectl logic to separate package

qinqon:
We have all the logic mixed up at pkg/helper/client.go, this will
move that to pkg/nmstatectl/nmstatectl.go

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/nmstatectl/nmstatectl.go
+++ b/pkg/nmstatectl/nmstatectl.go
@@ -1,0 +1,74 @@
+package nmstatectl
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os/exec"
+	"strconv"
+	"strings"
+	"time"
+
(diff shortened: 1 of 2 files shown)


pkg/nmstatectl/nmstatectl.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+var (
+	log = logf.Log.WithName("nmstatectl")
+)
+
+const nmstateCommand = "nmstatectl"
+
+func Show(arguments ...string) (string, error) {
+	cmd := exec.Command(nmstateCommand, "show")
```
phoracek: Could we use nmstatectl here too?

qinqon: Yep, let's pass an empty input.

PR #406, opened 2020-02-11
cleanup vlan100 Policy

phoracek:

We never cleaned up the policy, that resulted to vlan100 being randomly
created during the tests execution.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/demos_test.go
+++ b/test/e2e/demos_test.go
@@ -39,6 +39,13 @@ var _ = Describe("Introduction", func() {
 		kubectlAndCheck("wait", "nncp", "vlan100", "--for", "condition=Available", "--timeout", "2m")
 	}
 
+	// Policies are not deleted as a part of the tutorial, so we need additional function here
+	cleanupConfiguration := func() {
+		deletePolicy("vlan100")
+		updateDesiredState(interfaceAbsent("eth1.100"))
+		waitForAvailableTestPolicy()
+	}
+
 	runTroubleshooting := func() {
 		kubectlAndCheck("apply", "-f", "docs/examples/eth666_up.yaml")
 		kubectlAndCheck("wait", "nncp", "eth666", "--for", "condition=Degraded", "--timeout", "2m")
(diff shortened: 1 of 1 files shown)


test/e2e/demos_test.go
```diff
@@ -39,6 +39,15 @@ var _ = Describe("Introduction", func() {
 		kubectlAndCheck("wait", "nncp", "vlan100", "--for", "condition=Available", "--timeout", "2m")
 	}
 
+	// Policies are not deleted as a part of the tutorial, so we need additional function here
+	cleanupConfiguration := func() {
+		deletePolicy("eth1")
+		deletePolicy("eth2")
```
qinqon: This are delete by `"delete", "nncp", "bond0-eth1-eth2"` although the deletePolicy failing if the policy is not there feels REALLY wrong.

phoracek: You are right

test/e2e/demos_test.go
```diff
@@ -39,6 +39,13 @@ var _ = Describe("Introduction", func() {
 		kubectlAndCheck("wait", "nncp", "vlan100", "--for", "condition=Available", "--timeout", "2m")
 	}
 
+	// Policies are not deleted as a part of the tutorial, so we need additional function here
+	cleanupConfiguration := func() {
+		deletePolicy("vlan100")
+		updateDesiredState(interfaceAbsent("eth1.100"))
+		waitForAvailableTestPolicy()
```
qinqon: since we use the TestPolicy to remove the interface we have to call after this 
resetDesiredStateForNodes()

phoracek: I don't get this. Here we use test policy to remove the VLAN, we wait until it succeeds. Good. Then we use the same policy to cleanup NICs, I don't see where is the collision

qinqon: So tutorals play around with eth1 and eth2 and that change it states so we fail on 
```yaml
/tmp/knmstate/kubernetes-nmstate/test/e2e/main_test.go:92
Expected
    <string>: {
      "cni0": "up",
      "docker0": "up",
      "eth0": "up",
      "eth1": "down",
      "eth2": "down",
      "lo": "down"
    }
to match JSON of
    <string>: {
      "cni0": "up",
      "docker0": "up",
      "eth0": "up",
      "eth1": "up",
      "eth2": "up",
      "lo": "down"
    }
```
That's why we decide on add put secondary nics down after each test if we don't really care about sideffect of using them.

test/e2e/demos_test.go
```diff
@@ -68,6 +73,10 @@ var _ = Describe("Introduction", func() {
 	})
 
 	Context("All tutorials in a row", func() {
+		AfterEach(func() {
```
qinqon: btw you can do AfterEach(cleanConfiguration)

phoracek: Right!

PR #408, opened 2020-02-12
Making config map configuration dynamic, adding fswatcher for config file

tsorya:
Make configmap optional and creating fs watch to it.
Configmap must be created from file another way it will add file per volume
kubectl -n nmstate create configmap nmstate-config --from-file=nmstate-config.yaml 
 
Tests needs to be added if it will be relevant
**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/nmstate/kubernetes-nmstate/pkg/apis"
+	"github.com/nmstate/kubernetes-nmstate/pkg/configurations"
 	"github.com/nmstate/kubernetes-nmstate/pkg/controller"
 	"github.com/nmstate/kubernetes-nmstate/pkg/webhook"
 	"github.com/nmstate/kubernetes-nmstate/version"
@@ -91,6 +92,19 @@ func main() {
 	namespace, err := k8sutil.GetWatchNamespace()
 	if err != nil {
 		log.Error(err, "Failed to get watch namespace")
+		os.Exit(1)
+	}
+
+	// Creating global config
+	err = configurations.CreateGlobalConfig()
+	if err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	// Start configuration watcher
+	if err := configurations.Start(configurations.GetConfigPath(), configurations.SetConfig); err != nil {
+		log.Error(err, "Failing to start configuration watcher")
 		os.Exit(1)
 	}
 
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -50,20 +50,19 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-            - name: INTERFACES_FILTER
-              valueFrom:
-                configMapKeyRef:
-                  name: nmstate-config
-                  key: interfaces_filter
             - name: WEBHOOK_PORT
               value: "54874"
             - name: ENABLE_PROFILER
               value: "False"
             - name: PROFILER_PORT
               value: "6060"
+            - name: CONFIG_PATH
+              value: "/etc/config/nmstate.yaml"
           volumeMounts:
           - name: dbus-socket
             mountPath: /run/dbus/system_bus_socket
+          - name: config-volume
+            mountPath: /etc/config
           securityContext:
             privileged: true
       volumes:
@@ -71,6 +70,19 @@ spec:
         hostPath:
           path: /run/dbus/system_bus_socket
           type: Socket
+      - name: config-volume
+        configMap:
+          # Provide the name of the ConfigMap containing the file
+          # configmap can be created from file or with this data :
+          # Example:
+          # data:
+          #   nmstate.yaml: |
+          #     interfaces_filter: "veth*"
+          # it's full path must
+          # be set in CONFIG_PATH param
+          # to add to the container
+          name: nmstate-config
+          optional: true
 ---
 apiVersion: apps/v1
 kind: DaemonSet
@@ -125,18 +137,17 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-            - name: INTERFACES_FILTER
-              valueFrom:
-                configMapKeyRef:
-                  name: nmstate-config
-                  key: interfaces_filter
             - name: ENABLE_PROFILER
               value: "False"
             - name: PROFILER_PORT
               value: "6060"
+            - name: CONFIG_PATH
+              value: "/etc/config/nmstate.yaml"
           volumeMounts:
             - name: dbus-socket
               mountPath: /run/dbus/system_bus_socket
+            - name: config-volume
+              mountPath: /etc/config
           securityContext:
             privileged: true
       volumes:
@@ -144,14 +155,19 @@ spec:
           hostPath:
             path: /run/dbus/system_bus_socket
             type: Socket
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: nmstate-config
-  namespace: nmstate
-data:
-  interfaces_filter: "veth*"
+        - name: config-volume
+          configMap:
+            # Provide the name of the ConfigMap containing the file
+            # configmap can be created from file or with this data :
+            # Example:
+            # data:
+            #   nmstate.yaml: |
+            #     interfaces_filter: "veth*"
+            # it's full path must
+            # be set in CONFIG_PATH param
+            # to add to the container
+            name: nmstate-config
+            optional: true
 ---
 apiVersion: v1
 kind: Service
--- a/pkg/configurations/configurations.go
+++ b/pkg/configurations/configurations.go
@@ -1,0 +1,135 @@
+package configurations
+
+import (
+	"fmt"
+	"github.com/thoas/go-funk"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/gobwas/glob"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	"github.com/spf13/viper"
+)
+
+const defaultInterfacesFilter = "veth*"
+
+var (
+	log   = logf.Log.WithName("configurations")
+	mutex = sync.Mutex{}
+)
+
+// Struct to match config file
+type configParams struct {
+	InterfaceFilter      string `mapstructure:"interfaces_filter"`
+	InterfacesFilterGlob glob.Glob
+}
+
+type config struct {
+	configPath   string
+	v            *viper.Viper
+	configParams configParams
+	initialized  bool
+}
+
+func newConfig() (*config, error) {
+	configPathTemp, isSet := os.LookupEnv("CONFIG_PATH")
+	if !isSet {
+		return nil, fmt.Errorf("CONFIG_PATH is mandatory")
+	}
+
+	configDir, _ := filepath.Split(configPathTemp)
+	if !pathExists(configDir) {
+		return nil, fmt.Errorf("folder %s doesn't exist, can't start configuration watcher", configDir)
+	}
+
+	ext := filepath.Ext(configPathTemp)
+	if !funk.Contains(viper.SupportedExts, ext[1:]) {
+		return nil, fmt.Errorf("file extension %s is not supported", ext)
+	}
+
+	c := config{
+		configPath:  configPathTemp,
+		initialized: true,
+		configParams: configParams{
+			InterfaceFilter:      defaultInterfacesFilter,
+			InterfacesFilterGlob: glob.MustCompile(defaultInterfacesFilter),
+		},
+	}
+
+	c.v = viper.New()
+	c.v.SetConfigFile(c.configPath)
+	c.v.SetTypeByDefaultValue(true)
+
+	return &c, nil
+}
+
+var globalConfig config
+
+func CreateGlobalConfig() error {
+	mutex.Lock()
+	defer mutex.Unlock()
+	if globalConfig.initialized {
+		return nil
+	}
+	conf, err := newConfig()
+	if err != nil {
+		return err
+	}
+	globalConfig = *conf
+	return nil
+}
+
+func GetConfigPath() string {
+	return globalConfig.configPath
+}
+
+func GetCurrentConfig() configParams {
+	mutex.Lock()
+	defer mutex.Unlock()
+	return globalConfig.configParams
+}
+
+func GetInterfacesFilterGlob() glob.Glob {
+	mutex.Lock()
+	defer mutex.Unlock()
+	return globalConfig.configParams.InterfacesFilterGlob
+}
+
+// Updating relevant values with new config settings
+func SetConfig() {
+	if globalConfig.initialized == false {
+		log.Info("Config is not initialized, can't set values")
+		return
+	}
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	if err := globalConfig.v.ReadInConfig(); err != nil { // Find and read the config file
+		log.Info("Not able to read configuration, will update default values")
+	}
+
+	log.Info("Updating configuration")
+	err := globalConfig.v.Unmarshal(&globalConfig.configParams)
+
+	if err != nil {
+		log.Error(err, "Failed to unmarshal config file, skipping configuration update")
+		return
+	}
+
+	globalConfig.configParams.InterfacesFilterGlob = glob.MustCompile(globalConfig.configParams.InterfaceFilter)
+	log.Info(fmt.Sprintf("New configuration is %+v", globalConfig.configParams))
+}
+
+func pathExists(path string) bool {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"os"
 	"os/exec"
 	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -15,7 +16,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	yaml "sigs.k8s.io/yaml"
 
@@ -23,6 +23,7 @@ import (
 
 	"github.com/gobwas/glob"
 	nmstatev1alpha1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1"
+	"github.com/nmstate/kubernetes-nmstate/pkg/configurations"
 	"github.com/nmstate/kubernetes-nmstate/pkg/nmstatectl"
 )
 
@@ -34,18 +35,6 @@ const vlanFilteringCommand = "vlan-filtering"
 const defaultGwRetrieveTimeout = 120 * time.Second
 const defaultGwProbeTimeout = 120 * time.Second
 const apiServerProbeTimeout = 120 * time.Second
-
-var (
-	interfacesFilterGlob glob.Glob
-)
-
-func init() {
-	interfacesFilter, isSet := os.LookupEnv("INTERFACES_FILTER")
-	if !isSet {
-		panic("INTERFACES_FILTER is mandatory")
-	}
-	interfacesFilterGlob = glob.MustCompile(interfacesFilter)
-}
 
 func applyVlanFiltering(bridgeName string, ports []string) (string, error) {
 	command := []string{bridgeName}
@@ -108,8 +97,7 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
 		return errors.Wrap(err, "error running nmstatectl show")
 	}
 	observedState := nmstatev1alpha1.State{Raw: []byte(observedStateRaw)}
-
-	stateToReport, err := filterOut(observedState, interfacesFilterGlob)
+	stateToReport, err := filterOut(observedState, configurations.GetInterfacesFilterGlob())
 	if err != nil {
 		fmt.Printf("failed filtering out interfaces from NodeNetworkState, keeping orignal content, please fix the glob: %v", err)
 		stateToReport = observedState
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -7,6 +7,8 @@ RUN dnf install -b -y dnf-plugins-core && \
     dnf remove -y dnf-plugins-core && \
     dnf clean all
 
+RUN mkdir -p /etc/config
+
 COPY build/_output/bin/kubernetes-nmstate /usr/local/bin/kubernetes-nmstate
 COPY build/bin /usr/local/bin
 
--- a/docs/user-guide-101-reporting.md
+++ b/docs/user-guide-101-reporting.md
@@ -157,15 +157,47 @@ periodically and won't get updated while the node is not reachable (e.g. during
 reconfiguration of networking), this value can be used to evaluate whether the
 observed state is fresh enough.
 
-## Configure refresh interval
+## Reported state
 
 The reported state is updated every 5 seconds.
 
-## Filter reported interfaces
+## Configuration
+
+#### How To
+We support configuration through configmap created from file:
+
+    Create from file:
+        1. Create file `nmstate.yaml` (default name)
+            ----- content --------
+            interfaces_filter: "vnet*"
+            -------------
+        2. Create configuration map from previously created file with name `nmstate-config`:
+            `kubectl -n nmstate create configmap nmstate-config --from-file=nmstate.yaml`
+
+    Create from yaml:
+        1. Create yaml file with content:
+             ------------- content -------------
+             apiVersion: v1
+             kind: ConfigMap
+             metadata:
+               name: nmstate-config
+               namespace: nmstate
+             data:
+               nmstate.yaml: |
+                 interfaces_filter: "veth*"
+             ----------------------------
+        2. Create configuration map
+           `kubectl create -f cm.yaml`
+
+   The interface filtering will be changed to "vnet*".
+
+#### Configuration Options:
+
+##### Filter reported interfaces
 
 By default, all `veth*` interfaces are omitted from the report in order not to
 clutter the output with all Pod connections. This filtering can be adjusted via
-`interfaces_filter` variable in kubernetes-nmstate configmap.
+`interfaces_filter` variable in configuration file.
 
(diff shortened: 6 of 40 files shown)


cmd/manager/main.go
```diff
@@ -93,7 +95,11 @@ func main() {
 		log.Error(err, "Failed to get watch namespace")
 		os.Exit(1)
 	}
-
+	err = configurations.ConfigWatcher()
```
phoracek: Please add empty space above and below this newly added code.

cmd/manager/main.go
```diff
@@ -16,7 +16,9 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/nmstate/kubernetes-nmstate/pkg/apis"
+	"github.com/nmstate/kubernetes-nmstate/pkg/configurations"
 	"github.com/nmstate/kubernetes-nmstate/pkg/controller"
+
```
phoracek: Please drop the empty line. These are all part of this package.

deploy/operator.yaml
```diff
@@ -50,30 +50,34 @@ spec:
...
-              valueFrom:
-                configMapKeyRef:
-                  name: nmstate-config
-                  key: interfaces_filter
             - name: ENABLE_PROFILER
               value: "False"
             - name: PROFILER_PORT
               value: "6060"
+            - name: CONFIG_NAME
+              value: "nmstate-config.yaml"
```
phoracek: nmstate.yaml is enough since it is already under config/. (just a nit)

deploy/operator.yaml
```diff
@@ -50,30 +50,34 @@ spec:
...
           volumeMounts:
           - name: dbus-socket
             mountPath: /run/dbus/system_bus_socket
+          - name: config-volume
+            mountPath: /etc/config
           securityContext:
             privileged: true
+          ports:
```
phoracek: How is this related to the change?

deploy/operator.yaml
```diff
@@ -131,13 +135,15 @@ spec:
             - name: NODE_NETWORK_STATE_REFRESH_INTERVAL
               valueFrom:
                 configMapKeyRef:
+                  optional: true
```
phoracek: You need to mount CM as a volume here too (this is worker DS, the one you changed is the master)

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,147 @@
...
+const defaultNodeNetworkStateRefresh = "5"
+
+var (
+	configPath = defaultConfigPath
+	configName = defaultConfigName
+)
+
+// Take os environment values if given else use defaults
+func init() {
+	configPathTemp, isSet := os.LookupEnv("CONFIG_PATH")
```
phoracek: You put the default for CONFIG_NAME into the manifest, please be consistent and do it with this one too. I think we can drop the defaults from here and just make it required. Require only one parameter - the path to the config, you can get the directory out of it here.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	configNameTemp, isSet := os.LookupEnv("CONFIG_NAME")
+	if isSet {
+		configName = configNameTemp
+	}
+}
+
+// Struct to match config file
+type Config struct {
+	NodeNetworkRefreshInterval string `mapstructure:"node_network_state_refresh_interval"`
```
phoracek: You suffixed the config with .yaml, but here you don't treat it as a yaml. Please let's start with writting docs with examples how to set the config and figure out there what would be the simplest option for users (edit the parameter with a single kubectl command).

pkg/helper/client.go
```diff
@@ -38,16 +38,14 @@ const vlanFilteringCommand = "vlan-filtering"
...
 var (
 	interfacesFilterGlob glob.Glob
 )
 
-func init() {
-	interfacesFilter, isSet := os.LookupEnv("INTERFACES_FILTER")
-	if !isSet {
-		panic("INTERFACES_FILTER is mandatory")
-	}
+func InitClient(interfacesFilter string) {
```
phoracek: Let's be more specific here and call it SetInterfacesFilter? That's all it does.

pkg/helper/config_controller_helper.go
```diff
@@ -0,0 +1,13 @@
+package helper
```
phoracek: Why do we still keep config_controller? Did not we obsolete it?

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -118,6 +114,7 @@ func (r *ReconcileNodeNetworkState) Reconcile(request reconcile.Request) (reconc
 			return err
 		}
 		return nil
+
```
phoracek: Unrelated change, please drop it

pkg/controller/nodenetworkstate/nodenetworkstate_controller.go
```diff
@@ -29,11 +28,8 @@ var (
 	nodenetworkstateRefresh time.Duration
 )
 
-func init() {
-	refreshTime, isSet := os.LookupEnv("NODE_NETWORK_STATE_REFRESH_INTERVAL")
-	if !isSet {
-		panic("NODE_NETWORK_STATE_REFRESH_INTERVAL is mandatory")
-	}
+func InitNodeNetworkController(refreshTime string) {
```
phoracek: Let's call it specifically SetNodeNetworkStateRefreshValue.

pkg/configurations/configurations.go
tsorya: Mapstructure will treat is yaml. Viper will handle it. This is the way they describe using there code.

pkg/helper/config_controller_helper.go
tsorya: We did. Missed it

pkg/configurations/configurations.go
tsorya: Same way we can have json or toml . Just took yaml as default.

phoracek:
@tsorya one other comment. Please make sure to change the value of our global attributes in a sync manner. Wrap it with a Mutex or something.

tsorya:
@phoracek i don't think we need mutex to change simple variable. What do i miss?

phoracek:
I don't think that assignment of a string variable is atomic. If you could reference me a source that says it is, I would be glad to accept it without mutex. But again, it sounds unsafe.

pkg/helper/client.go
```diff
@@ -40,15 +41,15 @@ const defaultGwProbeTimeout = 120 * time.Second
 const apiServerProbeTimeout = 120 * time.Second
 
 var (
+	setMutex             = &sync.Mutex{}
```
phoracek: Can we move out variables to the configurations/ package? Keep it a global, but private. Expose a structure with Synchronous getters and setters to set and read them. Does it sound good? Put all the configuration on a single space instead of sharing it between modules and put it behind a nice abstraction.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+	c.updateConfig(*v)
+	WatchConfig(v, configPath)
+	return nil
+}
+
+//Adapted from viper WatchConfig to match nmstate configmap watch needs
+//The main changes is that there is no need to
+//preexist config file before starting the watch
+//and it will not exit on file deletion
+func WatchConfig(v *viper.Viper, fullFilePath string) {
```
phoracek: This function can be private

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := v.ReadInConfig(); err != nil { // Find and read the config file
+		log.Info("Not able to read configuration, will update default values")
+	}
+
+	c.updateConfig(*v)
+	WatchConfig(v, configPath)
+	return nil
+}
+
+//Adapted from viper WatchConfig to match nmstate configmap watch needs
```
phoracek: Please add space after `//` to all lines

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func ConfigWatcher() error {
```
phoracek: Please add a few lines describing what it does, what it watches, what it changes.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+
```
phoracek: please drop the empty line at the beginning of the function.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+func init() {
+	configPathTemp, isSet := os.LookupEnv("CONFIG_PATH")
+	if !isSet {
+		panic("CONFIG_PATH is mandatory")
+	}
+	configPath = configPathTemp
+}
+
+// Struct to match config file
+type Config struct {
```
phoracek: This can be private. Please add getters (Public) and setters (private) to interface with users and with the watcher. It should be a little cleaner.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+const defaultInterfacesFilter = "veth*"
+const defaultNodeNetworkStateRefresh = 5
+
+var (
+	log        = logf.Log.WithName("configurations")
+	configPath string
+)
+
+// Takes os environment values
+func init() {
```
phoracek: We can get rid of this init (I know, we use them on other places, but it is not a good practice since it randomly panics during imports).

This module should have two structs `Config`, that we already have and `ConfigWatcher`. In main, we should call `NewConfigWatcher`, initializing the structer, that would check the config path. Then we call configWatcher.Start.

Let's try to be clean and safe when dealing with singletons.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+	log.Info("Updating configuration")
+	err := v.Unmarshal(&c)
+	log.Info(fmt.Sprintf("New configuration is %+v", c))
+	if err != nil {
+		log.Error(err, "Failed to unmarshal config file")
+		return
+	}
+}
+
+func exists(path string) bool {
```
phoracek: `exists` is often used in go as a variable. Let's be explicit and call it fileExists

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	v.SetConfigFile(configPath)
+	v.SetTypeByDefaultValue(true)
+
+	if err := v.ReadInConfig(); err != nil { // Find and read the config file
+		log.Info("Not able to read configuration, will update default values")
+	}
+
+	c.updateConfig(*v)
```
phoracek: Why do we need to updateConfig here? Won't it be done by the WatchConfig? This seems redundant

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+	initWG.Add(1)
+	go func() {
+		newWatcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			log.Error(err, "Failed to start fsnotify watcher")
+			return
+		}
+		defer newWatcher.Close()
+
+		configFile := filepath.Clean(fullFilePath)
```
phoracek: This seems like initialization that can be done before we enter the goroutine

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,157 @@
...
+					eventsWG.Done()
+					return
+				}
+			}
+		}()
+		newWatcher.Add(configDir)
+		initWG.Done()   // done initalizing the watch in this go routine, so the parent routine can move on...
+		eventsWG.Wait() // now, wait for event loop to end in this go-routine...
+	}()
+	initWG.Wait() // make sure that the go routine above fully ended before returning
```
phoracek: Why do we need the goroutine above? You just execute it and wait for it to end, can't it be dropped?

pkg/configurations/configurations.go
tsorya: It is the case where file already exists. Watch config will run updateConfig only on change

pkg/configurations/configurations.go
tsorya: nope, the above goroutine is added cause of fsnotify watcher. We need to configure it and be able to run defer on if goroutine is finished. I can do it without closing fsnotify watcher but  it looks like it is not a best practice

pkg/configurations/configurations.go
tsorya: why does it matter?

phoracek: Would make it easier to read. Goroutines add complexity, less they contain, the less I can think about what is happening in parallel. `go` tells me, there will be something running in parallel, I ask "what can go wrong" "what could it confict with"...

pkg/configurations/configurations.go
phoracek: I'm sorry, I don't see it.  If we keep everything, but drop `go func() {` on line 115, what does it change?

pkg/helper/client.go
```diff
@@ -96,7 +85,7 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1alpha1.
 		return fmt.Errorf("error running nmstatectl show: %v", err)
 	}
 	observedState := nmstatev1alpha1.State{Raw: []byte(observedStateRaw)}
-
+	interfacesFilterGlob := glob.MustCompile(configurations.GetCurrentConfig().InterfaceFilter)
```
phoracek: We can compile it once in before we call the config setter.

pkg/configurations/configurations.go
phoracek: You would be able to drop the initWG too. Then you can keep the defer, it would have exactly the same scope as it has now.

tsorya: As I checked this is the way to work with fsnotify. U open routine for ents loop and wait for it.

pkg/helper/client.go
tsorya: U want me to add it to configurations? I think it will be a little strange but it's ok

pkg/configurations/configurations.go
phoracek: Example from fsnotify docs: https://godoc.org/github.com/fsnotify/fsnotify#example-NewWatcher

You indeed run a goroutine, I have nothing against the inner one (line 129). The outer one is the issue.

pkg/helper/client.go
phoracek: Meh, no need to.

tsorya: No it's ok. This function is called A lot. Not sure what is the cost of creating glob but sounds reasonable

pkg/configurations/configurations.go
tsorya: But I need to run the watcher and continue on main thread without closing it. So I create go goroutine and wait in it for inner one.
I think I am missing something. But I don't see how it will work without it.

pkg/helper/client.go
tsorya: So let's decide :)

phoracek: What makes more sense for you. I don't mind neither.

tsorya: The question is the cost of glob creation

pkg/configurations/configurations.go
phoracek: You do initWG.Wait(), so you wait until the outer one is done. So it runs the goroutine and then it  blocks forever. You can drop it and wait for eventsWG, would do the same thing. Or maybe I need more sleep!

pkg/helper/client.go
phoracek: Huuuumph, let's do it in configurations then. So there is no regressions (now we call it only once, let's keep that)

phoracek:
@tsorya I managed to set a wrong secret for the prow webhook. We should fix it tomorrow.

qinqon:
@tsorya @phoracek can we use the client-go Watch on configMap so we get notified when they change ? 
https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/configmap.go#L97

tsorya:
@qinqon but then it will create controller no? I was told not to create controller.
Actually my first version was with controller

phoracek:
@qinqon I'm against having another controller just for the configmap. It would watch all the CMs in the cluster and require us to run the even watcher. With the current approach, we let Kubernetes update the config actively, we just employ an fsWatcher which is way cheaper and simpler than controller mechanism.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+	realConfigFile, _ := filepath.EvalSymlinks(c.configPath)
+	newWatcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Error(err, "Failed to start fsnotify watcher")
+		return
+	}
+	newWatcher.Add(configDir)
+
+	go func() {
+		defer newWatcher.Close()
```
phoracek: This should be set right after  the error checking of NewWatcher.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Adapted from viper WatchConfig to match nmstate configmap watch needs
+// The main changes is that there is no need to
+// preexist config file before starting the watch
+// and it will not exit on file deletion
+func (c *configWatcher) watchConfigFile() {
+	configFile := filepath.Clean(c.configPath)
+	configDir, _ := filepath.Split(configFile)
+	realConfigFile, _ := filepath.EvalSymlinks(c.configPath)
+	newWatcher, err := fsnotify.NewWatcher()
```
phoracek: Please add an empty line above, to split path handling and initialization of the watcher.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		log.Error(err, "Failed to start fsnotify watcher")
+		return
+	}
```
phoracek: Please add an empty line below error handling.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+					realConfigFile = currentConfigFile
+					err := c.v.ReadInConfig()
+					if err != nil {
+						log.Error(err, "error reading config file.\n")
+					} else {
+						setConfig(*c.v)
+					}
+				}
+			case err, ok := <-newWatcher.Errors:
+				if ok { // 'Errors' channel is not closed
```
phoracek: The default flow should be in the main body and the exceptional one in `if`. Then a reader can do less hops to conditions if they are just following the positive flow. So in this case:

```
if !ok {
    return
}
log.Error(...)

File: pkg/configurations/configurations.go [hunk 28]
```diff
@@ -0,0 +1,159 @@
...
+				// 1 - if the config file was modified or created
+				// 2 - if the real path to the config file changed (eg: k8s ConfigMap replacement)
+				const writeOrCreateMask = fsnotify.Write | fsnotify.Create
+				if (filepath.Clean(event.Name) == configFile &&
+					event.Op&writeOrCreateMask != 0) ||
+					(currentConfigFile != "" && currentConfigFile != realConfigFile) {
+					realConfigFile = currentConfigFile
+					err := c.v.ReadInConfig()
+					if err != nil {
+						log.Error(err, "error reading config file.\n")
```
phoracek: If err: Log the error and `continue`. Then we can drop the else branch and just handle the positive flow in the main body

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+			case event, ok := <-newWatcher.Events:
+				if !ok { // 'Events' channel is closed
+					return
+				}
+				currentConfigFile, _ := filepath.EvalSymlinks(c.configPath)
+				// we only care about the config file with the following cases:
+				// 1 - if the config file was modified or created
+				// 2 - if the real path to the config file changed (eg: k8s ConfigMap replacement)
+				const writeOrCreateMask = fsnotify.Write | fsnotify.Create
+				if (filepath.Clean(event.Name) == configFile &&
```
phoracek: Please assign all the conditions into named variabls. e.g. eventFileMatches := filepath.Clean .... == configFile, etc. Then the `if` will be way easier to read. I would read "if X Y Z, set the config".

deploy/operator.yaml
```diff
@@ -50,30 +50,32 @@ spec:
...
             privileged: true
       volumes:
       - name: dbus-socket
         hostPath:
           path: /run/dbus/system_bus_socket
           type: Socket
+      - name: config-volume
+        configMap:
+          # Provide the name of the ConfigMap containing the file
+          # configmap must be created from file it's full path must
```
phoracek: How is the configmap created is not important. The format is. Could you please provide an example of the body instead?

With something like:

```
data:
  nmstate.yaml: |
    key: value
    key: value

File: pkg/configurations/configurations.go [hunk 31]
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/thoas/go-funk"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	"github.com/spf13/viper"
+)
+
+const defaultInterfacesFilter = "veth*"
```
phoracek: We can get rid of it and handle it in the initialization of `config`. Just add a comment that we are initializing with default values.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+	v          *viper.Viper
+}
+
+func NewConfigWatcher() *configWatcher {
+	configPathTemp, isSet := os.LookupEnv("CONFIG_PATH")
+	if !isSet {
+		panic("CONFIG_PATH is mandatory")
+	}
+	cw := configWatcher{
+		configPath: configPathTemp,
```
phoracek: Might as well just set `v` here.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	cw.v = viper.New()
+	return &cw
+}
+
+func GetCurrentConfig() config {
+	return c
```
phoracek: We need the mutex for reading.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// Struct to match config file
+type config struct {
+	NodeNetworkRefreshInterval int    `mapstructure:"node_network_state_refresh_interval"`
+	InterfaceFilter            string `mapstructure:"interfaces_filter"`
+	InterfacesFilterGlob       glob.Glob
+}
+
+var c = config{
```
phoracek: I'm not a big fan of coupling watcher and config. Responsibilities are not so clear here. What about following:

* Split watcher and configurations to two separate files in this package
* Extend config with getters and setters. Maybe one setter to set the whole thing and one getter per each value. All these will employ a mutex.
* Both watcher and user (readers of loaded values) will use these synchronous/safe getters and setters
* With this, the user facing logic (config alone) will be split from the logic of updating it.

tsorya: But config setter is loading yaml to struct. How do u see and I need viper for it. It will look really strange but I'll try.

pkg/configurations/configurations.go
tsorya: Why we need mutex for reading? Reading is not atomic?

pkg/configurations/configurations.go
phoracek: Viper has nothing to do with the fact that reload is dynamic, so it can be part of the config struct. Config can have loadFromYaml/File method

pkg/configurations/configurations.go
tsorya: Then I will need 2 go routines

pkg/configurations/configurations.go
phoracek: Not while somebody is writting to it. Afaik go has no guarantees on atomicity of string reading/writting. If you find a trustworthy source saying otherwise, please let me know.

pkg/configurations/configurations.go
phoracek: Oh now I get it. Sorry. Don't mind the comment.

pkg/configurations/configurations.go
```diff
@@ -0,0 +1,159 @@
...
+func (c *configWatcher) watchConfigFile() {
+	configFile := filepath.Clean(c.configPath)
+	configDir, _ := filepath.Split(configFile)
+	realConfigFile, _ := filepath.EvalSymlinks(c.configPath)
+	newWatcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Error(err, "Failed to start fsnotify watcher")
+		return
+	}
+	newWatcher.Add(configDir)
```
phoracek: Can this fail?

deploy/operator.yaml
tsorya: It is important. If it is not created from file. It will not work.

pkg/configurations/configurations.go
tsorya: but why not to have defaults as consts?

pkg/configurations/configurations.go
tsorya: hmm need to think about it. 
now i am doing it using viper "err := v.Unmarshal(&c)". It is easy and very easily extendable and u ask me to read yaml.
I really don't think there is any need to be so generic and decoupled.

pkg/configurations/configurations.go
phoracek: It's unnecessary duplication. Unless we want to reinitialize the config. Mm, maybe it's just a matter of style. I leave it to you, since I don't really have a convincing explanation.

pkg/configurations/configurations.go
phoracek: I care about the interface for the rest of the application and decoupling of the daemon which reads the current implementation of the config (ConfigMap as a file) from the config object.

deploy/operator.yaml
phoracek: I'm pretty sure user can just create a config map which would be a prepared manifest with populated nmstate.yaml key. I haven't tried it yet, but I believe that --from-file is just a helper in kubectl to do exactly that.

phoracek: Could you dump ConfigMap with -o yaml for me?

pkg/configurations/configurations.go
tsorya: It can only if fsnotify watcher is closed. In this case we control it so it doesn't need to be handled as for me

pkg/configurations/configurations.go
tsorya: got it

deploy/operator.yaml
tsorya: maybe you are right i will check.

pkg/apis/nmstate/v1alpha1/zz_generated.openapi.go
```diff
@@ -11,15 +11,15 @@ import (
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
 	return map[string]common.OpenAPIDefinition{
-		"./pkg/apis/nmstate/v1alpha1.Condition":                               schema_pkg_apis_nmstate_v1alpha1_Condition(ref),
```
tsorya: remove this file

deploy/operator.yaml
tsorya: Checked your right

qinqon:
We have remove nmstate-config since it consumes watchers per handler and at a huge cluster this hits apiserver. Use env vars at handler is good enough, other options will be configured by the operator CRD.

PR #417, opened 2020-02-14
explicitly bring unavailable interfaces up

phoracek:

There is a bug in Kernel/NetworkManager on systems with NetworkManager
1.20, where sometimes after disconnecting a NIC from a bonding, the NIC
remains in 'unavailable' state and cannot be used for a new connection. This
is likely caused by an issue with autonegotiation where the NIC appears to
be disconnected and the only thing that can bring it available again is
explicitly calling `ip link set <name> up` on it. In order to workaround
this issue until it gets solved, we iterate all devices during `nmstatectl
set` and if we find some with 'unavailable' we explicitly set them up.

**Release note**:

```release-note
Workaround 'unavailable' ifaces by setting them UP explicitly
```


pkg/helper/client.go
```diff
@@ -185,10 +186,13 @@ func ApplyDesiredState(desiredState nmstatev1alpha1.State) (string, error) {
 		return "Ignoring empty desired state", nil
 	}
 
+	var setDoneCh = make(chan struct{})
+	go setUnavailableUp(setDoneCh)
```
qinqon: Can't this be done at fix-unavailable.go that implements Set and just run when needed ? (like checking we are at bonding and we are at kernel <= foobar)

pkg/helper/client.go
```diff
@@ -242,6 +246,63 @@ func ApplyDesiredState(desiredState nmstatev1alpha1.State) (string, error) {
...
+		select {
+		case <-stopCh:
+			break Loop
+		default:
+			time.Sleep(time.Second)
+		}
+	}
+}
+
+func ipLinkUp(iface string) error {
```
qinqon: This is not returning ipLinkUp so let's call it `setIpLinkUp`

pkg/helper/client.go
```diff
@@ -242,6 +246,63 @@ func ApplyDesiredState(desiredState nmstatev1alpha1.State) (string, error) {
...
+	var stdout, stderr bytes.Buffer
+
+	cmd := exec.Command("ip", "link", "set", iface, "up")
```
qinqon: We cannot use pnm library here right ? that's the main issue.

pkg/helper/client.go
```diff
@@ -242,6 +246,63 @@ func ApplyDesiredState(desiredState nmstatev1alpha1.State) (string, error) {
...
 
+// There is a bug in Kernel/NetworkManager on systems with NetworkManager
+// 1.20, where sometimes after disconnecting a NIC from a bonding, the NIC
+// remains in 'unavailable' state and cannot be used for a new connection. This
+// is likely caused by an issue with autonegotiation where the NIC appears to
+// be disconnected and the only thing that can bring it available again is
+// explicitly calling `ip link set <name> up` on it. In order to workaround
+// this issue until it gets solved, we iterate all devices during `nmstatectl
+// set` and if we find some with 'unavailable' we explicitly set them up.
+func setUnavailableUp(stopCh chan struct{}) {
```
qinqon: Let's implement it with https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L486 so we don't re-invento the wheel.

pkg/helper/client.go
phoracek: We don't know what exactly causes the issue. I managed to reproduce it on two different systems. I would prefer to run it all the time until we get to support a new enough platforms without this bug (maybe even 1.22, this however needs to get backported).

pkg/helper/client.go
```diff
@@ -185,10 +186,13 @@ func ApplyDesiredState(desiredState nmstatev1alpha1.State) (string, error) {
 		return "Ignoring empty desired state", nil
 	}
 
+	var setDoneCh = make(chan struct{})
+	go setUnavailableUp(setDoneCh)
 	// commit timeout doubles the default gw ping probe timeout, to
 	// ensure the Checkpoint is alive before rolling it back
 	// https://nmstate.github.io/cli_guide#manual-transaction-control
 	setOutput, err := nmstatectl.Set(desiredState, defaultGwProbeTimeout*2)
+	close(setDoneCh)
```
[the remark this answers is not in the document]
qinqon: We cannot defer it right ?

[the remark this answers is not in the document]
phoracek: We could but then the scope would be the whole function. Which should not be an issue in theory, but I prefer to keep it explicitly small scope.

pkg/helper/unavailable_link_workaround.go
```diff
@@ -0,0 +1,66 @@
...
+	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout
+
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("ip link set up failed, rc: %v, stdout: %v, stderr: %v", err, stdout.String(), stderr.String())
+	}
+
+	return nil
+}
```
qinqon: Can we create at Set(... function here that delegates on = nmstatectl.Set(, so we don't expose channels outside ? and we keep inpact on client.go minimal ?

pkg/apis/nmstate/v1alpha1/zz_generated.openapi.go
```diff
@@ -220,14 +220,14 @@ func schema_pkg_apis_nmstate_v1alpha1_NodeNetworkConfigurationPolicySpec(ref com
...
-							Ref:         ref("./pkg/apis/nmstate/v1alpha1.State"),
+							Ref:         ref("github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.State"),
 						},
 					},
 				},
 			},
 		},
 		Dependencies: []string{
-			"./pkg/apis/nmstate/v1alpha1.State"},
+			"github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1alpha1.State"},
```
qinqon: bad flags ?

pkg/nmstatectl/nmstatectl.go
```diff
@@ -57,7 +57,12 @@ func Show(arguments ...string) (string, error) {
 }
 
 func Set(desiredState nmstatev1alpha1.State, timeout time.Duration) (string, error) {
-	return nmstatectlWithInput([]string{"set", "--no-commit", "--timeout", strconv.Itoa(int(timeout.Seconds()))}, string(desiredState.Raw))
+	var setDoneCh = make(chan struct{})
```
qinqon: I mean to SetWithUnavailabeLinkWorkround(... at unavailable_link_workaround.go and there
do the thing and at the end nmstatectl.Set

and at client.go just replace nmstatectl.Set for SetWIthUnabl... so client.go is just one liner change and all the integration happends a new file we can easily delete.

phoracek: The bug is only related to nmstatectl execution, so I moved it here.

I don't see how introducing wrappers is easier to revert than three short lines. Both will be just `git revert` anyway.

qinqon: fair enough

PR #419, opened 2020-02-17
remove user_setup

phoracek:

AFAIK there is no need for such a setup. It should work with root out
of the box.

**Release note**:

```release-note
NONE
```


qinqon:
/lgtm
/approve

This is a special privileged container that is designed to change host so we good here with root I suppose.

PR #420, opened 2020-02-17
Change webhook port to 54874

qinqon:
We have found that 8443 is a conflicting port [1] so we change to 54874 here.

[1]: https://github.com/openshift/cluster-kube-apiserver-operator/blob/f5e8c4e9a81c5f5d7477c535d8c7e55af29321c0/manifests/0000_20_kube-apiserver-operator_06_deployment.yaml

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -60,6 +60,8 @@ spec:
                 configMapKeyRef:
                   name: nmstate-config
                   key: interfaces_filter
+            - name: WEBHOOK_PORT
+              value: "54874"
             - name: ENABLE_PROFILER
               value: "False"
             - name: PROFILER_PORT
@@ -173,7 +175,7 @@ spec:
   publishNotReadyAddresses: true
   ports:
     - port: 443
-      targetPort: 8443
+      targetPort: 54874
   selector:
     name: nmstate-handler
 ---
--- a/pkg/webhook/nodenetworkconfigurationpolicy/server.go
+++ b/pkg/webhook/nodenetworkconfigurationpolicy/server.go
@@ -1,6 +1,12 @@
 package nodenetworkconfigurationpolicy
 
 import (
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/pkg/errors"
+
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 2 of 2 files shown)


phoracek:
> Well-known ports range from 0 through 1023. Registered ports are 1024 to 49151. Dynamic ports (also called private ports) are 49152 to 65535.

Please move it to the private range. Ideally to something that is completely random, not even including 8443.

deploy/operator.yaml
```diff
@@ -161,6 +166,7 @@ metadata:
 data:
   node_network_state_refresh_interval: "5"
   interfaces_filter: "veth*"
+  webhook_port: "10443"
```
phoracek: This is something coupled with deployment, not with actual configuration of the application. The port should be setup by the person who sets up the cluster and knows what ports can collide. nmstate configuration contains stuff that adjusts behavior of the app and is important to the user.

I think we should keep it solely in the DaemonSet definition.

deploy/operator.yaml
```diff
@@ -161,6 +166,7 @@ metadata:
 data:
   node_network_state_refresh_interval: "5"
   interfaces_filter: "veth*"
+  webhook_port: "54874"
```
phoracek: This is something coupled with deployment, not with actual configuration of the application. The port should be setup by the person who sets up the cluster and knows what ports can collide. nmstate configuration contains stuff that adjusts behavior of the app and is important to the user.

I think we should keep it solely in the DaemonSet definition.

pkg/webhook/nodenetworkconfigurationpolicy/server.go
```diff
@@ -13,6 +19,17 @@ const (
...
 
+	webhookPortString, isSet := os.LookupEnv("WEBHOOK_PORT")
+	if !isSet || len(webhookPortString) == 0 {
+		return fmt.Errorf("WEBHOOK_PORT env var is mandatory")
+	}
+
+	var err error
+	webhookPort, err := strconv.Atoi(webhookPortString)
+	if err != nil {
+		errors.Wrap(err, "WEBHOOK_PORT env var has bad format")
```
phoracek: misses return

qinqon:

Locally it passes and prow is stuck, let's override so we merge.

PR #427, opened 2020-02-18
Add dockerignore to project

qinqon:
It will prevent copying the kubevirtci dir and golang installations into
docker build context when we build the handler

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,32 @@
+# We cannot call `docker build` out of top dir since this is controlled by
+# operator-sdk and it does not has the option to change that, but future
+# versin of operator-sdk is going to split the build process into small
+# makefile targets and we will be able to tune that, so this file will
+# be smaller
+
+# git stuff
+.git
+.github
+
+# The path for kubevirtci installation
+_kubevirtci
+
+# The old path for kubevirtci installation, in case checkout is done from and
+# old branch this can exists
+kubevirtci
+
+# No need for source code is already compiled
+pkg
+cmd
+vendor
+docs
+
+# Test and infra
+test
+test_logs
+automation
+
+
+# go tools
+build/**/go*
+build/**/go


.dockerignore
```diff
@@ -0,0 +1,2 @@
+kubevirtci
+*/**/go*
```
phoracek: This would prevent cloning of any random module that starts with "go". e.g. gotohell.go. Can we lock it more specifically?

.dockerignore
```diff
@@ -0,0 +1,3 @@
+kubevirtci
```
phoracek, line 1: Is not this _kubevirtci?

qinqon: ack.

phoracek: Why do we still have kubevirtci here? https://github.com/nmstate/kubernetes-nmstate/blob/master/cluster/kubevirtci.sh#L5

phoracek:
/lgtm
/approve
/override pull-kubernetes-nmstate-e2e-handler-k8s
/hold

Not related. Feel free to cancel the hold if you are satisfied with the PR.

.dockerignore
```diff
@@ -0,0 +1,26 @@
...
+.github
+
+# The path for kubevirtci installation
+_kubevirtci
+
+# The old path for kubevirtci installation, in case checkout is done from and
+# old branch this can exists
+kubevirtci
+
+# No need for source code is already compiled
```
phoracek, line 12: If you don't care for so many directories, maybe you can just selectively `COPY` exactly what you need instead of copying the whole project and then blacklisting here?

qinqon: Issue is not about  `COPY` issue is about where do you run `docker build [dir]` and that's done by `operator-sdk` for example for handler we don't need the manifests so doing `docker build build/ -t ...` would be enough, but that's not the case for operator that need the manifests

phoracek: Then please add a huge comment on the top of this file saying it is only for operator-sdk due to this and that, so we can drop it once the situation with operator-sdk gets better.

PR #428, opened 2020-02-18
Removing nodestate_controller and moving its logic to node controller

tsorya:
Signed-off-by: Igal Tsoiref <itsoiref@redhat.com>

No need to have controller for read only crd.

Some tests maybe broken will fix if it looks ok

**Release note**:

```release-note
NONE
```


phoracek:
CC @qinqon

tsorya:
Node controller reconcile every five seconds, so yes it will be updated every 5 seconds. 
The problem is that now there is no need for configuration of refresh rate. 
I can add some time validation in node controller but not sure it is needed. If it is running every 5 seconds let it run or we can move this configuration here

phoracek:
@qinqon what do you think? Do we want to give up on refresh interval or investigate more how to keep it? I don't mind neither.

qinqon:
@tsorya  @phoracek This is nice we remove one controller and one config param let's go with defaults (aka 5 seconds) we can always configure it somehow if needed, also I don't see change at documentation.

pkg/controller/node/node_controller.go
```diff
@@ -76,10 +77,14 @@ var _ reconcile.Reconciler = &ReconcileNode{}
...
-	client client.Client
-	scheme *runtime.Scheme
+	client         client.Client
+	scheme         *runtime.Scheme
+	nmstateUpdater NmstateUpdater
 }
 
+// Added for test purposes
+type NmstateUpdater func(client client.Client, node *corev1.Node, namespace client.ObjectKey) error
+
```
qinqon: This will add some complexity only for unit test, but alternative are faking nmstatectl with at bash script and using FakeClient, but this also add a lot of complexity to unit test.

tsorya: after discussion it was decide to leave it this way for this PR

PR #429, opened 2020-02-19
Use centos8, fedora:32 is problematic.

qinqon:
We couldn't use fedora:31 since nm 1.22 was missing there and it's a dependency for
nmstate-0.2, but we can isntall it from copr but for that we need to use centos8 image, also
is looks like f31 + 1.22 is no-no.

Also the following is implemented here:
- Also add the `-b` option is passed to `dnf install` to ensure that docker build fail in case of rpm install is skipped.
- Use fix a bug at client.go to use probe api timeout to at nmstatectl set command 
- Check client.go probes after rollback to wait for thinkgs to settle up.
- Disable dhcp at secondary nics at tests.
- Fix test to match specific bonding options (broken after nmstate upgrade)

**Release note**:

```release-note
 NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ WHAT ?= ./pkg
 
 unit_test_args ?=  -r -keepGoing --randomizeAllSpecs --randomizeSuites --race --trace $(UNIT_TEST_ARGS)
 
-export KUBEVIRT_PROVIDER ?= k8s-1.17.0
+export KUBEVIRT_PROVIDER ?= k8s-1.17
 export KUBEVIRT_NUM_NODES ?= 1
 export KUBEVIRT_NUM_SECONDARY_NICS ?= 2
 
@@ -138,7 +138,6 @@ $(CLUSTER_DIR)/%: $(install_kubevirtci)
 	$(install_kubevirtci)
 
 cluster-prepare:
-	hack/install-ovs.sh
 	hack/install-nm.sh
 	hack/flush-secondary-nics.sh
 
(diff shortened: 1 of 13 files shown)


phoracek:
FIx the description please

qinqon:
@phoracek description is good

phoracek:
> Use fedora:31 at handler container end with nmstate skipped but
> the container build does not fail, with "-b" we instruct dnf
> to use latest packages or fail otherwise.

qinqon:
@phoracek done

qinqon:
default bond error is deterministic
`failed to execute nmstatectl commit: 'exit status 65' 'ERROR committing change: No checkpoint specified or found`

qinqon:
Found that we are not taking into account the API connectivity timeout to configure nmstatectl timeout... so we have being very lucky.

Also now with centos8 something related to runtime-controller is not right 
```
Could not wait for Cache to sync	{"controller": "node-controller", "error": "failed to wait for node-controller caches to sync"
```

Our cluster sync check is not good enough.

qinqon:
Also found that in case of lossing cluster connectivity, it's n ot able to progress policy state :-(

qinqon:
/override 
pull-kubernetes-nmstate-e2e-k8s

qinqon:
/test 
pull-kubernetes-nmstate-e2e-k8s

phoracek:
@qinqon please describe all the changes in the PR description, explain what was wrong and how did you fix it.

phoracek:
@qinqon please describe all the changes in the PR description, explain what was wrong and how did you fix it.

qinqon:
@phoracek done
@phoracek done
@phoracek done

phoracek:
@qinqon do we need this or was everything merged as a part of other PRs?

PR #443, opened 2020-03-03
Use centos8

qinqon:
We couldn't use fedora:31 since nm 1.22 was missing there and it's a dependency for
nmstate-0.2, but we can install it from copr but for that we need to use centos8 image, also
is looks like f31 + 1.22 is no-no. 

Also add the -b option is passed to dnf install to ensure that docker build fail in case of rpm install is skipped.

e2e tests are still broken since k8s nodes and knmstate container are very different, I will fix that at follow up PRs, this just fix make handler so we can build it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,6 +1,7 @@
-FROM fedora:32
+FROM centos:8
 
 RUN dnf install -b -y dnf-plugins-core && \
+    dnf copr enable -y networkmanager/NetworkManager-1.22 && \
     dnf copr enable -y nmstate/nmstate-0.2 && \
     dnf install -b -y nmstate iproute iputils && \
     dnf remove -y dnf-plugins-core && \
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,8 +1,8 @@
 FROM fedora:32
 
-RUN dnf install -y dnf-plugins-core && \
+RUN dnf install -b -y dnf-plugins-core && \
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM fedora:32
+FROM centos:8
 
-RUN dnf install -y dnf-plugins-core && \
+RUN dnf install -b -y dnf-plugins-core && \
+    dnf copr enable -y networkmanager/NetworkManager-1.22 && \
     dnf copr enable -y nmstate/nmstate-0.2 && \
-    dnf install -y nmstate iproute iputils && \
+    dnf install -b -y nmstate iproute iputils && \
```
oshoval: -b, --best            try the best available package versions in
                        transactions.

b fails the command if it fails to install ?

phoracek: Try the best available package versions in transactions. Specifically during dnf upgrade, which by default skips over updates that can not be installed for dependency reasons, the switch forces DNF to only consider the latest packages. **When running into packages with broken dependencies, DNF will fail giving a reason why the latest version can not be installed.**

oshoval: thanks, wasnt at the --help but i see its on
https://dnf.readthedocs.io/en/latest/command_ref.html

qinqon:
/override 
pull-kubernetes-nmstate-e2e-k8s

PR #455, opened 2020-03-18
Use release notes

qinqon:
/kind enhancement

knmstate project follows the kubevirt/kubernetes PRs template, it has a field for `release-notes` those notes compose the description of a release.

This PR add those release notes to the description, future approach would be to add those release notes at postsubmit.

Now you need GITHUB_TOKEN to call `make prepare-`

Support for labels has being activated at prow https://github.com/kubevirt/project-infra/pull/403

Generated `version/description` example

```markdown
v0.18.0

TODO: Add description here

# Changelog since v0.17.0

## Changes by Kind

### Test

 - Test: verify that bridge IP is persisted after reboot ([#456](https://github.com/nmstate/kubernetes-nmstate/pull/456), [@phoracek](https://github.com/phoracek))
 - The e2e tests now run at workers and skipping master. ([#483](https://github.com/nmstate/kubernetes-nmstate/pull/483), [@qinqon](https://github.com/qinqon))

### Feature

 - New operator for kubernetes-nmstate deployment. ([#461](https://github.com/nmstate/kubernetes-nmstate/pull/461), [@bcrochet](https://github.com/bcrochet))

### Other (Bug, Cleanup or Flake)

 - After desiredState is apply handler will wait for node to be ready in case the config change that. ([#480](https://github.com/nmstate/kubernetes-nmstate/pull/480), [@qinqon](https://github.com/qinqon))
 - Bug: post rollback probes were not being executed. ([#484](https://github.com/nmstate/kubernetes-nmstate/pull/484), [@qinqon](https://github.com/qinqon))

```
docker pull HANDLER_IMAGE
```
```

**Release note**:

```release-note
action requiered: Set GITHUB_TOKEN before call `make prepare-`
```


Makefile
```diff
@@ -137,12 +141,12 @@ $(description): version/description
...
-prepare-minor:
-	./hack/prepare-release.sh minor
-prepare-major:
-	./hack/prepare-release.sh major
+prepare-patch: $(RELEASE_NOTES)
+	export PATH=$(GOBIN):$$PATH && ./hack/prepare-release.sh patch
+prepare-minor: $(RELEASE_NOTES)
+	export PATH=$(GOBIN):$$PATH && ./hack/prepare-release.sh minor
+prepare-major: $(RELEASE_NOTES)
+	export PATH=$(GOBIN):$$PATH && ./hack/prepare-release.sh major
```
phoracek: Please pass path to the binary as a standalone variable. The same way we do it with GITHUB_RELEASE in the release target.

hack/prepare-release.sh
```diff
@@ -1,15 +1,38 @@
...
+end() {
+    rm $release_notes
+}
+
+trap end EXIT SIGINT SIGTERM SIGSTOP
 
+release-notes \
+    --format go-template:$script_dir/release-notes.tmpl \
+    --release-version $new_version \
+    --required-author kubevirt-bot \
```
phoracek: What does this stand for? Would not it filter only PRs opened by the bot?

hack/release-notes.tmpl
```diff
@@ -0,0 +1,18 @@
...
+{{end}}
+
+{{- with .NotesByKind -}}
+## Changes by Kind
+{{ range $kind, $notes = .}}
+### {{$kind | prettyKind}}
+{{range $note = $notes}} {{println "-" $note}} {{end -}}
+{{end -}}
+{{end -}}
+
```
phoracek: Please drop the trailing line from the template and handle it while writting to the `description`

hack/prepare-release.sh
```diff
@@ -1,15 +1,38 @@
...
 cat << EOF > version/description
 $new_version
 
 TODO: Add description here
 
+EOF
+
+cat $release_notes >> version/description
+
+cat << EOF >> version/description
```
phoracek: I think we can drop the following list of Features Bugs and Docs. The automatic release note collection just requires us to label PRs properly and then we can have the same list autogenerated.

hack/release-notes.tmpl
```diff
@@ -0,0 +1,18 @@
+# Changelog since {{ .PreviousRevision }}
+
+{{with .NotesWithActionRequired -}}
```
phoracek: How do we use this?

hack/release-notes.tmpl
```diff
@@ -0,0 +1,18 @@
...
+## Urgent Upgrade Notes
+
+### (No, really, you MUST read this before you upgrade)
+
+{{range .}} {{println "-" .}} {{end}}
+{{end}}
+
+{{- with .NotesByKind -}}
```
phoracek: This requires us to set kind on every PR properly. Would you please adjust the PR template to do that? To have a required preset with /kind feature, /kind bug (as done in KubeVirt) and ask the person opening the PR to select one.

Makefile
qinqon: Maybe what we want is to pass GOBIN at top of Makefile so all of our scripts just use the executable and they don't depend on external variables and can be run locally if proper tools are there, what do you think ?

hack/prepare-release.sh
qinqon: Ahh right, we need empty string there https://github.com/kubernetes/release/blob/master/cmd/release-notes/main.go#L159

qinqon: Ahh no at the end at master author is kubevirt-bot since it's squashing the PRs.

hack/release-notes.tmpl
qinqon: These are the steps:
1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
2. Follow the instructions for writing a release note from k8s: https://git.k8s.io/community/contributors/guide/release-notes.md
3. If no release note is required, just write "NONE".

hack/release-notes.tmpl
qinqon: Where do we have that doc ?

Makefile
qinqon: Nah let's do as you say and do a follow up with PATH.

hack/release-notes.tmpl
```diff
@@ -0,0 +1,20 @@
+# Changelog since {{ .PreviousRevision }}
```
phoracek, line 1: I think that this can remain implicit. Changelog is always related to the lust bump, isn't it.

hack/release-notes.tmpl
```diff
@@ -0,0 +1,20 @@
...
+
+{{with .NotesWithActionRequired -}}
+## Urgent Upgrade Notes
+
+### (No, really, you MUST read this before you upgrade)
```
phoracek, line 6: Redundant

hack/release-notes.tmpl
```diff
@@ -0,0 +1,20 @@
+# Changelog since {{ .PreviousRevision }}
+
+{{with .NotesWithActionRequired -}}
+## Urgent Upgrade Notes
```
phoracek, line 4: This can be h1 (#) if we drop the Changelog since header

hack/release-notes.tmpl
```diff
@@ -0,0 +1,20 @@
...
+
+### (No, really, you MUST read this before you upgrade)
+
+{{range .}} {{println "-" .}} {{end}}
+{{end}}
+
+{{- with .Notes -}}
+## Changes by Kind
```
phoracek, line 12: Redundant, we can directly lust Kinds

hack/release-notes.tmpl
qinqon: Why redundant? this are the notes that are marced with `action requiered` this is important.

qinqon: Ahh you mean the phrase.

phoracek: Ok, but it is not a header for sure. If you want to emphasize it, use `**Note:**` or something. Let's not be the guys who are using huge red text to get attention

phoracek: Also, the title above it is already saying URGENT. It somebody is dumb enough to ignore it, let them suffer

PR #463, opened 2020-03-23
Pin nmstate to 0.1

qinqon:
We were installing nmstate-0.2 but this depends on NetworkManager 1.22 and we are not suppose to use that on release-0.15

It was not failing because when dnf install is run with multiple package if one is skipped it does not fail and do a fallback to best approach.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,7 +1,7 @@
 FROM fedora:31
 
 RUN sudo dnf install -y dnf-plugins-core && \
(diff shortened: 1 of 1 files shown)


qinqon:
/cc @phoracek 

Looks we have a false possitive at release-0.15

qinqon:
/retest

it's failing at same test always...

PR #453, opened 2020-03-17
[release-0.15] Annotate tests with polarion IDs

phoracek:
**What this PR does / why we need it**:

In order to be able to track test results with Polarion, we need to add these annotations.

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #465, opened 2020-03-24
Adapt to external k8s

qinqon:

This PR add needed changes to test kubernetes-nmstate e2e tests at an external cluster, with kubernetes-nmstate being deployed as part of other deployments like HCO.

We have found a pair of bugs running this at external system:

- Calculate policy conditions counting nodes running nmstate pod.: - https://github.com/nmstate/kubernetes-nmstate/issues/473
- Race condition between rollback checks and Node state: https://github.com/nmstate/kubernetes-nmstate/issues/474
- Bad log at policy condition success: https://github.com/nmstate/kubernetes-nmstate/issues/475
- app label missing at daemonset : https://github.com/nmstate/kubernetes-nmstate/issues/476

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/label-workers.sh
+++ b/hack/label-workers.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -e
+
+script_dir=$(dirname "$(readlink -f "$0")")
+kubectl=$script_dir/../kubevirtci/cluster-up/kubectl.sh
+number_of_nodes=$($kubectl get node  --no-headers |wc -l)
+
+if [ $number_of_nodes -eq 1 ]; then
+    label="node-role.kubernetes.io/master"
+else
+    label="!node-role.kubernetes.io/master"
+fi
+
+$kubectl label node -l $label node-role.kubernetes.io/worker=''
--- a/test/e2e/openshift.cnv.workflow.sh
+++ b/test/e2e/openshift.cnv.workflow.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -xe
+
+# Configure kubeconfig
+export KUBECONFIG=${KUBECONFIG:-$HOME/oc4/working/auth/kubeconfig}
+export KUBECTL=${KUBECTL:-oc}
+export NAMESPACE=${NAMESPACE:-openshift-cnv}
+export SSH=${SSH:-./ssh.sh}
+export PRIMARY_NIC=${PRIMARY_NIC:-ens3}
+export FIRST_SECONDARY_NIC=${FIRST_SECONDARY_NIC:-ens7}
+export SECOND_SECONDARY_NIC=${SECOND_SECONDARY_NIC:-ens8}
+export TIMEOUT=${TIMEOUT:-60m}
+
+if [ ! -f  $SSH ]; then
+    cat << EOF > ${SSH}
+#!/bin/bash
+node_name=\${1}
+node_ip=\$($KUBECTL get node \${node_name} --no-headers -o wide | awk '{print \$6}')
+ssh core@\${node_ip} -- \${@:3}
+EOF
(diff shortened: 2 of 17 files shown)


test/e2e/prepare.go
```diff
@@ -43,6 +45,12 @@ func waitForDaemonSets(t *testing.T, kubeclient kubernetes.Interface, namespace
 			return true, errors.Wrapf(err, "failed retrieving daemon sets for namespace %s", namespace)
 		}
 		for _, daemonset := range daemonsets.Items {
+			// we don't have the app=kubernetes-nmstate labels at daemonsets so we
+			// filter by name
+			if !strings.Contains(daemonset.Name, "nmstate") {
```
phoracek: It is not great, but we can just check the app label in daemonset.Spec.Template. (We should add proper labels on master)

test/runner/pod.go
```diff
@@ -21,6 +21,9 @@ func RunAtPods(arguments ...string) {
 	nmstatePods, err := nmstatePods()
 	ExpectWithOffset(1, err).ToNot(HaveOccurred())
 	for _, nmstatePod := range nmstatePods {
+		if ! strings.Contains(nmstatePod, "worker") {
```
phoracek: Sure you ran gofmt on this?

test/e2e/openshift.cnv.workflow.sh
```diff
@@ -0,0 +1,30 @@
...
+#!/bin/bash
+node_name=\${1}
+node_ip=\$($KUBECTL get node \${node_name} --no-headers -o wide | awk '{print \$6}')
+ssh core@\${node_ip} -- \${@:3}
+EOF
+    chmod +x ${SSH}
+fi
+
+# Run workflow tests
+FOCUS_1='Nodes.*when.*are.*up.*and.*new.*interface.*is.*configured.*should.*update.*node.*network.*state.*with.*it'
```
phoracek: We should backport the polarion integration and replace this by test_id selectors.

hack/label-workers.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+
+script_dir=$(dirname "$(readlink -f "$0")")
+kubectl=$script_dir/../kubevirtci/cluster-up/kubectl.sh
+
+$kubectl label node -l '!node-role.kubernetes.io/master' node-role.kubernetes.io/worker=''
```
phoracek: What about single-node setup? Tests should still work on local env where 1 node is the default.

If there is only single node, let's make it both master and worker?

test/e2e/prepare.go
qinqon: I would keep this like this, browsing Spec.Template add more complicated code, let's add app label at master and we fix this.

https://github.com/nmstate/kubernetes-nmstate/issues/476

test/e2e/openshift.cnv.workflow.sh
qinqon: I am into it with https://github.com/nmstate/kubernetes-nmstate/pull/453 but we have some missing commits from master that stabilize e2e tests https://github.com/nmstate/kubernetes-nmstate/pull/472

hack/label-workers.sh
qinqon: Totally, also for the k8s-1.17 provider (we are not going to use this provider since it's centos8 node based).

test/runner/pod.go
qinqon: Yep, what's the issue ?

phoracek: The space after `!`

qinqon:
Something is fishy at test_id:3793 it's still failing after retest.

PR #466, opened 2020-03-25
Sync NMstate CRDs from Controller

bcrochet:
This add the ability for the operator to create the CRDs from YAML manifests in the image. It does not start the controllers if the operator is running. 

```release-note
Add an operator that will manage the lifecycle of the existing handler. It syncs the CRDs, the namespace (if different than the operator), the RBAC, and the handler itself. 
```


pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -114,3 +121,24 @@ func (r *ReconcileNMstate) syncKubernetesNMStateCRDs(instance *nmstatev1alpha1.N
 	}
 	return nil
 }
+
+func (r *ReconcileNMstate) addControllers(instance *nmstatev1alpha1.NMstate) error {
+	var err error
+
+	cfg, err := config.GetConfig()
```
phoracek: We don't need to use client-go directly here AFAIK. operator-sdk gives us controller-runtime client which is supposed to be used in this framework. https://github.com/operator-framework/operator-sdk/blob/11aa2af7aaa202cb38b2417031e6af2035c7948c/doc/user/client.md

bcrochet: AFAICT that's what's happening here... Look at the imports. Am I missing something?

qinqon:
/hold

@bcrochet why is the bindata at the PR ?

build/Dockerfile
```diff
@@ -9,5 +9,6 @@ RUN dnf install -b -y dnf-plugins-core && \
 
 COPY build/_output/bin/kubernetes-nmstate /usr/local/bin/kubernetes-nmstate
 COPY build/bin /usr/local/bin
+COPY bindata /bindata
```
qinqon: Let's forget about Docker for now, make template accessible for unit test, and let's just work the unit test.
After we are happy with Reconcile logic at unit test level, we start to wire things up.

pkg/controller/nmstate/nmstate_controller.go
phoracek: I'm not sure I understand what you plan to do in this structure. We cannot keep nmstate controllers (the existing ones) running as part of the operator process. The operator should merely deploy manifests. So here you would just render templates for our existing daemon sets and apply them.

The operator should be decoupled from the rest of the processes by this "create manifest" interface. So we can later simply split operator from the rest if needed.

cluster/clean.sh
```diff
@@ -15,6 +15,7 @@ $kubectl delete --ignore-not-found -f $MANIFESTS_DIR/operator.yaml
 $kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nodenetworkconfigurationenactments_crd.yaml
 $kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
 $kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nodenetworkstates_crd.yaml
+$kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nmstates_crd.yaml
```
qinqon: We are skipping the nmstate controller, let's not create/delete the CRDs for now.

cluster/sync.sh
```diff
@@ -78,6 +78,7 @@ function deploy() {
     $kubectl apply -f $MANIFESTS_DIR/service_account.yaml
     $kubectl apply -f $MANIFESTS_DIR/role.yaml
     $kubectl apply -f $MANIFESTS_DIR/role_binding.yaml
+    $kubectl apply -f $MANIFESTS_DIR/nmstate.io_nmstates_crd.yaml
```
qinqon: ditto

deploy/role.yaml
```diff
@@ -69,6 +69,12 @@ rules:
   - '*'
   verbs:
   - '*'
+- apiGroups:
```
qinqon: why do we need this ?

go.mod
```diff
@@ -3,21 +3,26 @@ module github.com/nmstate/kubernetes-nmstate
...
 	github.com/aktau/github-release v0.7.2
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/go-logr/logr v0.1.0
 	github.com/go-openapi/spec v0.19.4
 	github.com/gobwas/glob v0.2.3
 	github.com/kelseyhightower/envconfig v1.4.0
-	github.com/onsi/ginkgo v1.10.1
-	github.com/onsi/gomega v1.7.0
+	github.com/onsi/ginkgo v1.10.3
+	github.com/onsi/gomega v1.7.1
```
qinqon: This is unrelated to the test, can you justify why do we need to bump ginkgo ?, it add a lot of noise to the review, we should do bumping at different PR, to check results and easy review.

pkg/apply/apply.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ApplyObject applies the desired object against the apiserver,
+// merging it with any existing objects if already present.
+func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstructured) error {
```
qinqon: Can we programatically call Cobra cmds from https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go so we re-use the code here ? 

This apply add a lot of complexity to the project that we have to maintain.

pkg/apply/merge.go
```diff
@@ -0,0 +1,178 @@
+package apply
```
qinqon: ditto, let's investigate if we can re-use code from https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go or similar.

pkg/controller/add_node.go
```diff
@@ -1,10 +1,14 @@
...
 import (
+	"os"
+
 	"github.com/nmstate/kubernetes-nmstate/pkg/controller/node"
 )
 
 func init() {
 	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
-	AddToManagerFuncs = append(AddToManagerFuncs, node.Add)
+	if _, runOperator := os.LookupEnv("RUN_OPERATOR"); !runOperator {
```
qinqon: we don't need this here, is only for nmstate controller.

pkg/controller/add_nodenetworkconfigurationpolicy.go
```diff
@@ -1,10 +1,14 @@
...
 import (
+	"os"
+
 	"github.com/nmstate/kubernetes-nmstate/pkg/controller/nodenetworkconfigurationpolicy"
 )
 
 func init() {
 	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
-	AddToManagerFuncs = append(AddToManagerFuncs, nodenetworkconfigurationpolicy.Add)
+	if _, runOperator := os.LookupEnv("RUN_OPERATOR"); !runOperator {
```
qinqon: ditto

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -81,6 +86,31 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	objs := []*uns.Unstructured{}
+
+	data := render.MakeRenderData()
+	objs, err = render.RenderDir(filepath.Join("/bindata", "kubernetes-nmstate", "crds"), &data)
+	if err != nil {
+		return fmt.Errorf("failed to render kubernetes-nmstate crd: %v", err)
+	}
+
+	for _, obj := range objs {
+		if err = controllerutil.SetControllerReference(instance, obj, r.scheme); err != nil {
```
qinqon: Add a comment explaining why we need this.

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -81,6 +86,31 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			return fmt.Errorf("failed to set owner reference: %v", err)
+		}
+		if err = apply.ApplyObject(context.TODO(), r.client, obj); err != nil {
```
qinqon: As say let's investigate if we can do this using a library so we don't have to maintain apply/merge code.

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -81,6 +86,31 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
 }
+
+func (r *ReconcileNMstate) syncKubernetesNMStateCRDs(instance *nmstatev1alpha1.NMstate) error {
+	var err error
+	objs := []*uns.Unstructured{}
+
+	data := render.MakeRenderData()
+	objs, err = render.RenderDir(filepath.Join("/bindata", "kubernetes-nmstate", "crds"), &data)
+	if err != nil {
+		return fmt.Errorf("failed to render kubernetes-nmstate crd: %v", err)
```
qinqon: Please use errors.Wrap/Wrapf

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -81,6 +86,31 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+
+	for _, obj := range objs {
+		if err = controllerutil.SetControllerReference(instance, obj, r.scheme); err != nil {
+			return fmt.Errorf("failed to set owner reference: %v", err)
```
qinqon: Please use errors.Wrap/Wrapf

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -81,6 +86,31 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		}
+		if err = apply.ApplyObject(context.TODO(), r.client, obj); err != nil {
+			return fmt.Errorf("failed to apply object %v with err: %v", obj, err)
```
qinqon: Please use errors.Wrap/Wrapf

pkg/render/render.go
```diff
@@ -0,0 +1,108 @@
...
+		Data:  map[string]interface{}{},
+	}
+}
+
+// RenderDir will render all manifests in a directory, descending in to subdirectories
+// It will perform template substitutions based on the data supplied by the RenderData
+func RenderDir(manifestDir string, d *RenderData) ([]*unstructured.Unstructured, error) {
+	out := []*unstructured.Unstructured{}
+
+	if err := filepath.Walk(manifestDir, func(path string, info os.FileInfo, err error) error {
```
qinqon: Isn't it better to use https://golang.org/pkg/html/template/#Template.ParseGlob, so it does the walk for us ? we can use (*/*.yaml) to parse subdirs although we need to run one pre level (*.yaml), (*/*.yaml).

pkg/render/render.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			return err
```
qinqon: add some specific error with errors.Wrap,  so it's easy to debug.

pkg/render/render.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		// Skip non-manifest files
+		if !(strings.HasSuffix(path, ".yml") || strings.HasSuffix(path, ".yaml") || strings.HasSuffix(path, ".json")) {
+			return nil
+		}
+
+		objs, err := RenderTemplate(path, d)
+		if err != nil {
+			return err
```
qinqon: ditto

pkg/render/render.go
```diff
@@ -0,0 +1,108 @@
...
+	if err := filepath.Walk(manifestDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		// Skip non-manifest files
+		if !(strings.HasSuffix(path, ".yml") || strings.HasSuffix(path, ".yaml") || strings.HasSuffix(path, ".json")) {
```
qinqon: Maybe better here to use an array of suffix and check on a loop with a flag, other option what about using a regex ?

pkg/render/render.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		objs, err := RenderTemplate(path, d)
+		if err != nil {
+			return err
+		}
+		out = append(out, objs...)
+		return nil
+	}); err != nil {
+		return nil, errors.Wrap(err, "error rendering manifests")
```
qinqon: using word error here is redundant, can we use "failed rendering manifests" ?

pkg/render/render.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return out, nil
+}
+
+// RenderTemplate reads, renders, and attempts to parse a yaml or
+// json file representing one or more k8s api objects
+func RenderTemplate(path string, d *RenderData) ([]*unstructured.Unstructured, error) {
```
qinqon: Do we really need the Funcs ? at render-manifests.go it was enough with ParseGlob, and those templates are not going to change, but maybe you know some tunning needed at manifests for openshift.

pkg/controller/nmstate/nmstate_controller.go
qinqon: I think we are good, the PR also checks RUN_OPERATOR add the non operator controllers so they are not run, so we can run the kubernetes-nmstate-handler in two modes, running only the operator controller and running only the non-operator controllers.

That's right @bcrochet

phoracek: But with this code, we will run nmstate controllers as a part of the operator, that is wrong. The operator should just apply manifests with operands.

qinqon: Ahh wait, that's right, why do we do this ?

pkg/apply/apply.go
```diff
@@ -0,0 +1,65 @@
+package apply
```
yboaron: @bcrochet  the plan to reuse the code from CNO and to drop this directory, right?

yboaron:
IIUC, it's decided to use a single image for both operators and operand cases and control controller addition by ENV variable.
Don't we need to add node & nodenetconfig controllers only if the environment variable is not set?

build/Dockerfile.operator
```diff
@@ -0,0 +1,14 @@
+FROM centos:8
+
+RUN dnf install -b -y dnf-plugins-core && \
```
phoracek: None of these are needed. Let's just use minimal container and our built binary.

build/Dockerfile.operator
```diff
@@ -0,0 +1,14 @@
+FROM centos:8
```
phoracek: Please use ubi8: registry.access.redhat.com/ubi8/ubi-minimal

build/Dockerfile.operator
phoracek: Either let's build tiny separate image for the operator or just keep the binary as a part of the already existing image. It may help us get some basics in. We can split it later (that will be probably done on OpenShift anyway)

bindata/kubernetes-nmstate/crds/nmstate.io_nodenetworkconfigurationenactments_crd.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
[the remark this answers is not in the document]
bcrochet: I've separated them to set up the mechanism for finding them in a set location when the image is built. If we don't want to do this quite yet, I can be more explicit in the testing Dockerfile on the copy, and just copy them from either the build directory or from deploy directly.

bcrochet:
> IIUC, it's decided to use a single image for both operators and operand cases and control controller addition by ENV variable.
> Don't we need to add node & nodenetconfig controllers only if the environment variable is not set?

That should already be there.

build/Dockerfile.operator
bcrochet: On second thought... This is only here as testing so as not to pollute the main Dockerfile. Since we are sharing the image, it would need to have everything that the handler needs.

bcrochet:
> Ok, so after some brainstorming since nmstate-handler (aka operand) has big dependencies (network manager and this prevent us from using ubi8 image) we need a slim small operastor container, but we still want to stick with operator-sdk way of making things, so let's make this this way:
> 
> We create all the stuff we needed by calling the `handler` Makefile target, so we use operator-sdk tools, but for operator we just do `docker build` and copy needed stuff from builded handler.
> 
> So something like this at the Makefile:
> 
> ```makefile
> operator: handler
>     docker build . -f build/Dockerfile.operator
> ```
> 
> we will need `push-operator` too, let's tag with same version as handler for now.
> 
> Apart from that, let's just go for the gold and forget about Reconcile unit test and try to use the operator at the project to deploy the operand, that's the best test we can do.

I like how you think. That will be my next revision.

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,49 @@
...
+            - name: WATCH_NAMESPACE
+              value: ""
+            - name: OPERATOR_NAME
+              value: "kubernetes-nmstate-operator"
+            - name: ENABLE_PROFILER
+              value: "False"
+            - name: PROFILER_PORT
+              value: "6060"
+            - name: RUN_OPERATOR
+              value: ""
```
yboaron: I guess "True" is more suitable in that case, right?

hack/render-manifests.go
```diff
@@ -22,7 +22,7 @@ func main() {
 	}
 
 	if len(os.Args) < 6 {
```
yboaron: You add another parameter [handler/operator] , don't you need to update if statement? os.Args < 7 ?

hack/render-manifests.go
```diff
@@ -22,7 +22,7 @@ func main() {
...
-		fmt.Fprintf(os.Stderr, "usage: go run render-manifests.go [namespace] [image] [pull-policy] [input-dir] [output-dir]\n")
+		fmt.Fprintf(os.Stderr, "usage: go run render-manifests.go [namespace] [image] [pull-policy] [input-dir] [output-dir] [handler|operator]\n")
```
yboaron: Maybe you can use <> for mandatory parameter and [ ] for the last optional parameter?

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,119 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
 
+	err = r.syncKubernetesNMStateCRDs(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = r.syncKubernetesNMStateRBAC(instance)
```
yboaron: What is our error handling policy?  should we take care to rollback 'cleanup' ? 
Assuming the following use case:
1.  syncKubernetesNMStateCRDs - succeeded
2.  syncKubernetesNMStateRBAC - failed, so we exit with error. 
So next time reconcile function will be called, I assume that :
syncKubernetesNMStateCRDs  might fail because CRD already exists.

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,49 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
```
yboaron: Maybe I'm missing something here, could you please elaborate why we need to run the operator as DaemonSet and not as Deployment?

deploy/operator/operator.yaml
qinqon: The usual way of do this flag things at pods is just adding the var without value, so we don't care about the value, at golang we can check that var is set or not and that's enough, no need to parse a string value (numerical boolean ? etc...).

pkg/controller/nmstate/nmstate_controller.go
qinqon: I suppose rollback is not part of this PR, we have to really think about it, this PR is like a MVP kind of thing.

deploy/operator/operator.yaml
qinqon: kubernetes-nmstate is distributed, we did it by design, reason is we have to reduce the number of remote access so we can do node's networking rollback if we loss kubeapi connectivity after a bad network configuration.

pkg/controller/nmstate/nmstate_controller.go
phoracek: Typically, there is no rollback in operators. We only need to make sure that if we fail in the second step, we retry. And of course, during the retry we should not fail on the fact that stuff from step 1 is already deployed in the cluster.

deploy/operator/operator.yaml
phoracek: But Yossi is talking about the operator, not handler. With handler, we should deloy as a Deployment. In fact we have to, CSV does not support DaemonSets.

qinqon: Ohh my mistake, yep operator has to be a Daemon and operand at DaemonSet.

Makefile
```diff
@@ -100,12 +104,18 @@ gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
 manifests:
-	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR)
+	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(OPERATOR_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR) handler
+operator-manifests:
```
qinqon, line 108: We need just one "manifests:" since we are going to manually render manifest for operator the handler ones are done by the operator itself.

Makefile
```diff
@@ -100,12 +104,18 @@ gen-crds: $(OPERATOR_SDK)
...
+	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(OPERATOR_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR) operator
 
 handler: gen-openapi gen-k8s gen-crds $(OPERATOR_SDK) manifests
 	$(OPERATOR_SDK) build $(HANDLER_IMAGE) --image-builder $(IMAGE_BUILDER)
 push-handler: handler
 	$(IMAGE_BUILDER) push $(HANDLER_IMAGE)
+operator: handler operator-manifests
+	$(IMAGE_BUILDER) build --build-arg=BASE_IMAGE=$(HANDLER_IMAGE) -t $(OPERATOR_IMAGE) -f build/Dockerfile.operator .
```
qinqon, line 116: Kudos on using multistage!

deploy/operator/handler.yaml
```diff
@@ -0,0 +1,210 @@
+---
```
qinqon, line 1: I think we have to split manifests
deploy/operator/...
deploy/operand/...

For example operator and operand RBAC or namespace can be different.

hack/render-manifests.go
```diff
@@ -16,23 +16,29 @@ func exitWithError(err error, cause string, args ...interface{}) {
 
 func main() {
 	type Inventory struct {
```
qinqon, line 18: The render-manifests.go will be used only for operator now, since operator is going to be in charge of render operand.

Makefile
```diff
@@ -7,6 +7,10 @@ HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
 HANDLER_IMAGE_SUFFIX ?=
 HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME)$(HANDLER_IMAGE_SUFFIX)
 HANDLER_IMAGE ?= $(IMAGE_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)
+OPERATOR_IMAGE_NAME ?= kubernetes-nmstate-operator
+OPERATOR_IMAGE_SUFFIX ?=
+OPERATOR_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(OPERATOR_IMAGE_NAME)$(OPERATOR_IMAGE_SUFFIX)
+OPERATOR_IMAGE ?= $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE_FULL_NAME)
 
 NAMESPACE ?= nmstate
```
qinqon, line 15: We have to split handler and operator manifests I suppose
HANDLER_NAMESPACE, OPERATOR_NAMESPACE ?

Makefile
```diff
@@ -100,12 +104,16 @@ gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
 manifests:
-	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR)
+	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(OPERATOR_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR)
```
qinqon, line 107: pass both namespaces

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,50 @@
...
+            - name: PROFILER_PORT
+              value: "6060"
+            - name: RUN_OPERATOR
+              value: ""
+            - name: HANDLER_IMAGE
+              value: {{ .NMStateHandlerImage }}
+            - name: HANDLER_IMAGE_PULL_POLICY
+              value: {{ .ImagePullPolicy }}
+            - name: HANDLER_NAMESPACE
+              value: {{ .Namespace }}
```
qinqon, line 50: We should pass handler namespace here

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,50 @@
...
+            - name: ENABLE_PROFILER
+              value: "False"
+            - name: PROFILER_PORT
+              value: "6060"
+            - name: RUN_OPERATOR
+              value: ""
+            - name: HANDLER_IMAGE
+              value: {{ .NMStateHandlerImage }}
+            - name: HANDLER_IMAGE_PULL_POLICY
+              value: {{ .ImagePullPolicy }}
```
qinqon, line 48: Should operator PullPolicy be different than handler ?

hack/render-manifests.go
```diff
@@ -45,14 +47,23 @@ func main() {
...
 	if err != nil {
-		exitWithError(err, "failed parsing top dir manifests at %s", inputDir)
+		exitWithError(err, "failed parsing top dir operator manifests at %s", inputDir)
 	}
 
-	tmpl, err = tmpl.ParseGlob(path.Join(inputDir, "*/*.yaml"))
+	// Be explicit about which subdirs we render. Otherwise, we might inadvertently override
+	// a manifest with the same name. Once the operator is in place, we shouldn't need
+	// to do this anymore.
+	tmpl, err = tmpl.ParseGlob(path.Join(inputDir, "crds/*.yaml"))
```
qinqon, line 59: Don't we have to render just operator's crds ?

hack/render-manifests.go
```diff
@@ -45,14 +47,23 @@ func main() {
...
 	if err != nil {
-		exitWithError(err, "failed parsing sub dir manifests at %s", inputDir)
+		exitWithError(err, "failed parsing sub dir crds manifests at %s", inputDir)
+	}
+
+	tmpl, err = tmpl.ParseGlob(path.Join(inputDir, "openshift/*.yaml"))
```
qinqon, line 64: we need an handler/openshift/ and operator/openshift

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -16,7 +26,8 @@ import (
 )
 
 var (
-	log = logf.Log.WithName("controller_nmstate")
+	log         = logf.Log.WithName("controller_nmstate")
+	controllers = []func(manager.Manager) error{node.Add, nodenetworkconfigurationpolicy.Add}
```
qinqon, line 30: This is not needed.

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = r.syncKubernetesNMStateRBAC(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err = r.applyDaemonset(instance); err != nil {
```
qinqon, line 109: nit: treat errors like the code around.
```golang
err =
if err
```

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		return reconcile.Result{}, err
+	}
+
 	reqLogger.Info("Reconcile called")
 	return reconcile.Result{}, nil
 }
+
+func (r *ReconcileNMstate) syncKubernetesNMStateCRDs(instance *nmstatev1alpha1.NMstate) error {
```
qinqon, line 117: applyCRDs

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	data := render.MakeRenderData()
+	return r.renderAndApply(instance, data, "crds", true)
+}
+
+func (r *ReconcileNMstate) syncKubernetesNMStateNamespace(instance *nmstatev1alpha1.NMstate) error {
```
qinqon, line 122: applyNamespace

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	data := render.MakeRenderData()
+	data.Data["Namespace"] = os.Getenv("HANDLER_NAMESPACE")
+	return r.renderAndApply(instance, data, "namespace", false)
+}
+
+func (r *ReconcileNMstate) syncKubernetesNMStateRBAC(instance *nmstatev1alpha1.NMstate) error {
```
qinqon, line 128: applyRBAC

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	data := render.MakeRenderData()
+	data.Data["Namespace"] = os.Getenv("HANDLER_NAMESPACE")
+	data.Data["NMStateHandlerImage"] = os.Getenv("HANDLER_IMAGE")
+	data.Data["ImagePullPolicy"] = os.Getenv("HANDLER_IMAGE_PULL_POLICY")
+	return r.renderAndApply(instance, data, "rbac", true)
+}
+
+func (r *ReconcileNMstate) applyDaemonset(instance *nmstatev1alpha1.NMstate) error {
```
qinqon, line 136: applyHandler, since it's not only a daemonset what we are applying there.

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -49,10 +49,12 @@ var (
 )
 
 func init() {
-	var isSet = false
-	nodeName, isSet = os.LookupEnv("NODE_NAME")
-	if !isSet || len(nodeName) == 0 {
-		panic("NODE_NAME is mandatory")
+	if _, runOperator := os.LookupEnv("RUN_OPERATOR"); !runOperator {
```
qinqon, line 52: Let's remove a nest level by calling `return` if RUN_OPERATOR is set

pkg/helper/client.go
```diff
@@ -40,11 +40,13 @@ var (
 )
 
 func init() {
-	interfacesFilter, isSet := os.LookupEnv("INTERFACES_FILTER")
-	if !isSet {
-		panic("INTERFACES_FILTER is mandatory")
+	if _, runOperator := os.LookupEnv("RUN_OPERATOR"); !runOperator {
```
qinqon, line 43: ditto

pkg/names/names.go
```diff
@@ -0,0 +1,7 @@
+package names
+
+// Namespace is the namespace where resources are created
+const Namespace = "nmstate"
```
qinqon, line 4: Where do we use this ?

hack/render-manifests.go
bcrochet: Not sure we have to render CRDs at all, tbh. Are we doing this just to copy them? And we should probably add the handler namespace, pull policy, and image to the spec.

pkg/names/names.go
bcrochet: I don't think we do. I'll remove it.

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,50 @@
...
+            - name: OPERATOR_NAME
+              value: "kubernetes-nmstate-operator"
+            - name: ENABLE_PROFILER
+              value: "False"
+            - name: PROFILER_PORT
+              value: "6060"
+            - name: RUN_OPERATOR
+              value: ""
+            - name: HANDLER_IMAGE
+              value: {{ .NMStateHandlerImage }}
```
qinqon, line 46: Can we make it `HandlerImage` for consistency ( I have do same comment on prefix PR )

deploy/operator/role.yaml
```diff
@@ -0,0 +1,96 @@
+---
```
qinqon, line 1: There is a lot of stuffhere  that is not needed for the operator like the certificate part, I would say that we just be reactivate here and add after stuff is failing so we have minimal herre, at the end is important to reduce the attack surface of an operator.

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,50 @@
...
+  namespace: {{ .OperatorNamespace }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: nmstate-operator
+  template:
+    metadata:
+      labels:
+        app: kubernetes-nmstate
```
qinqon, line 15: let's have two `app` labels for knmstate

```yaml
app: kubernetes-nmstate (operand)
app: kubernetes-nmstate-operator (operator)
```
Since they serve to very different purpose, this way for example we can dump cluster wide logs with

```bash
# cluster wide operand logs
kubectl logs -n openshift-nmstate -l kubernetes-nmstate

# cluster wide operator logs
kubectl logs -n openshift-nmstate -l kubernetes-nmstate-operator
```

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,50 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nmstate-operator
+  namespace: {{ .OperatorNamespace }}
```
qinqon, line 6: can we add a label to Deployment too ? 
```yaml
app: kubernetes-nmstate-operator
```

hack/render-manifests.go
```diff
@@ -14,25 +14,44 @@ func exitWithError(err error, cause string, args ...interface{}) {
 	os.Exit(1)
 }
 
+const NUM_ARGS = 9
+
+const (
+	HANDLER_NAMESPACE = iota + 1
```
qinqon, line 20: I have commented the same at prefix PR, Let's use naming args (knowing correct position for 7 args is problematic).

hack/render-manifests.go
qinqon: You right, CRDs are generated let's just remove this line from here and copy them from deploy/crds directly into the containers.

pkg/controller/add_node.go
qinqon, line 11: Let's do it like in the policy controller

```golang
if _, runOperator := os.LookupEnv("RUN_OPERATOR"); runOperator {
    return
}
```

pkg/controller/add_nodenetworkconfigurationpolicy.go
qinqon, line 11: ditto

pkg/controller/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -49,6 +49,9 @@ var (
 )
 
 func init() {
+	if _, runOperator := os.LookupEnv("RUN_OPERATOR"); runOperator {
```
qinqon, line 52: Can we encapsulate this in a fuction like `isOperator()` we have multiple places.

qinqon:
@bcrochet can you rebase ?

cluster/sync.sh
```diff
@@ -13,6 +13,15 @@ function getNumberAvailable {
     echo ${numberAvailable:-0}
 }
 
+function getOperatorDesiredNumberScheduled {
```
qinqon, line 16: We don't need this for `Deployment`

cluster/sync.sh
```diff
@@ -13,6 +13,15 @@ function getNumberAvailable {
...
+    $kubectl get deployment -n nmstate $1 -o=jsonpath='{.status.replicas}'
+}
+
+function getOperatorNumberAvailable {
```
qinqon, line 20: ditto

cluster/sync.sh
```diff
@@ -78,14 +93,16 @@ function deploy() {
...
-    $kubectl apply -f $MANIFESTS_DIR/nmstate.io_nodenetworkconfigurationenactments_crd.yaml
+    $kubectl apply -f deploy/crds/nmstate.io_nmstates_crd.yaml
     $kubectl apply -f $MANIFESTS_DIR/operator.yaml
 }
 
+function deploy() {
+    $kubectl apply -f deploy/crds/nmstate.io_v1alpha1_nmstate_cr.yaml
+}
+
 function wait_ready() {
```
qinqon, line 104: Let's rename with handler now, `wait_ready_handler`

cluster/sync.sh
```diff
@@ -78,14 +93,16 @@ function deploy() {
...
-    # Wait until the handler becomes consistently ready on all nodes
+    # Wait until the operator becomes consistently ready on all nodes
     for ds in nmstate-handler nmstate-handler-worker; do
         # We have to re-check desired number, sometimes takes some time to be filled in
         if ! eventually isOk $ds; then
```
qinqon, line 108: `isHandlerOk`

cluster/sync.sh
```diff
@@ -101,5 +118,24 @@ function wait_ready() {
     done
 }
 
+function wait_ready_operator() {
+    # Wait until the operator becomes consistently ready on all nodes
+    for dep in nmstate-operator; do
```
qinqon, line 123: Let's remove this for loop operator has only one kind of pod

cluster/sync.sh
```diff
@@ -101,5 +118,24 @@ function wait_ready() {
...
+        if ! consistently isOperatorOk $dep; then
+            echo "Deployment $dep is not consistently ready within the given timeout"
+            exit 1
+        fi
+    done
+}
+
+deploy_operator
+wait_ready_operator
 deploy
```
qinqon, line 140: This is suppose to be done by operator we have to remove it (at the end is what we are testing).

deploy/operator/operator.yaml
```diff
@@ -0,0 +1,52 @@
...
+  labels:
+    app: kubernetes-nmstate
```
qinqon, line 8: `app: kubernetes-nmstate-operator`

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Owned objects are automatically garbage collected. For additional cle  anup logic use finalizers.
 			// Return and don't requeue
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
 
+	err = r.syncKubernetesNMStateCRDs(instance)
```
qinqon, line 94: applyCRDs

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	if err != nil {
+		return reconcile.Result{}, err
```
qinqon, line 96:
```golang
errors.Wrap(err, "failed applying CRDs")
```

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+
+	err = r.applyNamespace(instance)
+	if err != nil {
+		return reconcile.Result{}, err
```
qinqon, line 101: ditto

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+
+	err = r.applyRBAC(instance)
+	if err != nil {
+		return reconcile.Result{}, err
```
qinqon, line 106: ditto

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,90 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+
+	if err = r.applyHandler(instance); err != nil {
+		return reconcile.Result{}, err
```
qinqon, line 110: ditto

pkg/helper/operator.go
```diff
@@ -0,0 +1,9 @@
+package helper
```
qinqon, line 1: Let's not add more to `helper` it's a something that we did wrong at the beginning, let's create a `pkg/environment/` module and put operator.go there, in future we will add the other env vars

It would be use like:

```golang
environment.IsOperator()
```

qinqon:
@bcrochet can you add a release not to the PR ? This is an important PR that need a note on the next release.

Makefile
```diff
@@ -101,12 +107,16 @@ gen-crds: $(OPERATOR_SDK)
...
 
 manifests:
-	$(GO) run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(NAMESPACE) -handler-image=$(HANDLER_IMAGE) -handler-pull-policy=$(PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
+	$(GO) run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(HANDLER_NAMESPACE) -operator-namespace=$(OPERATOR_NAMESPACE) -handler-image=$(HANDLER_IMAGE) -operator-image=$(OPERATOR_IMAGE) -handler-pull-policy=$(HANDLER_PULL_POLICY) -operator-pull-policy=$(OPERATOR_PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
 
 handler: gen-openapi gen-k8s gen-crds $(OPERATOR_SDK) manifests
 	$(OPERATOR_SDK) build $(HANDLER_IMAGE) --image-builder $(IMAGE_BUILDER)
 push-handler: handler
 	$(IMAGE_BUILDER) push $(HANDLER_IMAGE)
+operator: handler manifests
```
qinqon, line 116: We don't need the `manifests` target as dep of `operator` target since we render the `operator` manifests not the `handler` ones, we have the dependency inside the `sync.sh` script.

cluster/clean.sh
```diff
@@ -12,9 +12,10 @@ if [ ! -d $MANIFESTS_DIR ]; then
 fi
 
 $kubectl delete --ignore-not-found -f $MANIFESTS_DIR/operator.yaml
-$kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nodenetworkconfigurationenactments_crd.yaml
-$kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
-$kubectl delete --ignore-not-found -f $MANIFESTS_DIR/nmstate.io_nodenetworkstates_crd.yaml
+$kubectl delete --ignore-not-found -f deploy/crds/nmstate.io_nodenetworkconfigurationenactments_crd.yaml
+$kubectl delete --ignore-not-found -f deploy/crds/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml
+$kubectl delete --ignore-not-found -f deploy/crds/nmstate.io_nodenetworkstates_crd.yaml
```
qinqon, line 17: Isn't NMState CR the owner of the operand CRDs ?

cluster/sync.sh
```diff
@@ -4,11 +4,11 @@ set -ex
 
 kubectl=./cluster/kubectl.sh
 
-function getDesiredNumberScheduled {
+function getDesiredNumberHandlerScheduled {
```
qinqon, line 7: nit: `getDesiredScheduledHandlers`

cluster/sync.sh
```diff
@@ -4,11 +4,11 @@ set -ex
...
     $kubectl get daemonset -n nmstate $1 -o=jsonpath='{.status.desiredNumberScheduled}'
 }
 
-function getNumberAvailable {
+function getNumberHandlerAvailable {
```
qinqon, line 11: nit: `getAvailableHandlers`

cluster/sync.sh
```diff
@@ -62,6 +62,9 @@ function deploy() {
     # Build new handler image from local sources and push it to the kubevirtci cluster
     IMAGE_REGISTRY=${registry} make push-handler
 
+    # Build new operator image from local sources and push it to the kubevirtci cluster
+    IMAGE_REGISTRY=${registry} make push-operator
```
qinqon, line 66: we can do `IMAGE_REGISTRY=${registry} make push-operator push-handler` or even add a target that do both 
`IMAGE_REGISTRY=${registry} make push` at the end we are going to need it too to publish both a quay and the postsubmit prow job.

cmd/manager/main.go
```diff
@@ -92,15 +93,17 @@ func main() {
...
+	if !environment.IsOperator() {
+		// Take exclusive instance lock, we need that to make sure that
+		// we don't have more than one working instance of k8s-nmstate
+		handlerLock, err := lockHandler()
+		if err != nil {
+			log.Error(err, "Failed to run lockHandler")
+			os.Exit(1)
+		}
+		log.Info("Successfully took nmstate exclusive lock")
+		defer handlerLock.Unlock()
```
qinqon, line 105: We have to call `defer` outside of if statement scope if we don't the handler will be unlock after scope.

The simplest solution is to return a pointer to lock that it's nil if it is an operator so we can do the following:

```golang
defer func() { if handlerLock != nil { handlerLock.Unlock() } }
```

Another option is to create a wrapper around the lock that has the `Unlock` function and to the line below so main function will just have `defer lockWrapper.Unlock()`

Other option is to create a `lockHandlerAndRun(mainFunc func() error) error` that lock handler if is not an operator and then run `mainFunc`

pkg/controller/nmstate/nmstate_controller.go
qinqon: @bcrochet we don't need `controllers` right ?

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -71,16 +81,103 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		return errors.Wrapf(err, "failed to render kubernetes-nmstate %s", sourceDirectory)
+	}
+
+	for _, obj := range objs {
+		// RenderDir seems to add an extra null entry to the list. It appears to be because of the
+		// nested templates. This just makes sure we don't try to apply an empty obj.
+		if obj.GetName() == "" {
+			continue
+		}
+		if setControllerReference {
```
qinqon, line 167: As said the owner is the NMState CR not the operator itself, so if we remove the NMState CR the operand is removed, we have also to block delete operation until all the operand resources are deleted, to set ownerRef do something similar to https://github.com/nmstate/kubernetes-nmstate/blob/36ea41d54974d409ad13e00a3520b9e0d1025e91/pkg/helper/client.go#L71-L79

Also you have to set there `BlockOwnerDeletion: true` so when we delete NMState CR with `foreground` cascade mode it's block until all the operands resources are delete.

Also I  think we have to put some params to `$kubectl delete`at clean.sh to specify `foreground` as cascade option but I haven't investigate that yet.

docs: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

Makefile
bcrochet: I would think it would be the other way around , would it not? That it should be a dep of operator, not handler? Since the handler will only be installed by the operator?

pkg/controller/nmstate/nmstate_controller.go
bcrochet: So we do want to use SetControllerReference. That will do what we want.

Makefile
qinqon: But we don't need the manifest to build the operator

cmd/manager/main.go
bcrochet: Not necessary. Defer runs on func boundaries, not containing block boundaries.

Makefile
```diff
@@ -109,22 +115,27 @@ gen-openapi: $(OPENAPI_GEN)
 gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
-manifests: $(GO)
-	$(GO) run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(NAMESPACE) -handler-image=$(HANDLER_IMAGE) -handler-pull-policy=$(PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
+manifests:
```
yboaron, line 118: `manifest` renders only operator yaml files, right?

Makefile
```diff
@@ -109,22 +115,27 @@ gen-openapi: $(OPENAPI_GEN)
...
+	$(GO) run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(HANDLER_NAMESPACE) -operator-namespace=$(OPERATOR_NAMESPACE) -handler-image=$(HANDLER_IMAGE) -operator-image=$(OPERATOR_IMAGE) -handler-pull-policy=$(HANDLER_PULL_POLICY) -operator-pull-policy=$(OPERATOR_PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
 
 handler: gen-openapi gen-k8s gen-crds $(OPERATOR_SDK) manifests
```
yboaron, line 121: Do we need `manifest` in `handler: gen-openapi gen-k8s gen-crds $(OPERATOR_SDK) manifests` ?
Since manifest renders operator YAML I think we don't neet it.

Makefile
```diff
@@ -109,22 +115,27 @@ gen-openapi: $(OPENAPI_GEN)
...
 	$(OPERATOR_SDK) build $(HANDLER_IMAGE) --image-builder $(IMAGE_BUILDER)
 push-handler: handler
 	$(IMAGE_BUILDER) push $(HANDLER_IMAGE)
+operator: handler manifests
```
yboaron, line 125: Don't we need to set `push-handler` as dependency for `operator` ? since operator build is based on the handler image from the registry.

cluster/clean.sh
```diff
@@ -12,9 +12,10 @@ if [ ! -d $MANIFESTS_DIR ]; then
 fi
 
 $kubectl delete --ignore-not-found -f $MANIFESTS_DIR/operator.yaml
```
yboaron, line 14: Don't we need to delete also NMState CRD definition?

cluster/clean.sh
yboaron: @qinqon , that make sense that NMState will be the owner of operand CRDs,  do you expect that in 'real' cluster, deletion of NMState instance means deletions for handler CRDs?
If your answer is yes, I guess NMState controller should be updated.
an another Q, do we assume that NMState is singletone ? (max number of  NMState instances = 1) ?

pkg/controller/nmstate/nmstate_controller.go
yboaron: I'm not sure that I understood our plan regarding create/delete (upgrade) flow,
IIUC, our design is : delete NMState instance --> Delete operand CRDs , who's responsible for deleting handler CRDs instances?

Makefile
qinqon: manifests generate operator manifest not operand also we don't need it to build it.

Makefile
qinqon: we don't need the handler to be pushed before building operator, we only need them be pused before cluster-sync since it's de moment when we apply the manifests.

Makefile
qinqon, line 118: we need to rebase this PR at master we have `manifests: $(GO)`

Makefile
```diff
@@ -109,22 +115,27 @@ gen-openapi: $(OPENAPI_GEN)
...
 
 test/unit: $(GINKGO)
 	INTERFACES_FILTER="" NODE_NAME=node01 $(GINKGO) $(unit_test_args) $(WHAT)
 
 test/e2e: $(OPERATOR_SDK)
 	mkdir -p test_logs/e2e
 	unset GOFLAGS && $(OPERATOR_SDK) test local ./test/e2e \
-		--kubeconfig $(KUBECONFIG) \
-		--namespace $(NAMESPACE) \
+		--kubeconfig $(shell ./cluster/kubeconfig.sh) \
```
qinqon, line 137: rebase needed, we have `--kubeconfig $(KUBECONFIG)` at master.

cluster/sync.sh
```diff
@@ -78,28 +78,36 @@ function deploy() {
...
             echo "Daemon set $ds is not consistently ready within the given timeout"
             exit 1
         fi
     done
 }
 
-deploy
-wait_ready
+function wait_ready_operator() {
+    $kubectl wait deployment -n nmstate -l app=kubernetes-nmstate-operator --for condition=Available --timeout=200s
```
qinqon, line 107: Can we use the `OPERATOR_NAMESPACE`and `HANDLER_NAMESPACE` at this script ?

cluster/sync.sh
```diff
@@ -41,13 +41,13 @@ function consistently {
     done
 }
 
-function isOk {
-    desiredNumberScheduled=$(getDesiredNumberScheduled $1)
-    numberAvailable=$(getNumberAvailable $1)
+function isHandlerOk {
```
qinqon, line 44: Just a note related to next PR after this, checking handler daemonsets has to be done at operator itself after apply and notify it using the `Status` CRD part with some `Conditions` following the k8s standards, so we will just monitor NMState CRD status.

cmd/manager/main.go
qinqon: Ahh right better!

EDIT: some gotchas on `defer` https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

pkg/controller/nmstate/nmstate_controller.go
qinqon: The operand/operator upgrade (they will share the version) it will work like a k8s `Deployment` I think it will work like this:
- Apply new operator manifests with new version(operator and operand the share the version), this will remove old operator and deploy new one done by k8s `Deployment`
- About CRDs they should not be owned by controllers and do the "migration" work at `Reconcile`.

qinqon:
@bcrochet @yboaron there is something fishy at the k8s-1.17 kubevirtci provider will try to fix it this week so we have stable CI, sorry about that...

cmd/manager/main.go
bcrochet: #4 from that same article is what informed me about it. So, is there something in particular that you are concerned about?

qinqon: All good

qinqon:
/retest

The webhook e2e tests race

qinqon:
@bcrochet can you rebase now the e2e tests are stable at master.

pkg/controller/nmstate/nmstate_controller.go
bcrochet: I've made the CRDs not be owned.

phoracek:
@qinqon I think that approval needs to be on the first line of a comment

PR #471, opened 2020-03-27
cleanup examples in documentation

phoracek:

Now when we have 101 tutorial series, examples of various manifests don't need
to be accompanied with their explanations. Therefore, this patch drops those
tutorials (so we don't need to maintain that much text) and replaces them with
mere yaml examples.

**Release note**:

```release-note
NONE
```


phoracek:
/hold

Will add test coverage for each example

test/e2e/utils.go
```diff
@@ -458,3 +459,15 @@ func vlan(node string, iface string) string {
...
+func kubectlAndCheck(command ...string) {
+	out, err := cmd.Kubectl(command...)
+	Expect(err).ShouldNot(HaveOccurred(), out)
+}
+
+func skipIfNotKubernetes() {
+	provider := getEnv("KUBEVIRT_PROVIDER", "k8s")
+	if !strings.Contains(provider, "k8s") {
+		Skip("Tutorials use interface naming that is available only on Kubernetes providers")
+	}
```
qinqon: /lgtm
/approve

PR #477, opened 2020-03-31
Count only nodes knmstate pods at policy conditions calculation

### LINKED ISSUE #473: Calculate policy conditions counting nodes running nmstate pod.
**What happened**:
The policy condition was at progressing state forever since one of the nodes was at bad state and no nmstate pod was running there so node counting was taking it into accout, but enactment was never created.

Whe we apply a policy if will be applied only at pods that are are running nmstate, so we have to take into account nodes that run nmstate pod.

One way to reproduce this is remove the toleration from nmstate daemonpods so they are not deployed at master nodes, after this policy will be forever at progressing state.

**Environment**:

- Others:

qinqon:
…alculation

Closes #473

Sometimes some nodes are NotReady temporally while they are configured or they
are ready but nmstate pod is not running for some reason.

**Release note**:

```release-note
The nncp conditions calculations take now into account the nodes running nmstate pod instead of readiness
```

### DIFF
--- a/pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
+++ b/pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
@@ -85,6 +85,32 @@ func setPolicyFailedToConfigure(conditions *nmstatev1alpha1.ConditionList, messa
 	)
 }
 
+func nodesRunningNmstate(cli client.Client) ([]corev1.Node, error) {
+	nodes := corev1.NodeList{}
+	err := cli.List(context.TODO(), &nodes)
+	if err != nil {
+		return []corev1.Node{}, errors.Wrap(err, "getting nodes failed")
+	}
+
+	pods := corev1.PodList{}
+	byApp := client.MatchingLabels{"app": "kubernetes-nmstate"}
+	err = cli.List(context.TODO(), &pods, byApp)
+	if err != nil {
+		return []corev1.Node{}, errors.Wrap(err, "getting pods failed")
+	}
+
+	filteredNodes := []corev1.Node{}
+	for _, node := range nodes.Items {
+		for _, pod := range pods.Items {
+			if node.Name == pod.Spec.NodeName {
+				filteredNodes = append(filteredNodes, node)
+				break
+			}
+		}
+	}
+	return filteredNodes, nil
+}
+
 func Update(cli client.Client, policyKey types.NamespacedName) error {
 	logger := log.WithValues("policy", policyKey.Name)
 	// On conflict we need to re-retrieve enactments since the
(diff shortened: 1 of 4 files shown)


qinqon:
We need some rbac

```
E0401 12:25:10.427810       1 reflector.go:123] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:204: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:nmstate:nmstate-handler" cannot list resource "pods" in API group "" at the cluster scope
```

qinqon:
/retest
```
+ hostnamectl
+ grep Transient
Waiting for dhclient to set the hostname from dnsmasq
+ echo 'Waiting for dhclient to set the hostname from dnsmasq'
+ sleep 5
+ timeout=0
+ '[' 0 -le 0 ']'
+ exit 1
provi
```

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions.go
```diff
@@ -85,6 +85,31 @@ func setPolicyFailedToConfigure(conditions *nmstatev1alpha1.ConditionList, messa
...
 
+func nodesRunningNmstate(cli client.Client) ([]corev1.Node, error) {
+	nodes := corev1.NodeList{}
+	err := cli.List(context.TODO(), &nodes)
+	if err != nil {
+		return []corev1.Node{}, errors.Wrap(err, "getting nodes failed")
+	}
+
+	pods := corev1.PodList{}
+	err = cli.List(context.TODO(), &pods) // TODO filter nmstate pods
```
phoracek, line 96: What about the todo? Once we add an operator, we would get masters listed here which may not have handlers running.

qinqon: I forgot the filtering... Let's add a unit test for that.

qinqon: Ok addind filter on `app: kubernetes-nmstate` and unit test, also operator should have diffrent app.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions_test.go
```diff
@@ -51,44 +51,67 @@ func p(conditionsSetter func(*nmstatev1alpha1.ConditionList, string), message st
...
 	}
+	return pod
 }
 
-func nodeNotReady() []corev1.NodeCondition {
-	return []corev1.NodeCondition{
-		corev1.NodeCondition{
-			Type:   corev1.NodeReady,
-			Status: corev1.ConditionFalse,
+func newNmstatePodAtNode(idx int) corev1.Pod {
```
phoracek, line 87: We can pass name and app to newPodAtNode and reuse it for both. Let's not get flooded by helper functions.

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions_test.go
```diff
@@ -232,20 +268,25 @@ var _ = Describe("Policy Conditions", func() {
...
-				newNode(3, nodeReady()),
-				newNode(4, nodeNotReady()),
+			Nodes: newNodes(4),
+			Pods: []corev1.Pod{
+				newNmstatePodAtNode(1),
+				newPodAtNode(1),
+				newNmstatePodAtNode(2),
+				newPodAtNode(2),
+				newNmstatePodAtNode(3),
+				newPodAtNode(3),
```
phoracek, line 288: Why do we need so many? Is not one nmstate pod and one non-nmstate enough?

qinqon: To be really sure that we filter nodes with both nmstate and non-nmstate

phoracek: Ack

pkg/controller/nodenetworkconfigurationpolicy/policyconditions/conditions_test.go
qinqon: Well I am going to add another helper function :-P, but going to make a generic one that compose the Pod struct.

qinqon:
/retest

Looks like we have less stable lanes with bigger cluster (kind of make sense)
```

NodeNetworkConfigurationPolicy default bridged network when there is a default interface with dynamic address and linux bridge is configured on top of the default interface should keep the default IP address after node reboot expand_less | 1m28s
-- | --
/tmp/knmstate/kubernetes-nmstate/test/e2e/default_bridged_network_test.go:139 Timed out after 15.000s. Interface brext has not take over the eth0 address Expected     <string>:  to equal     <string>: 192.168.66.102 /tmp/knmstate/kubernetes-nmstate/test/e2e/default_bridged_network_test.go:122


```

qinqon:
@phoracek ci is green now

version/version.go
```diff
@@ -1,7 +1,7 @@
 package version
 
 var (
-	Version = "0.17.0"
+	Version = "0.18.0"
```
phoracek, line 4: This change is unrelated.

PR #480, opened 2020-03-31
Add node ready state probe

### LINKED ISSUE #474: Race condition between rollback checks and Node state
**What happened**:
When applying a policy that deals with primary nic and to test rollback policy was a success state but enacment was progressing and ended with failure (sice a rollback was expected) the issue is related to node counting for policy conditions, since only ready nodes are being counted and playing with primary nic can render one node temporally at NotReady, the comparation between number of nodes and not matching enactments was  passing.

To fix this we have to add another probe after apply and after rollback to check that Node is at Ready state so we block there until node is ok again.

Policy to be a Failure state after a rollback from a bad primary nic change.

Apply a bad primary nic policy at multinic env, it has to be exercise multiple time until race appear.

**Environment**:

- Others:

qinqon:
There is a race condition between rollback and node ready state, since we are counting ready nodes to calculate policy conditions.

This PR add a new probe that check/wait for the node to be ready after desired state is apply and rollback otherwise (and also check back node ready) if is not ready.

Closes #474

It also contains at different commit a little refactoring that moves all the probes we use at `pkg/probe` module.

**Release note**:

```release-note
After desiredState is apply handler will wait for node to be ready in case the config change that.
```

### DIFF
--- a/pkg/probe/probes.go
+++ b/pkg/probe/probes.go
@@ -1,0 +1,123 @@
+package probe
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+	"time"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	yaml "sigs.k8s.io/yaml"
+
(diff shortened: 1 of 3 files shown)


pkg/probe/probes.go
```diff
@@ -73,6 +75,24 @@ func checkApiServerConnectivity(timeout time.Duration) error {
 	})
 }
 
+func checkNodeReadiness(client client.Client, timeout time.Duration) error {
```
phoracek: What if the node was not ready even before we started applying any changes (maybe we applied them to make it ready?). Do you think that may be an issue?

qinqon: This is something that we have to decide and document, then we merge this PR or we don't or maybe we can make this configurable, what do you think ?

phoracek: I think we can just expect the cluster to be Ready before applying anything. If we find out it is an issue for somebody, we can get quorum to discuss it. Sorry, I guess I just wanted to nitpick.

PR #483, opened 2020-04-01
Run e2e test at workers

qinqon:
At release-0.15 branch we are starting to test for external system and running e2e tests only at workers, this PR do this at master so we discover same issues that we have found there.

https://github.com/nmstate/kubernetes-nmstate/pull/465

The rollback test is expected to fail, so we fix it at master on the product not the test.

**Release note**:

```release-note
The e2e tests now run at workers and skipping master.
```

### DIFF
--- a/automation/check-patch.e2e-k8s.sh
+++ b/automation/check-patch.e2e-k8s.sh
@@ -16,7 +16,7 @@ teardown() {
 
 main() {
     export KUBEVIRT_PROVIDER='k8s-1.17'
-    export KUBEVIRT_NUM_NODES=2
+    export KUBEVIRT_NUM_NODES=3 # 1 master, 2 workers
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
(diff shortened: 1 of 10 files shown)


qinqon:
/retest

The cluster-up issue with k8s-1.17  related to transient hostname from dnsmasq

automation/check-patch.e2e-k8s.sh
```diff
@@ -16,7 +16,7 @@ teardown() {
 
 main() {
     export KUBEVIRT_PROVIDER='k8s-1.17'
-    export KUBEVIRT_NUM_NODES=2
+    export KUBEVIRT_NUM_NODES=3 # 1 master, 2 workers
```
phoracek, line 19: Could you please update the developer docs? So it is clear that we need two or three nodes on the cluster. It would be also worth to set it as the default in our cluster/

qinqon: Changed Makefile although we may want to transition kubevirtci vars to cluster/kubevirtci.sh, about docs I don't know where to set this since we don't have a doc about running e2e tests.

PR #487, opened 2020-04-01
Add lock mechanism for handlers instances

yboaron:
In upgrade process we might have more than one working nmstate handler
for some time, to avoid such a case we should use lock mechanism between handlers.

**Release note**:

```release-note
Add an exclusive lock for nmstate-handler instances
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -8,9 +8,12 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/kelseyhightower/envconfig"
+	"github.com/pkg/errors"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
@@ -20,6 +23,7 @@ import (
 	"github.com/nmstate/kubernetes-nmstate/pkg/webhook"
 	"github.com/nmstate/kubernetes-nmstate/version"
 
+	"github.com/nightlyone/lockfile"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
@@ -87,6 +91,16 @@ func main() {
 	}
 
 	printVersion()
+
+	// Take exclusive instance lock, we need that to make sure that
+	// we don't have more than one working instance of k8s-nmstate
+	handlerLock, err := lockHandler()
+	if err != nil {
+		log.Error(err, "Failed to run lockHandler")
+		os.Exit(1)
+	}
+	log.Info("Successfully took nmstate exclusive lock")
+	defer handlerLock.Unlock()
 
 	namespace, err := k8sutil.GetWatchNamespace()
 	if err != nil {
@@ -211,4 +225,25 @@ func setProfiler() {
 			}
 		}()
 	}
+}
+
+func lockHandler() (lockfile.Lockfile, error) {
+	lockFilePath, ok := os.LookupEnv("NMSTATE_INSTANCE_NODE_LOCK_FILE")
+	if !ok {
+		return "", errors.New("Failed to find NMSTATE_INSTANCE_NODE_LOCK_FILE ENV var")
+	}
+	log.Info(fmt.Sprintf("Try to take exclusive lock on file: %s", lockFilePath))
+	handlerLock, err := lockfile.New(lockFilePath)
+	if err != nil {
+		return handlerLock, errors.Wrapf(err, "failed to create lockFile for %s", lockFilePath)
+	}
+	err = wait.PollImmediateInfinite(5*time.Second, func() (done bool, err error) {
+		err = handlerLock.TryLock()
+		if err != nil {
+			log.Error(err, "retrying to lock handler")
+			return false, nil // Don't return the error here, it will not re-poll if we do
+		}
+		return true, nil
+	})
+	return handlerLock, err
 }
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -61,9 +61,13 @@ spec:
               value: "False"
             - name: PROFILER_PORT
               value: "6060"
+            - name: NMSTATE_INSTANCE_NODE_LOCK_FILE
+              value: "/var/k8s_nmstate/handler_lock"
           volumeMounts:
           - name: dbus-socket
             mountPath: /run/dbus/system_bus_socket
+          - name: nmstate-lock
+            mountPath: /var/k8s_nmstate
           securityContext:
             privileged: true
       volumes:
@@ -71,6 +75,9 @@ spec:
         hostPath:
           path: /run/dbus/system_bus_socket
           type: Socket
+      - name: nmstate-lock
+        hostPath:
+          path: /var/k8s_nmstate
 ---
 apiVersion: apps/v1
 kind: DaemonSet
@@ -134,9 +141,13 @@ spec:
               value: "False"
             - name: PROFILER_PORT
               value: "6060"
+            - name: NMSTATE_INSTANCE_NODE_LOCK_FILE
+              value: "/var/k8s_nmstate/handler_lock"
           volumeMounts:
             - name: dbus-socket
               mountPath: /run/dbus/system_bus_socket
+            - name: nmstate-lock
+              mountPath: /var/k8s_nmstate
           securityContext:
             privileged: true
       volumes:
@@ -144,6 +155,9 @@ spec:
           hostPath:
             path: /run/dbus/system_bus_socket
             type: Socket
+        - name: nmstate-lock
+          hostPath:
+            path: /var/k8s_nmstate
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/go-openapi/spec v0.19.4
 	github.com/gobwas/glob v0.2.3
 	github.com/kelseyhightower/envconfig v1.4.0
+	github.com/nightlyone/lockfile v1.0.0
 	github.com/onsi/ginkgo v1.10.1
 	github.com/onsi/gomega v1.7.0
 	github.com/operator-framework/operator-sdk v0.15.1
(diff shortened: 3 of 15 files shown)


cmd/manager/main.go
```diff
@@ -88,6 +90,27 @@ func main() {
...
+	// we don't have more than one working instance of k8s-nmstate
+	if lockFilePath, ok := os.LookupEnv("NMSTATE_INSTANCE_NODE_LOCK_FILE"); ok {
+		log.Info("Try to take exclusive lock on file: %s", lockFilePath)
+		lockFileMutex, err := lockfile.New(lockFilePath)
+		if err != nil {
+			log.Error(err, "Failed to create lockFile for NMSTATE_INSTANCE_NODE_LOCK_FILE")
+			os.Exit(1)
+		}
+		for {
+			err = lockFileMutex.TryLock()
```
phoracek, line 103: We can use some syntactic sugar here:

if err = lockFileMutex.TryLock(); err == nil {
sleep
log info
}

cmd/manager/main.go
```diff
@@ -88,6 +90,27 @@ func main() {
...
+			if err == nil {
+				log.Info("Successfully lock %s", lockFilePath)
```
phoracek, line 105: locked

cmd/manager/main.go
```diff
@@ -88,6 +90,27 @@ func main() {
...
+				break
+			}
+			time.Sleep(10 * time.Second)
+			log.Info("retry to get lock")
```
phoracek, line 109: Please add it at the beginning of the `for` as Trying to acquire the lock

deploy/operator.yaml
```diff
@@ -59,18 +59,23 @@ spec:
               value: "54874"
             - name: ENABLE_PROFILER
               value: "False"
-            - name: PROFILER_PORT
```
phoracek, line 62: This should be kept

phoracek:
It also seems that knmstate is unable to start serving

cmd/manager/main.go
```diff
@@ -88,6 +90,26 @@ func main() {
...
+	// Take exclusive instance lock if required, we need that to make sure that
+	// we don't have more than one working instance of k8s-nmstate
+	if lockFilePath, ok := os.LookupEnv("NMSTATE_INSTANCE_NODE_LOCK_FILE"); ok {
+		log.Info("Try to take exclusive lock on file: %s", lockFilePath)
+		lockFileMutex, err := lockfile.New(lockFilePath)
+		if err != nil {
+			log.Error(err, "Failed to create lockFile for NMSTATE_INSTANCE_NODE_LOCK_FILE")
+			os.Exit(1)
+		}
+		for {
```
qinqon, line 102: We need also timeouts to free resources in case of lock defects.
For that at knmstate usually use https://godoc.org/k8s.io/apimachinery/pkg/util/wait

Also I have found at project that mix  mutext and context so we don't need the for loop at all https://github.com/sasha-s/go-csync.

yboaron: Not sure that I understood your point here, could you please elaborate about timeouts to free resources thing (what do you mean by lock defects?)
And also about the usage of context mutex .

cmd/manager/main.go
```diff
@@ -88,6 +90,26 @@ func main() {
...
+			log.Info("try to get lock")
+			if err = lockFileMutex.TryLock(); err == nil {
+				log.Info("Successfully locked %s", lockFilePath)
+				break
+			}
+			time.Sleep(10 * time.Second)
+		}
+		defer lockFileMutex.Unlock()
```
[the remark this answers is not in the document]
yboaron: Yep, I'll update that - 10x!

cmd/manager/main.go
```diff
@@ -88,6 +90,26 @@ func main() {
 
 	printVersion()
 
+	// Take exclusive instance lock if required, we need that to make sure that
+	// we don't have more than one working instance of k8s-nmstate
+	if lockFilePath, ok := os.LookupEnv("NMSTATE_INSTANCE_NODE_LOCK_FILE"); ok {
```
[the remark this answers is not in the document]
yboaron: Will change that

yboaron:
> Good start, there is a defect at Unlock scope and we miss a more robust retry.
> 
> Question, do we need a `finalizer` or similar to clean the lock after delete in case there is no nmstate-handler running ?

Thanks @qinqon 
Could you please elaborate on this case ?, I'm not familiar with nmstate handlers details so maybe I'm missing something here.

cmd/manager/main.go
```diff
@@ -88,6 +90,20 @@ func main() {
 
 	printVersion()
 
+	// Take exclusive instance lock, we need that to make sure that
+	// we don't have more than one working instance of k8s-nmstate
+	lockFilePath, ok := os.LookupEnv("NMSTATE_INSTANCE_NODE_LOCK_FILE")
```
qinqon, line 95: Move this to the function.

cmd/manager/main.go
```diff
@@ -88,6 +90,20 @@ func main() {
...
+	if !ok {
+		log.Info("Failed to find NMSTATE_INSTANCE_NODE_LOCK_FILE ENV var")
+		os.Exit(1)
+	}
+	lockFileMutex, err := lockCreateAndLock(lockFilePath)
```
qinqon, line 100: name the function as `lockHandler` at the end this is what we do.

cmd/manager/main.go
```diff
@@ -212,3 +228,20 @@ func setProfiler() {
 		}()
 	}
 }
+
+func lockCreateAndLock(lockFilePath string) (lockfile.Lockfile, error) {
+	log.Info("Try to take exclusive lock on file: %s", lockFilePath)
+	lockFileMutex, err := lockfile.New(lockFilePath)
+	if err != nil {
+		log.Error(err, "Failed to create lockFile for NMSTATE_INSTANCE_NODE_LOCK_FILE")
```
qinqon, line 236: Do log the error here, just returning with 

```golang
return lockFileMutex, errors.Wrap(err, "failed to create lockFile for %s", lockFile) 
```

cmd/manager/main.go
```diff
@@ -212,3 +224,23 @@ func setProfiler() {
...
+	if err != nil {
+		return lockFileMutex, errors.Wrapf(err, "failed to create lockFile for %s", lockFilePath)
+	}
+	for {
+		log.Info("try to get lock")
+		if err = lockFileMutex.TryLock(); err == nil {
+			log.Info("Successfully locked %s", lockFilePath)
+			return lockFileMutex, err
+		}
+		time.Sleep(10 * time.Second)
```
oshoval, line 244: wonder if there is already blocking Lock , and then retry isnt needed ?
(prints + context switch will be eliminated, also less code)

for example 
https://github.com/juju/fslock

func (*Lock) Lock
func (l *Lock) Lock() error
Lock locks the lock. This call will block until the lock is available.

yboaron: @oshoval  - thanks for the feedback.

https://github.com/juju/fslock is based on flock, AFAIK, there are some cases that flock doesn't work (NFS).  I think we should stick with solution that doesn't sue flock.

oshoval: thanks

cmd/manager/main.go
```diff
@@ -212,3 +224,23 @@ func setProfiler() {
...
+	if err != nil {
+		return lockFileMutex, errors.Wrapf(err, "failed to create lockFile for %s", lockFilePath)
+	}
+	for {
+		log.Info("try to get lock")
+		if err = lockFileMutex.TryLock(); err == nil {
+			log.Info(fmt.Sprintf("Successfully locked %s", lockFilePath))
+			return lockFileMutex, err
+		}
+		time.Sleep(10 * time.Second)
```
qinqon, line 244: Is not very robust to use an infinite loop here we need a timeout let's use `PollInmmediate` from https://godoc.org/k8s.io/apimachinery/pkg/util/wait.

I suppose we need a bit timeout since we are waitting for CNAO installed kubernetes-nmstate-handler to finish.

cmd/manager/main.go
```diff
@@ -88,6 +91,15 @@ func main() {
 
 	printVersion()
 
+	// Take exclusive instance lock, we need that to make sure that
+	// we don't have more than one working instance of k8s-nmstate
+	lockFileMutex, err := lockHandler()
```
qinqon, line 96: `s/lockFileMutex/handlerLock/g`

deploy/operator.yaml
```diff
@@ -61,16 +61,23 @@ spec:
...
           securityContext:
             privileged: true
       volumes:
       - name: dbus-socket
         hostPath:
           path: /run/dbus/system_bus_socket
           type: Socket
+      - name: nmstate-lock
+        hostPath:
+          path: /etc/k8s_nmstate
```
qinqon, line 80: Who is creating this dir ?

qinqon:
@yboaron can you add a release note ? I think this PR merits it.

deploy/operator.yaml
yboaron: It should be created automatically (by the lock library), BTW, I run the following tests on local-cluster [1]:
A. run nmstste-handler deployment  - looks OK  (directory+file created)
B. ssh to node and delete nmstste-handler container (using docker ), a new container was created - looks OK

I guess we should add a test that runs two nmstate-handlers concurrently - maybe we can have it in a follow-up PR. 

[1] https://github.com/nmstate/kubernetes-nmstate/blob/master/docs/deployment-local-cluster.md

cmd/manager/main.go
```diff
@@ -212,3 +226,22 @@ func setProfiler() {
...
+	log.Info(fmt.Sprintf("Try to take exclusive lock on file: %s", lockFilePath))
+	handlerLock, err := lockfile.New(lockFilePath)
+	if err != nil {
+		return handlerLock, errors.Wrapf(err, "failed to create lockFile for %s", lockFilePath)
+	}
+	err = wait.PollImmediate(5*time.Second, 600*time.Second, func() (done bool, err error) {
+		if handlerLock.TryLock() == nil {
+			return true, nil
+		}
+		return false, nil
```
qinqon, line 244: Can we handle this like normal golang err + if and print the error ?

```golang
err := handlerLock.TryLock()
if err != nil {
    log.Error(err, "retrying to lock handler")
    return false, nil // Don't return the error here, it will not re-poll if we do
}
return true, nil
```

cmd/manager/main.go
```diff
@@ -212,3 +226,24 @@ func setProfiler() {
...
+	lockFilePath, ok := os.LookupEnv("NMSTATE_INSTANCE_NODE_LOCK_FILE")
+	if !ok {
+		return "", errors.New("Failed to find NMSTATE_INSTANCE_NODE_LOCK_FILE ENV var")
+	}
+	log.Info(fmt.Sprintf("Try to take exclusive lock on file: %s", lockFilePath))
+	handlerLock, err := lockfile.New(lockFilePath)
+	if err != nil {
+		return handlerLock, errors.Wrapf(err, "failed to create lockFile for %s", lockFilePath)
+	}
+	err = wait.PollImmediate(5*time.Second, 600*time.Second, func() (done bool, err error) {
```
phoracek, line 240: Why do we have a timeout here? Do you assume that during the upgrade, old operator will be gone within 10 minutes?

phoracek: @qinqon ^^

qinqon: Is not very rebust to leave this loop without timeout, we can increase the timeout if we want.

qinqon:
/retest

Humm we need to fix webhook test to start pod status check before we start the test with a goroutine.

qinqon:
/retest

hostname...

PR #488, opened 2020-04-01
Introduce a prefix for manifests

bcrochet:
In order to keep one operator from removing the resources of a
different operator, introduce a prefix that will add an additional
namespace to the resources.

Puts a prefix, if specified, in front of certain resources in order to keep them separate from possibly existing resources installed by a different operator.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
 HANDLER_IMAGE_SUFFIX ?=
 HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME)$(HANDLER_IMAGE_SUFFIX)
 HANDLER_IMAGE ?= $(IMAGE_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)
+HANDLER_PREFIX ?=
 
 NAMESPACE ?= nmstate
 PULL_POLICY ?= Always
@@ -100,7 +101,7 @@ gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
 manifests:
-	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR)
+	$(GO) run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(NAMESPACE) -handler-image=$(HANDLER_IMAGE) -handler-pull-policy=$(PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
 
 handler: gen-openapi gen-k8s gen-crds $(OPERATOR_SDK) manifests
 	$(OPERATOR_SDK) build $(HANDLER_IMAGE) --image-builder $(IMAGE_BUILDER)
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: {{ .Namespace }}
+  name: {{ .HandlerNamespace }}
   labels:
-    name: {{ .Namespace }}
+    name: {{ .HandlerNamespace }}
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -1,24 +1,25 @@
+{{define "handlerPrefix"}}{{with $prefix := .HandlerPrefix}}{{$prefix | printf "%s-"}}{{end -}}{{end}}
 ---
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: nmstate-handler
-  namespace: {{ .Namespace }}
+  name: {{template "handlerPrefix" .}}nmstate-handler
+  namespace: {{ .HandlerNamespace }}
 spec:
   selector:
     matchLabels:
-      name: nmstate-handler
+      name: {{template "handlerPrefix" .}}nmstate-handler
   template:
     metadata:
       labels:
         app: kubernetes-nmstate
-        name: nmstate-handler
+        name: {{template "handlerPrefix" .}}nmstate-handler
     spec:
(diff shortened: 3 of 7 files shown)


deploy/operator.yaml
```diff
@@ -2,23 +2,23 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: nmstate-handler
-  namespace: {{ .Namespace }}
+  name: {{with .HandlerPrefix}}{{printf "%s-" .}}{{end}}nmstate-handler
```
qinqon, line 5: Can we use a golang template var at the top of the template with the `with` case ? https://golang.org/pkg/text/template/#hdr-Variables

Makefile
```diff
@@ -7,6 +7,9 @@ HANDLER_IMAGE_NAME ?= kubernetes-nmstate-handler
 HANDLER_IMAGE_SUFFIX ?=
 HANDLER_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(HANDLER_IMAGE_NAME)$(HANDLER_IMAGE_SUFFIX)
 HANDLER_IMAGE ?= $(IMAGE_REGISTRY)/$(HANDLER_IMAGE_FULL_NAME)
+# If you are going to set a prefix for the manifests, you must also specify the flag:
+# HANDLER_PREFIX = '-handler-prefix myprefix'
```
qinqon, line 11: The `-handler-prefix` is not needed just `HANDLER_PREFIX = myprefix`

hack/render-manifests.go
```diff
@@ -14,25 +15,40 @@ func exitWithError(err error, cause string, args ...interface{}) {
...
+		fmt.Fprintf(os.Stderr, "usage: go run render-manifests.go [-handler-prefix handler-prefix] <handler-namespace> <handler-image> <handler-pull-policy> <input-dir> <output-dir>\n")
 		os.Exit(1)
 	}
 
 	inventory := Inventory{
-		Namespace:           os.Args[1],
-		NMStateHandlerImage: os.Args[2],
-		ImagePullPolicy:     os.Args[3],
+		HandlerNamespace:    flag.Arg(HandlerNamespaceIdx),
+		NMStateHandlerImage: flag.Arg(HandlerImageIdx),
```
qinqon, line 46: Rename to `HandlerImage` for consistency, It also means renaming the template token.

Makefile
```diff
@@ -100,7 +103,7 @@ gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
 manifests:
-	$(GO) run hack/render-manifests.go $(NAMESPACE) $(HANDLER_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR)
+	$(GO) run hack/render-manifests.go $(HANDLER_PREFIX) $(NAMESPACE) $(HANDLER_IMAGE) $(PULL_POLICY) deploy/ $(MANIFESTS_DIR)
```
qinqon, line 106: The named argument is missing `--handler-prefix`

deploy/namespace.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: {{ .Namespace }}
+  name: {{ .HandlerNamespace }}
```
qinqon, line 4: I think we have to put the prefix to the namespace too, if not the other operator will remove both namespaces, but correct way to do it is override at Makefile `HANDLER_NAMESPACE` variable.

deploy/operator.yaml
```diff
@@ -45,15 +45,15 @@ spec:
...
                 fieldRef:
                   fieldPath: spec.nodeName
             - name: INTERFACES_FILTER
               valueFrom:
                 configMapKeyRef:
-                  name: nmstate-config
+                  name: {{with .HandlerPrefix}}{{printf "%s-" .}}{{end}}nmstate-config
                   key: interfaces_filter
             - name: WEBHOOK_PORT
               value: "54874"
```
qinqon, line 59: More of a question, do we have to ensure that this port does not collide (the pod is running with network=host to be able to configure node networking) with the webhook installed by the other operator ? or is webhook going to be not run until we undeploy the other (with the lock PR).

hack/render-manifests.go
```diff
@@ -14,25 +15,40 @@ func exitWithError(err error, cause string, args ...interface{}) {
...
+		HandlerNamespace    string
 		NMStateHandlerImage string
-		ImagePullPolicy     string
+		HandlerPullPolicy   string
+		HandlerPrefix       string
 	}
 
-	if len(os.Args) < 6 {
-		fmt.Fprintf(os.Stderr, "usage: go run render-manifests.go [namespace] [image] [pull-policy] [input-dir] [output-dir]\n")
+	handlerPrefix := flag.String("handler-prefix", "", "Name prefix for the handler's resources")
```
qinqon, line 36: Maybe it's time to put naming args for all of it, this is growing like crazy.

deploy/namespace.yaml
bcrochet: I struggled with this one. I was thinking it should be there, but to me, the namespace is the one place where one can pass the whole thing, at once.  But I'm fine either way.

hack/render-manifests.go
bcrochet: I was thinking that same thing. That's why I did the iota!

deploy/operator.yaml
bcrochet: We can parameterize it if we want. You tell me.

qinqon: Nah let's skip it we will see.

deploy/namespace.yaml
```diff
@@ -1,6 +1,7 @@
+{{define "handlerPrefix"}}{{with $prefix := .HandlerPrefix}}{{$prefix | printf "%s-"}}{{end -}}{{end}}
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: {{ .Namespace }}
+  name: {{template "handlerPrefix" .}}{{ .HandlerNamespace }}
```
qinqon, line 5: Hummm I was all wrong, namespaces are different between CNAO and operator, we don't need to prefix it, sorry... can we go back to just namespace ?

PR #490, opened 2020-04-06
Add an exclusive lock for nmstate  handler instances (#487)

phoracek:

In upgrade process we might have more than one working nmstate handler
for some time, to avoid such a case we should use lock mechanism between handlers.

**Release note**:

```release-note
NONE
```


phoracek:

Only the tutorial test failed.

PR #496, opened 2020-04-13
DNM: This make policy work at ocp-4.4

qinqon:

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note

```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -57,7 +57,7 @@ spec:
                   name: {{template "handlerPrefix" .}}nmstate-config
                   key: interfaces_filter
             - name: WEBHOOK_PORT
-              value: "54874"
+              value: "54875"
             - name: ENABLE_PROFILER
               value: "False"
             - name: PROFILER_PORT
@@ -179,7 +179,7 @@ spec:
   publishNotReadyAddresses: true
   ports:
     - port: 443
-      targetPort: 54874
+      targetPort: 54875
   selector:
     name: {{template "handlerPrefix" .}}nmstate-handler
 ---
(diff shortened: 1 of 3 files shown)


deploy/operator.yaml
```diff
@@ -179,7 +179,7 @@ spec:
   publishNotReadyAddresses: true
   ports:
     - port: 443
-      targetPort: 54874
+      targetPort: 54875
```
phoracek, line 182: We randomly generated the original port, are you saying it is taken? Maybe this fixed the issue for you since we failed to cleanup the original allocation of the port.

deploy/role.yaml
```diff
@@ -51,13 +51,19 @@ metadata:
   name: {{template "handlerPrefix" .}}nmstate-handler
   namespace: {{ .HandlerNamespace }}
 rules:
-- apiGroups:
-  - certificates.k8s.io
+  # Istiod and bootstrap.
```
phoracek, line 54: Why do we need istio and what is bootstrap?

deploy/operator.yaml
qinqon: Looks like we were facing collision, I think the best solution (although it will use a litle more memory at masters) is to use DaemonSet for handling Reconcile and Deployment for webhook (so we can run it without host networking and we don't have the port issue)

deploy/role.yaml
qinqon: Copy paste from issue fixd at isstio related to k8s 1.18 also we are facing the same at ocp 4.4 but it only happends if we create csrs ourselves, if we go to openshift solution we don't care for this.

phoracek: If we know what it stands for, we can keep the comment. It is just a copy-paste with no descriptive value for us, we should drop it.

PR #497, opened 2020-04-14
Move webhook to a Deployment

qinqon:
/kind bug

This PR moves the webhook to a unprivileged `Deployment` without `hostNetwork: true`, this prevent port collision since the webhook port will be open at pod's netns.

It also runs `nmstate-handler` only at workers this is the common scenario and remove the WEBHOOK_PORT env var since it's no longer needed and remove complexity a little.

We run with two replicas so cert-manager using LeaderElection is being tested already.

**Release note**:

```release-note
Webhook runs now as unprivileged `Deployment`
action requiered: Wait for webhook Deployment Ready status too after deploy/upgrade.
```


pkg/webhook/nodenetworkconfigurationpolicy/server.go
```diff
@@ -38,7 +21,6 @@ func Add(mgr manager.Manager) error {
 	// 2.- Since we have delete the condition the status-mutate webhook get called and
 	//     there we set conditions to Unknown this final result will be updated.
 	server := webhookserver.New(mgr, webhookName, certificate.MutatingWebhook,
-		webhookserver.WithPort(webhookPort),
```
phoracek, line 41: We still have targetPort in service. How does it work now without explicit port being specified?

pkg/helper/client.go
```diff
@@ -37,6 +37,11 @@ var (
 )
 
 func init() {
+	_, runWebhookServer := os.LookupEnv("RUN_WEBHOOK_SERVER")
```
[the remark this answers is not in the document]
qinqon: Yep

pkg/webhook/nodenetworkconfigurationpolicy/server.go
qinqon: It will use kube-admission-webhook default port https://github.com/qinqon/kube-admission-webhook/blob/master/pkg/webhook/server/server.go#L35

phoracek: Then you gotta fix the manifest, `Service` still has `      targetPort: 54874`

deploy/operator.yaml
```diff
@@ -223,3 +191,14 @@ webhooks:
...
         apiVersions: ["v1alpha1"]
         resources: ["nodenetworkconfigurationpolicies", "nodenetworkconfigurationpolicies/status"]
+---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  namespace: {{ .HandlerNamespace }}
+  name: {{template "handlerPrefix" .}}nmstate-webhook
+spec:
+  minAvailable: "90%"
```
phoracek, line 201: We have two replicas, 90% would mean we need 2/2. I think we should set it so it requires at least one instance to be up and running.

deploy/operator.yaml
```diff
@@ -46,69 +45,38 @@ spec:
...
         app: kubernetes-nmstate
-        name: {{template "handlerPrefix" .}}nmstate-handler-worker
+        name: {{template "handlerPrefix" .}}nmstate-handler
     spec:
       # Needed to force vlan filtering config with iproute commands until
       # future nmstate/NM is in place.
       # https://github.com/nmstate/nmstate/pull/440
       hostNetwork: true
       serviceAccountName: {{template "handlerPrefix" .}}nmstate-handler
-      affinity:
```
phoracek, line 103: We need to add the `toleration` we had in the master.

deploy/operator.yaml
```diff
@@ -1,24 +1,23 @@
...
   selector:
     matchLabels:
-      name: {{template "handlerPrefix" .}}nmstate-handler
+      name: {{template "handlerPrefix" .}}nmstate-webhook
   template:
     metadata:
       labels:
         app: kubernetes-nmstate
-        name: {{template "handlerPrefix" .}}nmstate-handler
+        name: {{template "handlerPrefix" .}}nmstate-webhook
```
phoracek, line 19: * please rename the container from the pod as well. it is still "handler"
* I believe we don't need to set WATCH_NAMESPACE since we don't have any controller, do we? are we using POD_NAME?

pkg/webhook/nodenetworkconfigurationpolicy/server.go
phoracek: https://github.com/nmstate/kubernetes-nmstate/pull/497/files#diff-5e5f804a8ee2fb8c81711b728eb5a21dR150 still applies

deploy/operator.yaml
qinqon: Let's remove all we can to reduce stuff and see if it works.

qinqon: WHACH_NAMESPACE is needed

deploy/handler/operator.yaml
```diff
@@ -39,78 +40,49 @@ spec:
...
-      - name: dbus-socket
-        hostPath:
-          path: /run/dbus/system_bus_socket
-          type: Socket
-      - name: nmstate-lock
-        hostPath:
-          path: /var/k8s_nmstate
 ---
 apiVersion: apps/v1
 kind: DaemonSet
```
yboaron, line 57: Do we have a single daemonset? I can't find the workers Ds.

bcrochet:
Please rebase.

qinqon:
@bcrochet done, but I have to fix something there there are some race conditions at webhooks and controller-rutime cache and also I have to fix the cert-manager.

qinqon:
Manual test on openshift 4.5 works fine with a linux bridge without ports (just simple use case)

```bash
[ellorent@localhost kubernetes-nmstate]$ oc get nncp
NAME   STATUS
lbr1   SuccessfullyConfigured
[ellorent@localhost kubernetes-nmstate]$ oc get nnce
NAME                                    STATUS
lbednar-ocp45-8s8t2-master-0.lbr1       SuccessfullyConfigured
lbednar-ocp45-8s8t2-master-1.lbr1       SuccessfullyConfigured
lbednar-ocp45-8s8t2-master-2.lbr1       SuccessfullyConfigured
lbednar-ocp45-8s8t2-worker-268bg.lbr1   SuccessfullyConfigured
lbednar-ocp45-8s8t2-worker-cwnhs.lbr1   SuccessfullyConfigured
lbednar-ocp45-8s8t2-worker-fr6wm.lbr1   SuccessfullyConfigured
```

cmd/manager/main.go
```diff
@@ -124,18 +135,18 @@ func main() {
...
-	if err := controller.AddToManager(mgr); err != nil {
-		log.Error(err, "Cannot initialize controller")
-		os.Exit(1)
-	}
-
-	// Setup webhook on master only
-	if _, runWebhookServer := os.LookupEnv("RUN_WEBHOOK_SERVER"); runWebhookServer {
+	// Runs only webhook controllers if it's specified
+	if environment.IsWebhook() {
 		if err := webhook.AddToManager(mgr); err != nil {
```
yboaron, line 140: Any special reason not to add controllers in the same approach for all controllers? 
I mean for operator and handlers, we call all controllers, and each controller decides if it should be registered or not, and for webhook case, we only Add the relevant controller here.

deploy/handler/operator.yaml
```diff
@@ -1,37 +1,35 @@
 {{define "handlerPrefix"}}{{with $prefix := .HandlerPrefix}}{{$prefix | printf "%s-"}}{{end -}}{{end}}
 ---
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
```
yboaron, line 4: Q: We want the webhook pods to run on Masters only because of performance (k8s-api run on masters), right?

deploy/handler/operator.yaml
```diff
@@ -42,81 +40,74 @@ spec:
...
-                  - key: node-role.kubernetes.io/master
-                    operator: NotIn
-                    values:
-                      - ""
       nodeSelector:
         beta.kubernetes.io/arch: amd64
       {{- range $key, $value := .HandlerNodeSelector }}
       {{"  "}}{{- $key }}: {{  $value }}
       {{- end}}
+      tolerations:
```
yboaron, line 107: I can't see the master handlers DS, do we want to limit the handlers to run only on worker nodes?

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -109,24 +109,32 @@ var _ = Describe("NMState controller reconcile", func() {
...
-					Expect(ds.Spec.Template.Spec.NodeSelector).To(HaveKeyWithValue(k, v))
-				}
+			handlerKey := types.NamespacedName{Namespace: handlerNamespace, Name: handlerPrefix + "-nmstate-handler"}
+			err := cl.Get(context.TODO(), handlerKey, ds)
+			Expect(err).ToNot(HaveOccurred())
+			for k, v := range handlerNodeSelector {
+				Expect(ds.Spec.Template.Spec.NodeSelector).To(HaveKeyWithValue(k, v))
+			}
+		})
+		It("should NOT add NodeSelector to webhook deployment", func() {
```
yboaron, line 131: Any special reason to add this test?  you didn't include the  '.HandlerNodeSelector' in webhook deployment manifest?

deploy/handler/operator.yaml
```diff
@@ -42,81 +40,74 @@ spec:
...
-                fieldRef:
-                  fieldPath: spec.nodeName
-            - name: INTERFACES_FILTER
-              valueFrom:
-                configMapKeyRef:
-                  name: {{template "handlerPrefix" .}}nmstate-config
-                  key: interfaces_filter
-            - name: WEBHOOK_PORT
-              value: "54874"
+              value: "{{template "handlerPrefix" .}}nmstate"
```
yboaron, line 50: Why does the webhook deployment need to know the Operator name?

cmd/manager/main.go
```diff
@@ -81,9 +81,10 @@ func main() {
 
 	printVersion()
 
-	if !environment.IsOperator() {
-		// Take exclusive instance lock, we need that to make sure that
-		// we don't have more than one working instance of k8s-nmstate
+	// Lock only for handler we can run old and new version of
```
yboaron, line 84: nit, add coma  ( 'Lock only for handler, bla bla )

deploy/handler/operator.yaml
yboaron: I just read in the PR description that we removing the handlers to run only on workers by default. 
What the user should do if he wants to include also the masters?
If we want to change for example the k8s control plane network using nmstate ( for example run it as tagged traffic), I guess we need nmstate to run also on masters for that case

cmd/manager/main.go
qinqon: webhook are not controllers they are different concept at controler-runtime so they have different line.

deploy/handler/operator.yaml
qinqon: Is more about stability than performance, users will configure networking usually at workers so workers are going to be more instable alas breaking the webhook if we run it there also if you think about it the webhook is kind of knmstate control-plane so it belongs to masters.

deploy/handler/operator.yaml
qinqon: This PR has one type of handler daemonset that runs at master and workers, if we want to change just masters we can put in place a profile with nodeSelector `kubernetes.io/role/master=""`

pkg/controller/nmstate/nmstate_controller_test.go
qinqon: The operator's CR NodeSelector is only for handlers (we want to skip nodes that does not support nmstate) webhook is transparent for that since is just ensuring the nncp status is correctly reset.

deploy/handler/operator.yaml
qinqon: Manager was failing operator-sdk need it (I don't remeber, also we need it to point to MutatingWebhookConfiguration)

test/runner/pod.go
```diff
@@ -10,23 +10,23 @@ import (
...
+func runAtPods(pods []string, arguments ...string) {
+	for _, pod := range pods {
+		exec := []string{"exec", "-n", framework.Global.Namespace, pod, "--"}
 		execArguments := append(exec, arguments...)
 		_, err := cmd.Kubectl(execArguments...)
-		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+		ExpectWithOffset(2, err).ToNot(HaveOccurred())
 	}
 }
+func RunAtHandlerPods(arguments ...string) {
```
phoracek, line 28: Please add an empty line above

cluster/sync.sh
```diff
@@ -41,16 +32,25 @@ function consistently {
     done
 }
 
-function isHandlerOk {
-    desiredNumberScheduled=$(getDesiredScheduledHandlers $1)
-    numberAvailable=$(getAvailableHandlers $1)
+function isDaemonSetOk {
+
```
phoracek, line 36: Please drop the line

cluster/sync.sh
```diff
@@ -41,16 +32,25 @@ function consistently {
...
+    desiredNumberScheduled=$($kubectl get daemonset -n $1 -l $2 -o=jsonpath='{..status.desiredNumberScheduled}')
```
phoracek, line 37: Why do we have two dots in the jsonpath?

cmd/manager/main.go
```diff
@@ -81,9 +81,10 @@ func main() {
 
 	printVersion()
 
-	if !environment.IsOperator() {
-		// Take exclusive instance lock, we need that to make sure that
-		// we don't have more than one working instance of k8s-nmstate
+	// Lock only for handler, we can run old and new version of
+	// webhook without problems, policy status will be updated
+	// by multiple instances.
+	if environment.IsHandler() {
```
phoracek, line 87: The fact that we need this kind of `if` several times in the main function suggests we should maybe create a function mainWebhook, mainHandler, mainOperator. They would share some logic, but the result would be way more readable. If somebody wants to follow execution of a webhook, they just jump to that function.

deploy/handler/operator.yaml
```diff
@@ -1,37 +1,35 @@
 {{define "handlerPrefix"}}{{with $prefix := .HandlerPrefix}}{{$prefix | printf "%s-"}}{{end -}}{{end}}
 ---
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
-  name: {{template "handlerPrefix" .}}nmstate-handler
```
phoracek, line 6: We will need to add removal code for this to CNAO. Like the one we have for multus.

test/e2e/handler/prepare.go
```diff
@@ -59,3 +56,24 @@ func waitForDaemonSets(t *testing.T, kubeclient kubernetes.Interface, namespace
 	})
 	return err
 }
+
+func waitForDeployments(t *testing.T, kubeclient kubernetes.Interface, namespace string, retryInterval, timeout time.Duration) error {
+	if framework.Global.LocalOperator {
+		return nil
+	}
+	err := wait.PollImmediate(retryInterval, timeout, func() (done bool, err error) {
```
phoracek, line 64: Let's make this breathe, please insert an empty line above this one.

test/e2e/handler/prepare.go
```diff
@@ -59,3 +56,24 @@ func waitForDaemonSets(t *testing.T, kubeclient kubernetes.Interface, namespace
...
+		filterByApp := metav1.ListOptions{LabelSelector: "app=kubernetes-nmstate"}
+		deployments, err := kubeclient.AppsV1().Deployments(namespace).List(filterByApp)
+		if err != nil {
+			return true, errors.Wrapf(err, "failed retrieving daemon sets for namespace %s", namespace)
+		}
+		for _, deployment := range deployments.Items {
```
phoracek, line 70: ditto

test/e2e/handler/prepare.go
```diff
@@ -59,3 +56,24 @@ func waitForDaemonSets(t *testing.T, kubeclient kubernetes.Interface, namespace
...
+			By(fmt.Sprintf("Checking deployment %s", deployment.Name))
+			if int(deployment.Status.AvailableReplicas) >= 2 {
+				return true, nil
+			}
+		}
+		return true, nil
+	})
+	return err
```
phoracek, line 78: ditto

test/e2e/handler/prepare.go
```diff
@@ -59,3 +56,24 @@ func waitForDaemonSets(t *testing.T, kubeclient kubernetes.Interface, namespace
...
+		if err != nil {
+			return true, errors.Wrapf(err, "failed retrieving daemon sets for namespace %s", namespace)
+		}
+		for _, deployment := range deployments.Items {
+			By(fmt.Sprintf("Checking deployment %s", deployment.Name))
+			if int(deployment.Status.AvailableReplicas) >= 2 {
+				return true, nil
+			}
+		}
+		return true, nil
```
phoracek, line 76: ditto

pkg/environment/environment.go
```diff
@@ -0,0 +1,25 @@
+package environment
```
phoracek, line 1: I won't block this PR on this, but this is wrong. It keeps functions in a single module grouped by the function they use, not by their "unit". That makes this into a junkyard of env variable handling.

This points at the issue that we are coupling everything together here. We should have completely separated lane for every of these modes (operator, handler, webhook). We should read the parameter (IsOperator) in main and just run the right component. Just ranting, I think we will need to make some order in the code-base, sooner or later.

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -26,7 +26,7 @@ var _ = Describe("NMState controller reconcile", func() {
 		cl                  client.Client
 		reconciler          ReconcileNMState
 		existingNMStateName = "nmstate"
-		dsNodeSelector      = map[string]string{"selector_1": "value_1", "selector_2": "value_2"}
+		handlerNodeSelector = map[string]string{"selector_1": "value_1", "selector_2": "value_2"}
```
phoracek, line 29: so many unrelated changes making this harder to review and spot real issues, tsk tsk tsk.

deploy/handler/role.yaml
```diff
@@ -76,10 +76,9 @@ rules:
   - pods
   - nodes
   - namespaces
+  - secrets
   verbs:
-  - get
-  - list
-  - watch
+  - '*'
```
phoracek, line 81: We only need to create secrets, rights? Please add a new item in this list specifically for secrets and verbs they need.

cluster/sync.sh
qinqon: To simplify expression and just read the first occurrence of that field now we have just one DaemonSet

cmd/manager/main.go
qinqon: Will do a follow up PR.

pkg/environment/environment.go
qinqon: At the follow-up PR we will refactor this.

deploy/handler/operator.yaml
qinqon: noted

deploy/handler/role.yaml
```diff
@@ -77,9 +84,15 @@ rules:
   - nodes
   - namespaces
   verbs:
+  - '*'
```
phoracek, line 87: Do we need to keep this changed now when we moved secrets out?

PR #512, opened 2020-04-22
Only reconcile a single CR with a specific name

bcrochet:

>
/kind bug
> /kind enhancement

There should be no more than one kubernetes-nmstate handler. If the
CR is created with a name other than the constant, just ignore the CR
and return.

**Release note**:

```release-note
Only one NMstate.nmstate.io CR should be allowed. This change will ignore any CR that is not named `nmstate`.
```

### DIFF
--- a/pkg/controller/nmstate/nmstate_controller.go
+++ b/pkg/controller/nmstate/nmstate_controller.go
@@ -74,13 +74,19 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling NMstate")
 
+	// We won't create more than one kubernetes-nmstate handler
+	if request.Name != names.NMstateResourceName {
+		reqLogger.Info("Ignoring NMstate.nmstate.io without default name")
+		return reconcile.Result{}, nil
+	}
+
 	// Fetch the NMstate instance
 	instance := &nmstatev1alpha1.NMstate{}
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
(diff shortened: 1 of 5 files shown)


qinqon:
/hold
Let's wait to have proper operator e2e tests so we can add some testing for this

pkg/controller/nmstate/nmstate_controller.go
```diff
@@ -74,6 +74,12 @@ func (r *ReconcileNMstate) Reconcile(request reconcile.Request) (reconcile.Resul
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling NMstate")
 
+	// We won't create more than one kubernetes-nmstate handler
```
yboaron, line 77: Don't we need to check also the namespace name ? I mean we should support a singleton NMstate instance named: default_namespace.defaultNMstate_name

pkg/names/names.go
```diff
@@ -2,3 +2,6 @@ package names
 
 // ManifestDir is the directory where manifests are located.
 const ManifestDir = "./bindata"
+
+// NMstateResourceName is the name of the CR that the operator will reconcile
+const NMstateResourceName = "nmstate"
```
yboaron, line 7: IIUC, the user should create the NMstate instance, so I think It would be useful to document this requirement somewhere.

pkg/controller/nmstate/nmstate_controller.go
qinqon: I think this is a cluster-wide CRD so no namespace like NodeNetworkState and friends.

pkg/names/names.go
bcrochet: Well, yes and no. Under CVO, that would not be the case. CNO does it this way. I will see how they document it.

bcrochet:
> @bcrochet do we want to do unit test of this or do we do some proper e2e test stuff ?

I'm working on the e2e part. We could add some unit tests for the time being, merge it, then work on the e2e. WDYT?

qinqon:
@bcrochet let's add some unit test to this and we merge.

qinqon:
/hold cancel

We have a unit test now that covers it.

PR #514, opened 2020-04-22
Add optional node selector field to NMstate CRD

yboaron:
In current implementation, the handler pods deployed on all the masters and workers nodes.
With node selector, the user can specify on which nodes the handler should run on. in order to do that, the user should first set a label on the relevant nodes and then set the same value in NMstate node selector field.

Signed-off-by: Yossi Boaron yboaron@redhat.com

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
Add NodeSelector optional field to operator spec, to control on which nodes the handler should run on.
```


qinqon:
/hold
Are we sure we want this, since we can do same thing at policy level @phoracek ?

yboaron:
Well, yep, with policy we have the flexibility to define on which nodes the config should be applied. 
But for that to happen, we need to run the handler (consume resources on the node, just to check and ignore every policy request)  even on nodes that we don't want to run NMstate. 

I think that these two features are Complementary.

IIUC, the motivation for this feature is avoiding running NMState in nodes that don't satisfy the requirements (NM version and maybe other things), another way to tackle that could be by adding Initcontainer to handler DS (I need to check if it is possible to add initcontainer to DS), and this initcontainer should verify if node satisfies NMState requirements.

phoracek:
We cannot do it via Policies. This is about not running knmstate on some nodes (like masters and incompatible workers) at all. We don't want users to have any chance to run configuration on nodes we know will not work. And we don't want to take those extra resources.

pkg/apis/nmstate/v1alpha1/nmstate_types.go
```diff
@@ -11,6 +11,17 @@ import (
...
+}
+
+// NMstateSpec defines the desired state of NMstate
+type NMstateSpec struct {
+	// ExtraNodeSelector is an optional selector that will be added to handler DaemonSet manifest
+	// for both workers and masters (https://github.com/nmstate/kubernetes-nmstate/blob/master/deploy/handler/operator.yaml).
+	// If ExtraNodeSelector is specified, the handler will run only on nodes that have each of the indicated key-value pairs
+	// as labelsbe applied to the node.
+	// +optional
+	ExtraNodeSelector map[string]string `json:"extranodeselector"`
```
phoracek, line 24: I think that the extra node is extra. Maybe just NodeSelector? And please make the json to nodeSelector.

qinqon: +1

pkg/apis/nmstate/v1alpha1/nmstate_types.go
```diff
@@ -11,6 +11,17 @@ import (
 type NMstate struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              NMstateSpec `json:"spec,omitempty"`
+}
+
+// NMstateSpec defines the desired state of NMstate
+type NMstateSpec struct {
+	// ExtraNodeSelector is an optional selector that will be added to handler DaemonSet manifest
```
qinqon, line 19: Still `Extra` in the comment

deploy/handler/operator.yaml
```diff
@@ -23,6 +23,9 @@ spec:
       nodeSelector:
         beta.kubernetes.io/arch: amd64
         node-role.kubernetes.io/master: ""
+      {{range $key, $value := .HandlerNodeSelector}}
+        {{$value | printf "%s: %s" $key }}
```
qinqon, line 27: What about `{{ $key -}}: {{- $value }}` although dealing with new lines is a pain...

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -69,4 +69,32 @@ var _ = Describe("NMstate controller reconcile", func() {
...
+
+			reconciler.client = cl
+
+			request.Name = existingNMstateName
+		})
+		It("should return empty Result and no error", func() {
+			result, err := reconciler.Reconcile(request)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result).To(Equal(reconcile.Result{}))
+		})
```
qinqon, line 97: We have to retrieve DaemonSets from handler and check node selector is set there.

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -69,4 +69,32 @@ var _ = Describe("NMstate controller reconcile", func() {
...
 		})
 	})
+	Context("node selector is set ", func() {
+		var (
+			request reconcile.Request
+		)
+		BeforeEach(func() {
+			s := scheme.Scheme
+			s.AddKnownTypes(nmstatev1alpha1.SchemeGroupVersion,
+				&nmstatev1alpha1.NMstate{},
```
qinqon, line 79: maybe we have to add DaemonSet here so we can tretrieve it later to heck nodeSelctor ?

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -69,4 +129,36 @@ var _ = Describe("NMState controller reconcile", func() {
...
+			cl = fake.NewFakeClientWithScheme(s, objs...)
+			reconciler.client = cl
+			request.Name = existingNMStateName
+		})
+		It("NodeSelector labels should be defined for Daemonsets", func() {
+			result, err := reconciler.Reconcile(request)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result).To(Equal(reconcile.Result{}))
+			// verify NodeSelector value for both worker and master DS
+			ds := &appsv1.DaemonSet{}
```
qinqon, line 154: You can do cl.List and filter by namespace I think

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -16,32 +22,86 @@ import (
...
+	cpCmd = exec.Command("cp", "-f", "../../../deploy/handler/service_account.yaml", "./testdata/kubernetes-nmstate/rbac/")
+	cpCmd.Run()
+	cpCmd = exec.Command("cp", "-f", "../../../deploy/handler/role.yaml", "./testdata/kubernetes-nmstate/rbac/")
+	cpCmd.Run()
+	cpCmd = exec.Command("cp", "-f", "../../../deploy/handler/role_binding.yaml", "./testdata/kubernetes-nmstate/rbac/")
+	cpCmd.Run()
+}
+
+func deleteManifest() {
+	cpCmd := exec.Command("rm", "-f", "./testdata/kubernetes-nmstate/crds/nmstate.io_nodenetworkconfigurationenactments_crd.yaml")
```
qinqon, line 45: You can remove the directory with the golang os module

```golang
os.RemoveAll("/testdata/")
```

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -16,32 +22,86 @@ import (
 	"github.com/nmstate/kubernetes-nmstate/pkg/names"
 )
 
+func copyManifest() {
```
qinqon, line 25: We have to mimic
```
COPY deploy/crds/nmstate.io_nodenetwork*.yaml /bindata/kubernetes-nmstate/crds/
COPY deploy/handler/namespace.yaml /bindata/kubernetes-nmstate/namespace/
COPY deploy/handler/operator.yaml /bindata/kubernetes-nmstate/handler/handler.yaml
COPY deploy/handler/service_account.yaml /bindata/kubernetes-nmstate/rbac/
COPY deploy/handler/role.yaml /bindata/kubernetes-nmstate/rbac/
COPY deploy/handler/role_binding.yaml /bindata/kubernetes-nmstate/rbac/
```
So both CRDs can be use a wildcard

Also can we do a `copyFile` wrapper function to reduce code here ? so we do 
`copyFile("..", "...") 

Also check errors here with `Expect(err).ToNot(HaveOcurred())`

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -16,32 +22,86 @@ import (
...
+
+}
+
+var _ = BeforeSuite(func() {
+	copyManifest()
+})
+
+var _ = AfterSuite(func() {
+	deleteManifest()
+})
```
qinqon, line 70: Can we put this at the _suite_ file ?

pkg/controller/nmstate/nmstate_controller_test.go
yboaron: Yes, it's more elegant to List DS here, but I'm getting [1] error when trying to List DS [2]
Any idea what am I missing here?

[1] tems[0] can't assign or convert unstructured.Unstructured into v1.DaemonSet
[2]
                        dsList := appsv1.DaemonSetList{}
                        err := cl.List(context.TODO(), &dsList, &client.ListOptions{ Namespace: "nmstate"})

pkg/controller/nmstate/nmstate_controller_test.go
yboaron: I think that I addressed most of your comments here

pkg/controller/nmstate/nmstate_controller_test.go
qinqon: Can we test with client.InNamespace [1] ?

[1] https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client_test.go#L1793

pkg/controller/nmstate/nmstate_controller_test.go
```diff
@@ -69,4 +81,37 @@ var _ = Describe("NMState controller reconcile", func() {
 			Expect(result).To(Equal(reconcile.Result{}))
 		})
 	})
+	Context("node selector is set in Operator spec", func() {
```
qinqon, line 84: "when node ..."

pkg/controller/nmstate/nmstate_suite_test.go
```diff
@@ -9,6 +12,68 @@ import (
...
+		return err
+	}
+	if srcinfo, err = os.Stat(src); err != nil {
+		return err
+	}
+	return os.Chmod(dst, srcinfo.Mode())
+}
+
+func copyManifests() error {
+	dirArr := []string{"crds/", "namespace/", "handler/", "rbac/"}
```
qinqon, line 43: This is not needed is implicit at the srcToDest

pkg/controller/nmstate/nmstate_suite_test.go
```diff
@@ -9,6 +12,68 @@ import (
...
+		"../../../deploy/crds/nmstate.io_nodenetworkconfigurationenactments_crd.yaml": "./testdata/kubernetes-nmstate/crds/",
+		"../../../deploy/crds/nmstate.io_nodenetworkconfigurationpolicies_crd.yaml":   "./testdata/kubernetes-nmstate/crds/",
+		"../../../deploy/crds/nmstate.io_nodenetworkstates_crd.yaml":                  "./testdata/kubernetes-nmstate/crds/",
+		"../../../deploy/handler/namespace.yaml":                                      "./testdata/kubernetes-nmstate/namespace/",
+		"../../../deploy/handler/operator.yaml":                                       "./testdata/kubernetes-nmstate/handler/handler.yaml",
+		"../../../deploy/handler/service_account.yaml":                                "./testdata/kubernetes-nmstate/rbac/",
+		"../../../deploy/handler/role.yaml":                                           "./testdata/kubernetes-nmstate/rbac/",
+		"../../../deploy/handler/role_binding.yaml":                                   "./testdata/kubernetes-nmstate/rbac/",
+	}
+	for _, dir := range dirArr {
```
qinqon, line 54: We can move this to `copyFile` extracting path from destination file no problem on calling MkdirAll if we have already created the dir.

yboaron: Yep, good idea, I'll change that.

pkg/controller/nmstate/nmstate_suite_test.go
```diff
@@ -9,6 +12,66 @@ import (
...
+		"../../../deploy/crds/nmstate.io_nodenetworkstates_crd.yaml":                  "./testdata/kubernetes-nmstate/crds/",
+		"../../../deploy/handler/namespace.yaml":                                      "./testdata/kubernetes-nmstate/namespace/",
+		"../../../deploy/handler/operator.yaml":                                       "./testdata/kubernetes-nmstate/handler/handler.yaml",
+		"../../../deploy/handler/service_account.yaml":                                "./testdata/kubernetes-nmstate/rbac/",
+		"../../../deploy/handler/role.yaml":                                           "./testdata/kubernetes-nmstate/rbac/",
+		"../../../deploy/handler/role_binding.yaml":                                   "./testdata/kubernetes-nmstate/rbac/",
+	}
+
+	for src, dest := range srcToDest {
+		if err := os.MkdirAll(dest, os.ModePerm); err != nil {
```
qinqon, line 55: I mean doing `os.MkdirAll` at `copyFile`

PR #517, opened 2020-04-28
Move handler tests to new target

bcrochet:
In order to run the handler e2e on their own, this moves them to their
own Makefile target, but maintains the old e2e target by using the new
one as a dependency. Then, when operator e2e tests are introduced, they
can both be called individually or as a group via test/e2e.

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
NONE
```


test/e2e/operator/prepare.go
```diff
@@ -0,0 +1,64 @@
...
+	cleanupTimeout := time.Second * 5
+	ctx := framework.NewTestCtx(t)
+	err := ctx.InitializeClusterResources(&framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
+	Expect(err).ToNot(HaveOccurred(), "cluster resources not initialized")
+
+	// get namespace
+	By("Check operator is up and running")
+	namespace, err := ctx.GetNamespace()
+	Expect(err).ToNot(HaveOccurred())
+	err = waitForDeployment(t, framework.Global.KubeClient, namespace, time.Second*5, time.Second*90)
```
qinqon, line 34: We can use operator-sdk `WaitForDeployment` one it does not have `WaitForDaemonSet` that' why we copied here and modified

[1] https://github.com/operator-framework/operator-sdk/blob/master/pkg/test/e2eutil/wait_util.go#L43

test/e2e/operator/main_test.go
```diff
@@ -0,0 +1,83 @@
...
+	framework.MainEntry(m)
+}
+
+func TestE2E(tapi *testing.T) {
+	t = tapi
+	RegisterFailHandler(Fail)
+
+	By("Getting node list from cluster")
+	nodeList := corev1.NodeList{}
+	filterWorkers := dynclient.MatchingLabels{"node-role.kubernetes.io/worker": ""}
```
qinqon, line 52: Let's ignore the filters for operator tests.

Makefile
```diff
@@ -131,14 +131,24 @@ push: push-handler push-operator
...
+test-e2e-operator: $(OPERATOR_SDK)
+	mkdir -p test_logs/e2e/operator
+	unset GOFLAGS && $(OPERATOR_SDK) test local ./test/e2e/operator \
+		--kubeconfig $(KUBECONFIG) \
+		--namespace $(HANDLER_NAMESPACE) \
+		--no-setup \
+		--go-test-flags "$(e2e_test_args)"
+
+
+test/e2e: test-e2e-operator test-e2e-handler
```
qinqon, line 151: As extra mile can we rename this to test-e2e too ? It will be to be renamted at automation/... too

Makefile
```diff
@@ -131,14 +131,24 @@ push: push-handler push-operator
...
 		--go-test-flags "$(e2e_test_args)"
 
+test-e2e-operator: $(OPERATOR_SDK)
+	mkdir -p test_logs/e2e/operator
+	unset GOFLAGS && $(OPERATOR_SDK) test local ./test/e2e/operator \
+		--kubeconfig $(KUBECONFIG) \
+		--namespace $(HANDLER_NAMESPACE) \
+		--no-setup \
+		--go-test-flags "$(e2e_test_args)"
+
```
qinqon, line 149: Maybe it's time to move this out of Makefile into hack/run-e2e-test.sh or the like so we continue reducing Makefile complexity.

qinqon:
@bcrochet is all good for me, my last proporsal here is to have two prow jobs one for operator and the other for the handler what do you think ?

bcrochet:
> @bcrochet is all good for me, my last proporsal here is to have two prow jobs one for operator and the other for the handler what do you think ?

Sounds good to me.

qinqon:
/hold

@bcrochet I miss an important bit, for e2e operator tests we have to deploy just the operator not the CR so we need a `cluster-sync-operator` and call it at `cluster-sync` and deploy the CR.

Another option since this is a `normal` operator the operator-sdk allow you to run it out of the cluster so you don't have to deploy it to run e2e tests.

bcrochet:
> /hold
> 
> @bcrochet I miss an important bit, for e2e operator tests we have to deploy just the operator not the CR so we need a `cluster-sync-operator` and call it at `cluster-sync` and deploy the CR.
> 
> Another option since this is a `normal` operator the operator-sdk allow you to run it out of the cluster so you don't have to deploy it to run e2e tests.
> 
> https://github.com/operator-framework/operator-sdk/blob/93a4933c956752dc7543728c929fc281b90ef194/website/content/en/docs/golang/e2e-tests.md#up-local-flag

See latest commit

PR #519, opened 2020-05-05
Show only operator at release

qinqon:

/kind bug
/kind infrastructure

Right now we show the handler and also we don't include the CRDs, the
release has to show just the operator manifest and image since handler
is deployed by it.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -167,13 +167,16 @@ prepare-major: $(RELEASE_NOTES)
...
+release: OPERATOR_IMAGE_SUFFIX =: $(shell hack/version.sh)
+release: HANDLER_IMAGE_SUFFIX =: $(shell hack/version.sh)
+release: manifests push-handler push-operator $(description) $(GITHUB_RELEASE) version/version.go
 	DESCRIPTION=$(description) \
 	GITHUB_RELEASE=$(GITHUB_RELEASE) \
 	TAG=$(shell hack/version.sh) \
 				   hack/release.sh \
-						$(shell find $(MANIFESTS_DIR) -type f)
+						$(shell find $(MANIFESTS_DIR) -type f) \
+						deploy/crds/nmstate.io_nmstates_crd.yaml \
```
phoracek, line 178: Why do we need these two CRDs? Aren't they in the MANIFESTS_DIR already?

qinqon: Nope, at manifests we have only the rendered stuff, CRDs are not passed to hack/render-manifests.go

phoracek: Gotcha, thanks

PR #522, opened 2020-05-05
add installation guide to release notes

phoracek:

>
> /kind bug

/kind enhancement

For people looking for installation guide per release in GitHub release
notes.

Dependent on https://github.com/nmstate/kubernetes-nmstate/pull/521.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/prepare-release.sh
+++ b/hack/prepare-release.sh
@@ -33,8 +33,29 @@ cat $release_notes >> version/description
 
 cat << EOF >> version/description
 
+# Installation
+
+First, install kubernetes-nmstate operator:
+
 \`\`\`
-docker pull OPERATOR_IMAGE
+kubectl apply -f https://github.com/nmstate/kubernetes-nmstate/releases/download/$new_version/nmstate.io_nmstates_crd.yaml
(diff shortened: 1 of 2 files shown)


hack/prepare-release.sh
```diff
@@ -33,8 +33,29 @@ cat $release_notes >> version/description
...
+Once that's done, create an \`NMState\` CR, triggering deployment of
+kubernetes-nmstate handler:
+
+\`\`\`
+cat <<EOF | kubectl create -f -
+apiVersion: nmstate.io/v1alpha1
+kind: NMState
+metadata:
+  name: nmstate
+EOF
```
qinqon, line 58: We missing waiting for readiness

phoracek: We don't have a status section on the NMState CR, so there is no simple way other than waiting for all the DaemonSets which is far from elegant. I would keep the doc simple and add waiting only once we have it in the CR.

qinqon: Fair enough

/lgtm
/approve

PR #532, opened 2020-05-12
Fix release target

qinqon:

>
/kind bug
/kind infrastructure

We have a few problems with at the release infra, this PR fixes the following:

- Move release steps from Makefile into hack/release.sh
- Create `make tools` target to install vendoring tools that calls `hack/install-tools.sh`
- Remove BASE_TAG from operator's Dockerfile

**Release note**:

```release-note
NONE
```


phoracek:
Please split this into 3 PRs. It would reduce the number of touched files and more importantly it would allow us to revert part of the changes if they cause issues.

PR #533, opened 2020-05-13
Makefile: remove build dependency on generated files

bcrochet:
If generated files are not checked in when they should be, a diff
between a built image and what is in the repo could occur. This would
not be desirable. So, remove the Makefile dependency, and add a check
script that makes sure that what is commited to the repo is the same
as what should be committed.

>
> /kind bug
/kind enhancement

It removes the dependency of building the handler on generating files.

It would be undesirable for files to get generated at build time, and not be committed to the repo. It opens the possibility that a released container image would contain code that is not in the commit it was derived from. So, now, before the e2e tests, we check that files would not be generated. If they would be, then fail the e2e tests fast.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,10 +116,15 @@ gen-openapi: $(OPENAPI_GEN)
 gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
+check-gen: generate
+	./hack/check-gen.sh
+
+generate: gen-openapi gen-k8s gen-crds
+
 manifests: $(GO)
 	$(GO) run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(HANDLER_NAMESPACE) -operator-namespace=$(OPERATOR_NAMESPACE) -handler-image=$(HANDLER_IMAGE) -operator-image=$(OPERATOR_IMAGE) -handler-pull-policy=$(HANDLER_PULL_POLICY) -operator-pull-policy=$(OPERATOR_PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
 
-handler: gen-openapi gen-k8s gen-crds $(OPERATOR_SDK)
+handler: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) build $(HANDLER_IMAGE) --image-builder $(IMAGE_BUILDER)
 push-handler: handler
 	$(IMAGE_BUILDER) push $(HANDLER_IMAGE)
@@ -194,6 +199,8 @@ tools-vendoring:
 	push-handler \
 	test/unit \
 	test/e2e \
+	generate \
+	check-gen \
 	cluster-up \
 	cluster-down \
 	cluster-sync-handler \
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -194,6 +199,7 @@ tools-vendoring:
 	push-handler \
 	test/unit \
 	test/e2e \
+	generate \
```
qinqon, line 202: gen-check is missing

Makefile
```diff
@@ -116,10 +116,15 @@ gen-openapi: $(OPENAPI_GEN)
 gen-crds: $(OPERATOR_SDK)
 	$(OPERATOR_SDK) generate crds
 
+gen-check: generate
```
qinqon, line 119: Feels like it's generating a check, let's call this check-gen or check-generate

Makefile
bcrochet: We have some others that are missing. I'll fix this one, and send a separate PR for the others.

automation/check-patch.e2e-k8s.sh
```diff
@@ -20,6 +20,9 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    # Let's fail fast if generated files differ
+    make gen-check
```
qinqon, line 24: check-gen

automation/check-patch.e2e-ocp.sh
```diff
@@ -20,6 +20,9 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    # Let's fail fast if generated files differ
+    make gen-check
```
qinqon, line 24: ditto

automation/check-patch.e2e-okd.sh
```diff
@@ -23,6 +23,9 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    # Let's fail fast if generated files differ
+    make gen-check
```
qinqon, line 27: ditto

automation/check-patch.e2e-operator-k8s.sh
```diff
@@ -20,6 +20,9 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    # Let's fail fast if generated files differ
+    make gen-check
```
qinqon, line 24: ditto

automation/check-patch.e2e-operator-ocp.sh
```diff
@@ -20,6 +20,9 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    # Let's fail fast if generated files differ
+    make gen-check
```
qinqon, line 24: ditto

automation/check-patch.e2e-operator-okd.sh
```diff
@@ -23,6 +23,9 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    # Let's fail fast if generated files differ
+    make gen-check
```
qinqon, line 27: ditto

PR #539, opened 2020-05-13
Create some e2e tests for the operator

bcrochet:
This is a few tests:

Install handler
Ignore a second handler CR
Do nothing with a handler CR that has the wrong name
Uninstall handler

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


cluster/sync-operator.sh
```diff
@@ -0,0 +1,109 @@
...
+    # Cleanup previous deployment, if there is any
+    make cluster-clean
+
+    push
+
+    if isOpenshift; then
+        while ! $kubectl get securitycontextconstraints; do
+            sleep 1
+        done
+        $kubectl apply -f ${MANIFESTS_DIR}/scc.yaml
```
qinqon, line 81: I think we don't need scc for operator since there is no privileged container, right ?

cluster/sync-operator.sh
```diff
@@ -0,0 +1,109 @@
...
+    while ! $cmd; do
+        sleep $interval
+        timeout=$(( $timeout - $interval ))
+        if [ $timeout -le 0 ]; then
+            return 1
+        fi
+    done
+}
+
+function consistently {
```
qinqon, line 21: Can we move helper functions to to a place to reuse them ? like sync-common.sh and we source them ?
They would be:
- eventually
- consistently
- isExternal
- isOpenshift
- isDeploymentOk
- push

cluster/sync-operator.sh
qinqon, line 81: SCC are only needed for handler since it has a privileged containers, and it has to be deployed by operator, I think we don't need scc for operator itself right ?

hack/run-e2e-test-operator.sh
```diff
@@ -1,10 +1,8 @@
 #!/bin/bash -xe
 
-${KUBECTL} apply -f ${MANIFESTS_DIR}/namespace.yaml
-
 mkdir -p test_logs/e2e/operator
 unset GOFLAGS && ${OPERATOR_SDK} test local ./test/e2e/operator \
 	--kubeconfig ${KUBECONFIG} \
 	--namespace ${OPERATOR_NAMESPACE} \
-	--up-local \
+	--no-setup \
```
qinqon, line 7: Yep, --up-local was stoopid from my part :-/

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+	v1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	nmstate       nmstatev1beta1.NMState
+	nmstateKey    = types.NamespacedName{Namespace: "nmstate", Name: "nmstate"}
+	dsKey         = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-handler"}
```
qinqon, line 21: handlerKey

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+	deploymentKey = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-operator"}
```
qinqon, line 22: operatorKey

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+)
```
qinqon, line 23: We are missing the webhook deployment here

webhookKey = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-webhook"}

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func installHandler() {
```
qinqon, line 25: installNMState

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func uninstallHandler() {
+	err := framework.Global.Client.Get(context.TODO(), nmstateKey, &nmstate)
+	Expect(err).To(Not(HaveOccurred()))
+	err = framework.Global.Client.Delete(context.TODO(), &nmstate, &client.DeleteOptions{})
+	Expect(err).To(Not(HaveOccurred()))
+}
+
+var _ = Describe("handler install", func() {
```
qinqon, line 39: Describe("NMState operator")

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+	BeforeEach(func() {
+		nmstate = nmstatev1beta1.NMState{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "nmstate",
+				UID:  "12345",
+			},
+		}
+	})
+	Context("when handler not installed", func() {
```
qinqon, line 48: Context("when installed for the first time"), func() {
   BeforeEach(func(){
      installNMState()
})

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+		It("should install", func() {
```
qinqon, line 49: "should deploy handler daemonset and webhook deployment"

We can do two It one for handler and another for webhook but since this is not fast unit test is ok to check both here, otherwise we will have to install/uninstall twice.

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+			installHandler()
+			Eventually(func() string {
+				ds := v1.DaemonSet{}
```
qinqon, line 52: We have to check handler and webhook

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+				_ = framework.Global.Client.Get(context.TODO(), dsKey, &ds)
+				return ds.ObjectMeta.Name
+			}, 60*time.Second, 1*time.Second).ShouldNot(BeEmpty(), fmt.Sprintf("Could not find daemon set for handler"))
```
qinqon, line 55: No need for Sprintf since we are not using any string format.

Humm we usually return a flag and check IsNotFound but this is not bad.

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+		})
+	})
+	Context("when handler is installed", func() {
```
qinqon, line 58: when NMState is installed

Although more verbose at ginkgo this is usually done like this:

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+		It("should ignore a CR of the same name", func() {
```
qinqon, line 59: Maybe I am wrong but k8s does not permit to populate a CR with the same name right ? we don't need this test.

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+			nmstateList := nmstatev1beta1.NMStateList{}
+			err := framework.Global.Client.List(context.TODO(), &nmstateList, &client.ListOptions{})
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(nmstateList.Items).To(HaveLen(1))
+			err = framework.Global.Client.Create(context.TODO(), &nmstate, &framework.CleanupOptions{})
+			Expect(nmstateList.Items).To(HaveLen(1))
+		})
+		It("should ignore a CR of the wrong name", func() {
+			defer func() {
```
qinqon, line 68: Move this to AfterEach

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+	Context("when handler is installed", func() {
+		It("should ignore a CR of the same name", func() {
+			nmstateList := nmstatev1beta1.NMStateList{}
+			err := framework.Global.Client.List(context.TODO(), &nmstateList, &client.ListOptions{})
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(nmstateList.Items).To(HaveLen(1))
+			err = framework.Global.Client.Create(context.TODO(), &nmstate, &framework.CleanupOptions{})
+			Expect(nmstateList.Items).To(HaveLen(1))
+		})
+		It("should ignore a CR of the wrong name", func() {
```
qinqon, line 67:
```golang
Context("when NMState is installed", func() {
   BeforeEach(func() {
     s := v1.DaemonSetList{}
			err := framework.Global.Client.List(context.TODO(), &ds, &client.ListOptions{Namespace: "nmstate"})
			Expect(err).To(Not(HaveOccurred()))
			Expect(ds.Items).To(HaveLen(1))
  )}
   Context("and  the CR is created with a wrong name", func() {
      BeforeEach(func() {
           nmstate.SetName("nmstate-wrong")
          err = framework.Global.Client.Create(...)
	  Expect(nmstateList.Items).To(HaveLen(1))
      }
     AfterEach(func() {
        framework.Global.Client.Delete(context.TODO(), &nmstate, &client.DeleteOptions{})
    })
     It("should ignore it", func() {
         err := framework.Global.Client.List(context.TODO(), &nmstateList, &client.ListOptions{})
	 Expect(err).To(Not(HaveOccurred()))
	 Expect(nmstateList.Items).To(HaveLen(1))
     }) 
}
}
```

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+				framework.Global.Client.Delete(context.TODO(), &nmstate, &client.DeleteOptions{})
+			}()
+			ds := v1.DaemonSetList{}
+			err := framework.Global.Client.List(context.TODO(), &ds, &client.ListOptions{Namespace: "nmstate"})
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(ds.Items).To(HaveLen(1))
+			nmstate.SetName("nmstate-wrong")
+			err = framework.Global.Client.Create(context.TODO(), &nmstate, &framework.CleanupOptions{})
+			Expect(err).To(Not(HaveOccurred()))
+			err = framework.Global.Client.List(context.TODO(), &ds, &client.ListOptions{Namespace: "nmstate"})
```
qinqon, line 79: Check error missing

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+			Expect(ds.Items).To(HaveLen(1))
```
qinqon, line 80: This is suppose to be HaveLen(2) the two CR are there but it's ignored I think, to really check that it's ignored we can add a label or something to the CR at controller or check back handler/webhook.

test/e2e/operator/handler_install_test.go
```diff
@@ -0,0 +1,91 @@
...
+		})
+		It("should uninstall", func() {
```
qinqon, line 82:
```golang
Context("when NMState is uninstalled", func() {
   BeforeEach(func() {
      uninstallNMState()
   })
   It("should uninstall handler and webhook", func() {
      ....
   })
})
```

test/e2e/operator/prepare.go
```diff
@@ -15,6 +15,7 @@ func prepare(t *testing.T) (*framework.TestCtx, string) {
 	cleanupRetryInterval := time.Second * 1
 	cleanupTimeout := time.Second * 5
 	ctx := framework.NewTestCtx(t)
+	defer ctx.Cleanup()
```
qinqon, line 18: AfterSuite so we catch Ctrl-C

cluster/sync-operator.sh
bcrochet: Yes. Good idea. I was actually thinking that as I copied them. :)

cluster/sync-operator.sh
bcrochet: Correct. Good catch.

hack/run-e2e-test-operator.sh
bcrochet: Nah. It worked to get it up and running to begin with.

test/e2e/operator/handler_install_test.go
bcrochet: I like the flag better. Doing it that way.

test/e2e/operator/handler_install_test.go
bcrochet: That's checking the DaemonSetList, not the NMStatesList. That would have 2 CRs. This is just checking that another handler DaemonSet doesn't get deployed.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,110 @@
...
+var (
+	nmstate     nmstatev1beta1.NMState
+	nmstateKey  = types.NamespacedName{Namespace: "nmstate", Name: "nmstate"}
+	handlerKey  = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-handler"}
+	operatorKey = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-operator"}
+	webhookKey  = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-webhook"}
+)
+
+func installNMState() {
+	err := framework.Global.Client.Get(context.TODO(), nmstateKey, &nmstate)
```
phoracek, line 28: Passing the content of nmstate on the side as a global variable is a terrible practice. It is hard to connect all dots and spot potential issue where the variable does not contain what we expect.

Would you please pass both the key and the content as a parameter to this function? Unless you can make those vars into constants, but I'm afraid that's not possible with structs.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,110 @@
...
+func uninstallNMState() {
+	err := framework.Global.Client.Get(context.TODO(), nmstateKey, &nmstate)
+	Expect(err).To(Not(HaveOccurred()))
+	err = framework.Global.Client.Delete(context.TODO(), &nmstate, &client.DeleteOptions{})
+	Expect(err).To(Not(HaveOccurred()))
+}
+
+var _ = Describe("NMState operator", func() {
+	BeforeEach(func() {
+		nmstate = nmstatev1beta1.NMState{
```
phoracek, line 43: Please initialize the variable as a part of the Describe context. To keep it visible and reduce the scope.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	nmstate     nmstatev1beta1.NMState
+	nmstateKey  = types.NamespacedName{Namespace: "nmstate", Name: "nmstate"}
+	handlerKey  = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-handler"}
+	operatorKey = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-operator"}
+	webhookKey  = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-webhook"}
```
phoracek, line 24: This info is only partial. You keep the key here, but you also have to know it is a key of a webhook. Maybe you could define helper functions for these `gets` that would carry both the knowledge of the name and the type?

phoracek:
Great branch name btw.

test/e2e/operator/nmstate_install_test.go
bcrochet: I'm going to add one, and see if I'm understanding what you had in mind. If you like it, I'll do the others.

phoracek: I mean a function ~`getNMStateHandler` that knows the name of the handler and its type, so it just calls the API to obtain it and returns it. Now totally sure it would fit into the rest of testing.

bcrochet: I think I'm grokking what you're saying. Take a look at the latest patch?

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,123 @@
...
+	return ds, err
+}
+
+var _ = Describe("NMState operator", func() {
+	var (
+		nmstate    nmstatev1beta1.NMState
+		webhookKey = types.NamespacedName{Namespace: "nmstate", Name: "nmstate-webhook"}
+	)
+	BeforeEach(func() {
+		nmstate = nmstatev1beta1.NMState{
```
phoracek, line 60: Could you set the NMStateCRKey here directly? Then we can git of the helper get function and make installNMState signature simpler, just pass it nmstate.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,123 @@
...
+	return types.NamespacedName{Namespace: "nmstate", Name: "nmstate"}
+}
+
+func getNMStateOperator() (nmstatev1beta1.NMState, error) {
+	var nmstate nmstatev1beta1.NMState
+	err := framework.Global.Client.Get(context.TODO(), getNMStateCRKey(), &nmstate)
+	return nmstate, err
+}
+
+func getHandlerKey() types.NamespacedName {
```
phoracek, line 44: This is only used in the following function. Can we hardcode it there?

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func uninstallNMState(nmstate nmstatev1beta1.NMState, nmstateKey types.NamespacedName) {
+	_, err := getNMStateOperator()
+	Expect(apierrors.IsNotFound(err)).To(BeFalse(), "NMState CR exists")
+	err = framework.Global.Client.Delete(context.TODO(), &nmstate, &client.DeleteOptions{})
+	Expect(err).ToNot(HaveOccurred(), "NMState CR successfully removed")
+}
+
+func getNMStateCRKey() types.NamespacedName {
```
phoracek, line 34: If we hardcode this name+namspace while building NMState object, we don't need this helper function at all.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,123 @@
...
+	v1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nmstatev1beta1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1beta1"
+)
+
+func installNMState(nmstate nmstatev1beta1.NMState, nmstateKey types.NamespacedName) {
```
phoracek, line 20: I think that the key should be taken directly from nmstate.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,123 @@
...
+	_, err := getNMStateOperator()
+	Expect(apierrors.IsNotFound(err)).To(BeTrue(), "NMState CR does not exist yet")
+	err = framework.Global.Client.Create(context.TODO(), &nmstate, &framework.CleanupOptions{})
+	Expect(err).ToNot(HaveOccurred(), "NMState CR created without error")
+}
+
+func uninstallNMState(nmstate nmstatev1beta1.NMState, nmstateKey types.NamespacedName) {
```
phoracek, line 27: Here we don't need to pass nmstate, do we? The name and namespace alone is enough

cluster/sync-common.sh
```diff
@@ -0,0 +1,94 @@
...
+    while ! $cmd; do
+        sleep $interval
+        timeout=$(( $timeout - $interval ))
+        if [ $timeout -le 0 ]; then
+            return 1
+        fi
+    done
+}
+
+function consistently {
```
yboaron, line 15: Isn't this function always return True no matter what's cmd return value?

cluster/sync-common.sh
```diff
@@ -0,0 +1,94 @@
...
+    timeout=15
```
yboaron, line 16: and maybe we can use for loop here instead of the while?

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+func installNMState(nmstate nmstatev1beta1.NMState) {
+	_, err := getNMStateOperator(nmstate)
+	Expect(apierrors.IsNotFound(err)).To(BeTrue(), "NMState CR does not exist yet")
+	err = framework.Global.Client.Create(context.TODO(), &nmstate, &framework.CleanupOptions{})
+	Expect(err).ToNot(HaveOccurred(), "NMState CR created without error")
+}
+
+func uninstallNMState(nmstate nmstatev1beta1.NMState) {
+	_, err := getNMStateOperator(nmstate)
+	Expect(apierrors.IsNotFound(err)).To(BeFalse(), "NMState CR exists")
```
yboaron, line 29: We should expect no error at all here, right?

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func getNMStateWebhook(webhook v1.Deployment) (v1.Deployment, error) {
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: webhook.ObjectMeta.Namespace, Name: webhook.ObjectMeta.Name}, &webhook)
+	return webhook, err
+}
+
+func getNMStateHandlerList(handler v1.DaemonSet) (v1.DaemonSetList, error) {
+	ds := v1.DaemonSetList{}
+	err := framework.Global.Client.List(context.TODO(), &ds, &client.ListOptions{Namespace: handler.ObjectMeta.Namespace})
```
yboaron, line 51: Does get DsList work for you? it didn't work for me in nmstate controller unit tests

yboaron, line 51: I'm not familiar with List function, does it return an error if there is no DS in this NS?

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	})
+	Context("when installed for the first time", func() {
+		BeforeEach(func() {
+			installNMState(nmstate)
+		})
+		It("should deploy daemonset and webhook deployment", func() {
+			Eventually(func() bool {
+				_, err := getNMStateHandler(handler)
+				return apierrors.IsNotFound(err)
```
yboaron, line 88: What if we got error other than not found? I guess we should consider it as failure

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+			})
+		})
+	})
+	Context("when NMState is uninstalled", func() {
+		BeforeEach(func() {
+			uninstallNMState(nmstate)
+		})
+		It("should uninstall handler and webhook", func() {
+			Eventually(func() bool {
+				_, err := getNMStateHandler(handler)
```
yboaron, line 126: Do we delete only handler DS and webhook deployment when operator CR is removed?
Maybe it worth to add a test that checks: Create nsmatte, delete nmstate, and create nmstate once again, to verify that we don't fail due to leftovers?  but it could be in a follow up patch

test/e2e/operator/nmstate_install_test.go
bcrochet: No. We expect an IsNotFound error. The uninstall happens in the BeforeEach.

test/e2e/operator/nmstate_install_test.go
bcrochet: The tests wouldn't pass if it didn't. The unit tests don't have an actual cluster.

bcrochet: No. It would just return an empty set. That's why where I use it I check for both an error and the expected length.

test/e2e/operator/nmstate_install_test.go
bcrochet: It would just fail the test. We are just asserting here what we do expect.

test/e2e/operator/nmstate_install_test.go
bcrochet: That would definitely be in scope for a follow-up. There are some other resources that should be deleted when the handler is removed.

cluster/sync-common.sh
qinqon: There is `return 1` if we reach timeout, return 1 is an error at bash.

cluster/sync-common.sh
qinqon: Well we are testing `cmd` so feels better have it at the loop condition.

cluster/sync-common.sh
```diff
@@ -0,0 +1,94 @@
...
+        return 1
+    fi
+
+    # Make sure good state is keep for some time
+    if ! consistently isDeploymentOk ${OPERATOR_NAMESPACE} app=kubernetes-nmstate-operator; then
+        echo "Operator is not consistently ready within the given timeout"
+        return 1
+    fi
+}
+
```
qinqon, line 94: This is not common code, this belongs to operator sync

cluster/sync.sh
```diff
@@ -2,35 +2,9 @@
...
-        sleep $interval
-        timeout=$(( $timeout - $interval ))
-        if [ $timeout -le 0 ]; then
-            return 0
-        fi
-    done
-}
+kubectl=./cluster/kubectl.sh
 
 function isDaemonSetOk {
```
qinqon, line 9: We can move this to common, I know we use at one place but is general enough.

cluster/sync.sh
```diff
@@ -2,35 +2,9 @@
...
-    echo "Checking eventually $cmd"
-    while ! $cmd; do
-        sleep $interval
-        timeout=$(( $timeout - $interval ))
-        if [ $timeout -le 0 ]; then
-            return 1
-        fi
-    done
-}
+source ./cluster/sync-common.sh
```
qinqon, line 5: We are going to move deploy_operator to `sync-operator.sh` and since we are deployting handler and operator here we have to source `sync-operator.sh`

To do so we have to be sure that sync-operator.sh does not run stuff if we source it only if we run it.

test/e2e/operator/main_test.go
```diff
@@ -33,7 +34,11 @@ var _ = BeforeSuite(func() {
 
 	err := framework.AddToFrameworkScheme(apis.AddToScheme, nmstateList)
 	Expect(err).ToNot(HaveOccurred())
-	prepare(t)
+	ctx, _ = prepare(t)
+})
+
+var _ = AfterSuite(func() {
```
qinqon, line 40: There is something that we miss at e2e test, It's not related to this PR but can you check if "Ctrl-C" get catch by ginkgo test framework ? if it's not can you do some investigation why ?

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nmstatev1beta1 "github.com/nmstate/kubernetes-nmstate/pkg/apis/nmstate/v1beta1"
+)
+
+func installNMState(nmstate nmstatev1beta1.NMState) {
+	_, err := getNMStateOperator(nmstate)
+	Expect(apierrors.IsNotFound(err)).To(BeTrue(), "NMState CR does not exist yet")
```
qinqon, line 22: Let'ss call `uninstallNMState` here and then we just call `.Create`

test/e2e/operator/nmstate_install_test.go
qinqon, line 29: You don't need to check existence to call delete, just call delete and if error is NotFound it returns also checking that it's really delted and Eventually helps too so we block until etcd delete it.

Example here:
https://github.com/nmstate/kubernetes-nmstate/blob/da83e5924b6348e8c506e157f83ec169171206cb/test/e2e/handler/utils.go#L167-L177

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func uninstallNMState(nmstate nmstatev1beta1.NMState) {
+	_, err := getNMStateOperator(nmstate)
+	Expect(apierrors.IsNotFound(err)).To(BeFalse(), "NMState CR exists")
+	err = framework.Global.Client.Delete(context.TODO(), &nmstate, &client.DeleteOptions{})
+	Expect(err).ToNot(HaveOccurred(), "NMState CR successfully removed")
+}
+
+func getNMStateOperator(nmstate nmstatev1beta1.NMState) (nmstatev1beta1.NMState, error) {
```
qinqon, line 34: This is the `getNMStateCR` also don't pass any argument just return the default CR here and call it `getDefaultNMStateCR`

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: nmstate.ObjectMeta.Namespace, Name: nmstate.ObjectMeta.Name}, &nmstate)
+	return nmstate, err
+}
+
+func getNMStateHandler(handler v1.DaemonSet) (v1.DaemonSet, error) {
```
qinqon, line 39: There is nothing related to NMStateHandler here this just returns a DaemonSet with a daemonset, call this `func getDefaultNMStateHandler()` and don't pass arguments use directly the default nmstate namspace/name.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: handler.ObjectMeta.Namespace, Name: handler.ObjectMeta.Name}, &handler)
+	return handler, err
+}
+
+func getNMStateWebhook(webhook v1.Deployment) (v1.Deployment, error) {
```
qinqon, line 44: ditto

test/e2e/operator/nmstate_install_test.go
qinqon: This prevents from runing this tests at HCO for example since w are going to have multiple DaemonSet at "hyper-converged" you have to filter by label `component=kubernetes-nmtate-handler`

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+var _ = Describe("NMState operator", func() {
+	var (
+		nmstate nmstatev1beta1.NMState
+		handler v1.DaemonSet
+		webhook v1.Deployment
+	)
+	BeforeEach(func() {
+		nmstate = nmstatev1beta1.NMState{
```
qinqon, line 62: Let's just us `getDefault` functions is going to be like that 90% of use cases and for the other 10% put in place a some get functions with custom names.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+		webhook = v1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "nmstate-webhook",
+				Namespace: "nmstate",
+			},
+		}
+	})
+	Context("when installed for the first time", func() {
+		BeforeEach(func() {
+			installNMState(nmstate)
```
qinqon, line 83: Add a `installDefaultNmstate` that calls installNMstate with default names as said is toing ot be like that 90% of the time at tests, so we re-use test code and it's verbose enough.

cluster/sync-common.sh
```diff
@@ -0,0 +1,94 @@
...
+    interval=5
+    cmd=$@
+    echo "Checking consistently $cmd"
+    while $cmd; do
+        sleep $interval
+        timeout=$(( $timeout - $interval ))
+        if [ $timeout -le 0 ]; then
+            return 0
+        fi
+    done
```
yboaron, line 26: Don't you need to add after the 'done' return 1?  in current code, this function will always return OK

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,135 @@
...
+		})
+		It("should deploy daemonset and webhook deployment", func() {
+			Eventually(func() bool {
```
qinqon, line 86: We need some infra functions like:
`deploymenIsReady` and `deploymentIsReadyEventually`
`daemonSetIsReady` and `daemonSetIsReadyEventually`

Then we can pass Default resources

`daemonSetIsReadyEventually(DefaultNMStateHandler)`
`deploymentIsReadyEventually(DefaultNMStateWebhook)`

cluster/sync-common.sh
yboaron: @qinqon  I meant to consistently function, I guess we want to return 1  if $cmd failed.
IIUC, in current code 'consistently' function will return OK (0)

yboaron: @qinqon I meant to consistently function, seems it will always return OK (0) even if $cmd failed

cluster/sync-common.sh
bcrochet: It's used in both syncs.

test/e2e/operator/main_test.go
bcrochet: It appears to... 

^Cmake: *** [Makefile:144: test-e2e-operator] Interrupt                                                               
                                                                                                                      
---------------------------------------------------------                                                             
Received interrupt.  Running AfterSuite...                                                                            
^C again to terminate immediately                                                                                     
--- FAIL: TestE2E (3.83s)

test/e2e/operator/nmstate_install_test.go
bcrochet: Not sure I like the forcing of uninstall there... I might rather go with removing the assertion, and having it go straight to the Create, and leave the state checking to the caller.

test/e2e/operator/nmstate_install_test.go
qinqon: I have an idea to make test kind of readable and re-use as much as possible for Deployments and DaemonSets

The idea is to write the tests like this 

```golang
Deployment(DefaultNMWebhook).ReadyConditionEventually().Should(BeTrue())
```

for that we we need the following golang:

```golang
type DeploymentCommands struct {deployement appsv1.Deployment}

func (*DeploymentCommands) ReadyConditionEventually() AsyncMatcher {
    return Eventually(func() {
       // return here the Ready condition of Deployement
    })
}

func Deployement(deploymentKey types.NamespacedName) DeploymentCommands {
// Retrieve the deployement (using eventually to wait until it exists or fail)
    return DeploymentCommands{ deployment: deployment }
}

```
Also we can put this functions at test/e2e/deployment and test/e2e/daemonset or at test/e2e/app, or better name.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,167 @@
...
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: nmstate.ObjectMeta.Namespace, Name: nmstate.ObjectMeta.Name}, &nmstate)
+	return nmstate, err
+}
+
+func getDefaultNMStateOperator() (nmstatev1beta1.NMState, error) {
+	return getNMStateOperator(getDefaultNMState())
+}
+
+func getNMStateHandler(handler v1.DaemonSet) (v1.DaemonSet, error) {
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: handler.ObjectMeta.Namespace, Name: handler.ObjectMeta.Name}, &handler)
```
qinqon, line 142: call this `getDaemonSet` and remove all reference to handler since this is generic, move this to test/e2e/daemonset/

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,167 @@
...
+	return handler, err
+}
+
+func getDefaultNMStateHandler() (v1.DaemonSet, error) {
+	return getNMStateHandler(getDefaultHandler())
+}
+
+func getNMStateWebhook(webhook v1.Deployment) (v1.Deployment, error) {
```
qinqon, line 150: call this getDeployment and remove all reference to handler since this is generic, move this to test/e2e/deployment/

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,167 @@
...
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: webhook.ObjectMeta.Namespace, Name: webhook.ObjectMeta.Name}, &webhook)
+	return webhook, err
+}
+
+func getDefaultNMStateWebhook() (v1.Deployment, error) {
+	return getNMStateWebhook(getDefaultWebhook())
+}
+
+func getNMStateHandlerList(handler v1.DaemonSet) (v1.DaemonSetList, error) {
```
qinqon, line 159: This should be `getDamonSetList(filteringLabels map[string]string)` move to test/e2e/daemonset

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,43 @@
...
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type DaemonSetCommands struct{ daemonSet appsv1.DaemonSet }
+
+func (dc DaemonSetCommands) ReadyConditionEventually() AsyncAssertion {
```
qinqon, line 17: We have suffer a lot of not doing proper asserting at Ginkgo and hidden matching details, so is very difficult to debug errors, so let's prepare a custom matcher so we can log all we need on failure and plug it nicely with expectations.

So at matchers.go under this `daemonset` package we can add the following matcher to check daemonsetReadinness:

```golang
package daemonset                                                               
                                                                                
import (                                                                        
    "fmt"                                                                       
                                                                                
    "github.com/onsi/gomega/format"                                             
    "github.com/onsi/gomega/types"                                              
                                                                                
    appsv1 "k8s.io/api/apps/v1"                                                 
)                                                                               
                                                                                
func BeReady() types.GomegaMatcher {                                            
    return &isReady{}                                                           
}                                                                               
                                                                                
type isReady struct {                                                           
    obtainedDaemonSet *appsv1.DaemonSet                                         
}                                                                               
                                                                                
func (matcher *isReady) Match(obtained interface{}) (success bool, err error) { 
    obtainedDaemonSet, ok := obtained.(*appsv1.DaemonSet)                       
    if !ok {                                                                    
        return false, fmt.Errorf("daemonset.IsReady matcher expects an appvs1.DaemonSet")
    }                                                                           
    matcher.obtainedDaemonSet = obtainedDaemonSet                               
    return matcher.expectedNumberOfPods() == matcher.availableNumberOfPods(), nil
}                                                                               
                                                                                
func (matcher *isReady) FailureMessage(actual interface{}) (message string) {   
    return matcher.message("to equal")                                          
}                                                                               
                                                                                
func (matcher *isReady) NegatedFailureMessage(actual interface{}) (message string) {
    return matcher.message("to not equal")                                      
}                                                                               
                                                                                
func (matcher *isReady) message(message string) string {                        
    return format.Message(matcher.expectedNumberOfPods(), fmt.Sprintf("daemonset.Status.DesiredNumberScheduled %s daemonset.Status.NumberAvailable", message), matcher.availableNumberOfPods())
}                                                                               
                                                                                
func (matcher *isReady) expectedNumberOfPods() int32 {                          
    return matcher.obtainedDaemonSet.Status.DesiredNumberScheduled              
}                                                                               
                                                                                
func (matcher *isReady) availableNumberOfPods() int32 {                         
    return matcher.obtainedDaemonSet.Status.NumberAvailable                     
}                                                                               
```

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,43 @@
...
+	return Eventually(func() bool {
+		return dc.daemonSet.Status.DesiredNumberScheduled == dc.daemonSet.Status.NumberAvailable
+	}, 60*time.Second, 1*time.Second)
+}
+
+func DaemonSet(daemonSetKey types.NamespacedName) DaemonSetCommands {
```
qinqon, line 23: Let's rename it `Eventually` and return `AsyncAssertion` and since we have a custom matcher we just use the matcher with the `AsyncAssertion` so users will do:

```golang
daemonset.Eventually(DefaultNMStateHandler).Should(daemonset.BeReady())
```

This way the daemonset will be re-ready if `BeReady` fails and also we can negate it without extra code

```golang
daemonset.Eventually(DefaultNMStateHandler).ShouldNot(daemonset.BeReady())
```
This this in mind this functions should be:

```golang
func Eventually(daemonSetKey types.NamespacedName) AsyncAssertion {
return Eventually(func() appsv1.DaemonSet {
               daemonSet := appsv1.Daemonset{}
		_ = framework.Global.Client.Get(context.TODO(), daemonSetKey, &daemonSet)
		return daemonSet
}, 60*time.Second, 1*time.Second)
```

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,43 @@
...
+
+// GetDaemonSetList returns a DaemonSetList matching the labels passed
+func GetDaemonSetList(filteringLabels map[string]string) (appsv1.DaemonSetList, error) {
+	ds := appsv1.DaemonSetList{}
+	err := framework.Global.Client.List(context.TODO(), &ds, &client.ListOptions{LabelSelector: labels.SelectorFromSet(filteringLabels)})
+	return ds, err
+}
+
+// GetDaemonSet returns a DaemonSet matching the passed in DaemonSet name and namespace
+func GetDaemonSet(daemonSet appsv1.DaemonSet) (appsv1.DaemonSet, error) {
```
qinqon, line 40: pass as argument the key with not the daemonset with `types.NamespacedName`

test/e2e/deployment/deployment.go
```diff
@@ -0,0 +1,41 @@
...
+
+	. "github.com/onsi/gomega"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	deputils "k8s.io/kubectl/pkg/util/deployment"
+)
+
+type DeploymentCommands struct{ deployment appsv1.Deployment }
```
qinqon, line 15: Do same as DeamonSet here creating a custom matcher and storing it at `matcher.go`

Users should do:

```golang
deployment.Eventually(DefaultNMStateWebhook).Should(deployment.BeReady())
deployment.Eventually(DefaultNMStateWebhook).ShouldNot(deployment.BeReady())
```

test/e2e/deployment/deployment.go
```diff
@@ -0,0 +1,41 @@
...
+	deployment := appsv1.Deployment{}
+	Eventually(func() appsv1.Deployment {
+		_ = framework.Global.Client.Get(context.TODO(), deploymentKey, &deployment)
+		return deployment
+	}, 60*time.Second, 1*time.Second).ShouldNot(BeNil())
+	return DeploymentCommands{deployment: deployment}
+}
+
+// GetDeployment returns a deployment matching passing in deployment Name and Namespace
+func GetDeployment(deployment appsv1.Deployment) (appsv1.Deployment, error) {
```
qinqon, line 38: Pass key instead of deployment

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+var _ = Describe("NMState operator", func() {
+	Context("when installed for the first time", func() {
+		BeforeEach(func() {
+			installDefaultNMState()
+		})
+		It("should deploy daemonset and webhook deployment", func() {
+			daemonset.DaemonSet(getDefaultNMStateHandlerKey()).ReadyConditionEventually().Should(BeTrue())
+			deployment.Deployment(getDefaultNMStateWebhookKey()).ReadyConditionEventually().Should(BeTrue())
+		})
```
qinqon, line 30: We should uninstall on AfterEach here.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+	Context("when NMState is installed", func() {
+		BeforeEach(func() {
+			ds, err := getDefaultNMStateHandlerList()
+			Expect(err).ToNot(HaveOccurred(), "List daemon sets in namespace nmstate succeeds")
+			Expect(ds.Items).To(HaveLen(1), "and has only one item")
+		})
```
qinqon, line 37: This is a `It("should list one NMState CR")`

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+		Context("and the CR is created with a wrong name", func() {
+			var nmstate = getNMStateWrongName()
```
qinqon, line 39: Just use `getNMStateWrongName` at `BeforeEach` and `AfterEach` so it's clear there what is happening

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+				_, err := getDefaultNMStateHandler()
+				return apierrors.IsNotFound(err)
+			}, 60*time.Second, 1*time.Second).Should(BeTrue(), fmt.Sprint("Daemon set for NMState should be removed, but is not"))
+			Eventually(func() bool {
+				_, err := getDefaultNMStateWebhook()
+				return apierrors.IsNotFound(err)
+			}, 60*time.Second, 1*time.Second).Should(BeTrue(), fmt.Sprint("Deployment for NMState webhook should be removed, but is not"))
+		})
+	})
+})
```
qinqon, line 70: Move this to `AfterEach`of the parent `Context` so we just check there that handler and webhook are gone, I know is better to have `It` but is more convenient to have the AfterEach.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func getDefaultNMState() nmstatev1beta1.NMState {
```
qinqon, line 72: Don't use a function for this this is a constant

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	return nmstatev1beta1.NMState{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "nmstate",
+			Namespace: "nmstate",
+		},
+	}
+}
+
+func getDefaultHandler() v1.DaemonSet {
```
qinqon, line 81: ditto

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	return v1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "nmstate-handler",
+			Namespace: "nmstate",
+		},
+	}
+}
+
+func getDefaultWebhook() v1.Deployment {
```
qinqon, line 90: ditto

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	return v1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "nmstate-webhook",
+			Namespace: "nmstate",
+		},
+	}
+}
+
+func getDefaultNMStateWebhookKey() types.NamespacedName {
```
qinqon, line 99: ditto

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	return types.NamespacedName{Namespace: "nmstate", Name: "nmstate-webhook"}
+}
+
+func getDefaultNMStateHandlerKey() types.NamespacedName {
```
qinqon, line 103: ditto

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	return types.NamespacedName{Namespace: "nmstate", Name: "nmstate-handler"}
+}
+
+func getDefaultHandlerLabels() map[string]string {
```
qinqon, line 107: ditto

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	return map[string]string{"component": "kubernetes-nmstate-handler"}
+}
+
+func getNMStateWrongName() nmstatev1beta1.NMState {
```
qinqon, line 111: ditto

The reuse is minimal

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	nmstate := getDefaultNMState()
+	nmstate.ObjectMeta.Name = "nmstate-wrong"
+	return nmstate
+}
+
+func installNMState(nmstate nmstatev1beta1.NMState) {
+	_, err := getNMStateOperator(nmstate)
+	Expect(apierrors.IsNotFound(err)).To(BeTrue(), "NMState CR does not exist yet")
+	err = framework.Global.Client.Create(context.TODO(), &nmstate, &framework.CleanupOptions{})
```
qinqon, line 120: Just create it fails with expected error if already exists.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+	Expect(err).ToNot(HaveOccurred(), "NMState CR created without error")
+}
+
+func installDefaultNMState() {
+	installNMState(getDefaultNMState())
+}
+
+func uninstallNMState(nmstate nmstatev1beta1.NMState) {
+	_, err := getNMStateOperator(nmstate)
```
qinqon, line 129: Do we want to fail at uninstall if is not there ? I suppose we do.

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,158 @@
...
+func uninstallDefaultNMState() {
+	uninstallNMState(getDefaultNMState())
+}
+
+func getNMStateOperator(nmstate nmstatev1beta1.NMState) (nmstatev1beta1.NMState, error) {
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Namespace: nmstate.ObjectMeta.Namespace, Name: nmstate.ObjectMeta.Name}, &nmstate)
+	return nmstate, err
+}
+
+func getDefaultNMStateOperator() (nmstatev1beta1.NMState, error) {
```
qinqon, line 144: This Default business is only needed for the CR please for Operator, Webhook and Handler remove the Default concept.

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,34 @@
+package daemonset
+
+import (
+	"context"
+	"time"
+
+	"github.com/onsi/gomega"
```
qinqon, line 7: The special `gomega` matchers are imported as :

```
. "github.com/onsi/gomega"
```
So we do `Eventually` instead of `gomega.Eventually`, if the problem is that we cannot declare the new `Eventualy` let's call it `daemonset.GetEventually` in fact it make more sense that `daemonset.Eventually`

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,34 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Eventually(daemonSetKey types.NamespacedName) gomega.AsyncAssertion {
+	return gomega.Eventually(func() appsv1.DaemonSet {
+		daemonSet := appsv1.DaemonSet{}
+		_ = framework.Global.Client.Get(context.TODO(), daemonSetKey, &daemonSet)
```
qinqon, line 18: We can use `GetDaemonSet` here.

About error handling, if we make the function return daemonset and error Eventually will fail if there is an error, from gomega doc https://onsi.github.io/gomega/#eventually:

> The function that you pass to Eventually can have more than one return value. In that case, Eventually passes the first return value to the matcher and asserts that all other return values are nil or zero-valued. This allows you to use Eventually with functions that return a value and an error – a common pattern in Go

test/e2e/daemonset/daemonset.go
qinqon: @bcrochet are you ok with this ?

test/e2e/deployment/deployment.go
qinqon: @bcrochet are you ok doing the Deployment matcher ?

test/e2e/daemonset/daemonset.go
bcrochet: That's done.

test/e2e/deployment/deployment.go
bcrochet: Yes. Should be done.

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,35 @@
...
+		return daemonSet
+	}, 180*time.Second, 1*time.Second)
+}
+
+// GetDaemonSetList returns a DaemonSetList matching the labels passed
+func GetDaemonSetList(filteringLabels map[string]string) (appsv1.DaemonSetList, error) {
```
qinqon, line 24: we can name this `GetList` since package si daemonset

test/e2e/daemonset/daemonset.go
```diff
@@ -0,0 +1,35 @@
...
+	ds := appsv1.DaemonSetList{}
+	err := framework.Global.Client.List(context.TODO(), &ds, &client.ListOptions{LabelSelector: labels.SelectorFromSet(filteringLabels)})
+	return ds, err
+}
+
+// GetDaemonSet returns a DaemonSet matching the passed in DaemonSet name and namespace
+func GetDaemonSet(daemonSetKey types.NamespacedName) (appsv1.DaemonSet, error) {
```
qinqon, line 31: we can name this `Get`

test/e2e/deployment/deployment.go
```diff
@@ -0,0 +1,50 @@
...
+func (dc DeploymentCommands) ReadyConditionEventually() gomega.AsyncAssertion {
+	return gomega.Eventually(func() bool {
+		cond := deputils.GetDeploymentCondition(dc.deployment.Status, appsv1.DeploymentAvailable)
+		if cond == nil {
+			return false
+		}
+		return cond.Status == corev1.ConditionTrue
+	}, 120*time.Second, 1*time.Second)
+}
+
```
qinqon, line 26: This is not longer needed since we have the matcher, let's do it like the DamonSet

test/e2e/operator/nmstate_install_test.go
```diff
@@ -0,0 +1,124 @@
...
+				ds, err := daemonset.GetDaemonSetList(getDefaultHandlerLabels())
+				Expect(err).ToNot(HaveOccurred(), "Daemon set list is retreieved without error")
+				Expect(ds.Items).To(HaveLen(1), "and still only has one item")
+			})
+		})
+		AfterEach(func() {
+			uninstallDefaultNMState()
+		})
+	})
+	Context("when NMState is uninstalled", func() {
```
qinqon, line 64: If you move this `Context` under `Context("when NMState is installed", func() {` rename this as 

```golang
Context("and uninstalled", func() {
```

Then you don't need to install here again and Ginkgo will compose exactly what is doing at the test name.

test/e2e/daemonset/daemonset.go
bcrochet: So, yes, but we can just use it without the '.' here and then use daemonset.Eventually where it is used. Nothing says it *has* to be imported with '.', that's just a convenience. I'm fine with it as it is.

test/e2e/daemonset/daemonset.go
bcrochet: Like the thing about error. Done.

PR #551, opened 2020-05-27
Use k8s-1.18 at kubevirtci

qinqon:

/kind infrastructure

Since kubevirtci has the provider for it let's use it, it will help us
see if there issues at ocp-4.5 since it's based on 1.18.

**Release note**:

```release-note
Using k8s-1.18 at kubevirtci
```


test/e2e/handler/simple_bridge_and_bond_test.go
```diff
@@ -147,7 +147,7 @@ var _ = Describe("NodeNetworkState", func() {
 				}
 				resetDesiredStateForNodes()
 			})
-			It("should have the linux bridge at currentState", func() {
+			FIt("should have the linux bridge at currentState", func() {
```
phoracek, line 150: Bad!

test/e2e/handler/utils.go
```diff
@@ -327,16 +327,18 @@ func getVLANFlagsEventually(node string, connection string, vlan int) AsyncAsser
...
 			// [1] https://bugs.centos.org/view.php?id=16533
 			output, err := cmd.Run("test/e2e/get-bridge-vlans-flags-el8.sh", false, node, connection, strconv.Itoa(vlan))
 			Expect(err).ToNot(HaveOccurred())
 			return strings.Split(string(output), " ")
 		} else {
+			By("Getting vlan filtering from json output")
 			parsedBridgeVlans := gjson.Parse(bridgeVlans)
 
-			vlanFlagsFilter := fmt.Sprintf("%s.#(vlan==%d).flags", connection, vlan)
+			vlanFlagsFilter := fmt.Sprintf("#(ifname==%s).vlans.#(vlan==%d).flags", connection, vlan)
```
phoracek, line 341: This change has nothing to do with k8s, it is related to the version of `bridge` on the host. Are we breaking our compatibility with the previous version by doing this? Can we keep the compatibility somehow?

qinqon: I want to check if we can get the same information with nmcli so we don't depend on this output.

qinqon: I think we can keep previous compatibility somehow just checking if we have `ifname` string, let me do it.

PR #555, opened 2020-05-28
Use kubevirt junit and polarion ginkgo reporters

qinqon:
/kind enhancement

It add the kubevirt polarion test reporter and also is switch junit reporter to kubevirt one so we can specify the dump path of xml.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-k8s.sh
+++ b/automation/check-patch.e2e-k8s.sh
@@ -9,7 +9,6 @@
 
 teardown() {
     make cluster-down
-    cp $(find . -name "*junit*.xml") $ARTIFACTS
     # Don't fail if there is no logs
     cp ${E2E_LOGS}/handler/*.log ${ARTIFACTS} || true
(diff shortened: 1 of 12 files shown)


automation/check-patch.e2e-k8s.sh
```diff
@@ -9,7 +9,6 @@
 
 teardown() {
     make cluster-down
-    cp $(find . -name "*junit*.xml") $ARTIFACTS
```
phoracek, line 12: Why are we dropping this?

qinqon: With the kubevirt.io junit reporter we can specify the path directly [1]  and there we just use $ARTIFACTS so no need for cp

[1] https://github.com/nmstate/kubernetes-nmstate/pull/555/files#diff-2da872edd4c334f4c06df2de5bf629c7R32

PR #564, opened 2020-05-29
Bump operator-sdk-0.18.0

qinqon:

>
> /kind bug
> /kind enhancement

Let's bump to operator-sdk 0.18.0 to make easy k8s-1.18 integration

**Release note**:

```release-note
Bumped operator-sdk to 1.18.0
```


qinqon:
I think we are hit by PreserveUnknownFields [1] removed from CRDs after the operator-sdk since CRDs are v1 now, I think it remove the fields under CurrentState since they are unknown to the CRD.

https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/gen.go#L54

PR #565, opened 2020-06-05
Tag release

qinqon:

/kind enhancement
/kind infrastructure

The current kubernetes-nmstate release process is driven by a PR generated with the
`make prepare-[major|minor|patch]` targets, when this PR is merged prow detects a change
at version/version.go and triggers the release job.
    
To reduce manual steps this commit replaces `make prepare-[major|minor|patch]` targets with `make version-[major|minor|patch]`,
it will tag code with bumped version and push to repository, then a prow job will detect new tag and trigger the release job,
there release job just call release.sh and this modify that so it generate new release from previous tag.
    
**Release note**:

```release-note
Replaced PR driven release pipeline with tag driven.
```


hack/version.sh
```diff
@@ -1,9 +1,7 @@
...
-else
-    sed -i "s/= \".*\"$/= \"$1\"/g" $version_file
-fi
+#!/bin/bash
+
+set -e
+
+current_branch=$(git branch --show-current)
+versions=($(git tag --sort version:refname --merged qinqon/$current_branch |grep "^v[0-9]*\.[0-9]*\.[0-9]*$"))
+echo ${versions[$1]}
```
phoracek, line 7: drop qinqon

hack/release-notes.tmpl
```diff
@@ -1,3 +1,5 @@
+{{ .CurrentRevision}}
```
phoracek, line 1: This is redundant imo. It is already a part of the headline (the tag). But since this is only keeping the existing behavior, I don't mind keeping it and changing min a different PR. Not a big deal

hack/release.sh
```diff
@@ -2,7 +2,8 @@
 
 set -xe
 
-tag=$(hack/version.sh)
+old_version=$(hack/version.sh -2)
```
phoracek, line 5: Maybe we should rename the script to versions.

PR #588, opened 2020-07-29
build/Dockerfile.operator.openshift: fix COPY path

sosiouxme:
/kind bug

The downstream build that uses this file would fail because the path the binary was built in the first stage does not match the path the second stage tried to copy it from.

**Release note**:
NONE


qinqon:
@sosiouxme  can you add the Signed-off-by to the commit message with `git commit -s --amend`

sosiouxme:
closing this for now - will probably fix in downstream fork (or reopen if not).

PR #605, opened 2020-10-01
Remove installing nm 1.22 kubevircti already provides it

qinqon:
The k8s-1.19 kubevirtci provider has bump nm to 1.22 so kubernetes-nmstate
does not need to install it anymore.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


oshoval (commented):
what about < k8s-1.19 ?
btw does 1.19 have 1.22 or 1.26?
asking because when i install using those snippet manually i saw it updates to 1.26
is it ok if its 1.26 ?

qinqon:
@oshoval we are aligned with nmstate 0.2 and that's NM 1.22 with current centos8 we can even install 1.26 but we have to be super carefull here, about k8s 1.19 this is the current target version at kubevirt (although is wip there)

phoracek (approved):
/lgtm
/approve

PR #608, opened 2020-10-08
operator-sdk-1.y.z: Move from 'pkg/apis' to 'api' and 'pkg/controller…

qinqon:
…' to 'controllers'

Migrating to operator-sdk 1.y.z means adapt code to kubebuilder structure that means that
api code is at `api/` dir and controllers at at `controller/`

We cannot use `operator-sdk gen crds` or `operator-sdk gen k8s` since it expect the old structure,
in this commit we replace those with `controller-gen` tool used by kubebuilder, it expect some changes
at _types.go labels reflected at this commit and also the CRDs yamls does not have the suffix _crd.

Also discovered that openapi was missing the shared type, it's has being added here.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


qinqon:
Add changes from `gofmt`

qinqon:
Moved `cmd/manager/main.go` to `main.go`

qinqon:
Adapt all the Dockerfiles so now they compile `./main.go` into binary `manager` also removed multistage for operator Dockerfile.

qinqon:
Fix `Deployment` and `DaemonSet` command argument and replace `kubernetes-nmstate` with `manager` since this is the new binary name.

qinqon:
This PR loses the history of CRDs but I am possitive that `git mv` was used :-(

qinqon:
/testing CNAO here https://github.com/kubevirt/cluster-network-addons-operator/pull/597

qinqon:
Moved `api/nmstate/*` to `api/*` since we have only one api group `nmstate`

qinqon:
Call `make generate` again.

Makefile
```diff
@@ -66,26 +67,29 @@ LOCAL_REGISTRY ?= registry:5000
...
 vet: $(GO)
-	$(GO) vet ./cmd/... ./pkg/... ./test/...
+	$(GO) vet ./...
 
 whitespace-format:
 	hack/whitespace.sh format
 
 gofmt: $(GO)
-	$(GOFMT) -w cmd/ pkg/ test/e2e/
+	$(GOFMT) -w *.go test/ hack/ api/ controllers/ pkg/
```
phoracek, line 86: There is no *.go file in the root, is there?

Makefile
```diff
@@ -66,26 +67,29 @@ LOCAL_REGISTRY ?= registry:5000
...
 
 whitespace-check:
 	hack/whitespace.sh check
 
 gofmt-check: $(GO)
-	test -z "`$(GOFMT) -l cmd/ pkg/ test/e2e/`" || ($(GOFMT) -l cmd/ pkg/ test/e2e/ && exit 1)
+	test -z "`$(GOFMT) -l *.go test/ hack/ api/ controllers/ pkg/`" || ($(GOFMT) -l *.go test/ hack/ api/ controllers/ pkg/ && exit 1)
```
phoracek, line 92: ditto

phoracek (changes requested):
Just a nit but overall great job for making this reviewable! Does not look bad at all.

Makefile
qinqon: main.go is not at root level

Makefile
qinqon: https://github.com/nmstate/kubernetes-nmstate/pull/608#discussion_r504062906

Makefile
phoracek: So why do you run *.go here? Should not we target the directory that has main.go?

phoracek: Ahaaa, I missed that. Thanks

phoracek:
/lgtm
/approve

Let's break things.

PR #616, opened 2020-10-21
Bump controller-gen to v0.4.0 to get apiext/v1 generated CRDs

qinqon:

>
/kind bug
> /kind enhancement

After merging https://github.com/nmstate/kubernetes-nmstate/pull/608 CRDs are back to
apiextensions.k8s.io/v1beta1, we have to upgrade controller-tools so they get generated
with apiextensions.k8s.io/v1.

**Release note**:

```release-note
NONE
```


RamLavi:
can you split the auto generated stuff to a different commit?

qinqon:
@RamLavi done

api/shared/state_types.go
```diff
@@ -9,6 +9,7 @@ type RawState []byte
 //
 // [1] https://github.com/nmstate/nmstate/blob/master/libnmstate/schemas/operational-state.yaml
 // +k8s:openapi-gen=true
```
phoracek, line 11: You dropped openapi targets in this PR. Does it make sense to keep these comments?

phoracek (changes requested):

api/shared/state_types.go
qinqon: Good catch, removing.

phoracek (approved):
/lgtm
/approve

PR #622, opened 2020-10-27
Roll out configuration sequentially

rhrazdil:

>
> /kind bug

/kind enhancement

With this PR NNCPs are by default applied sequentially on cluster nodes. 
A new field is added to NNCP Spec: `Parallel`, which can be set to true
to roll out configuration in parallel, which was the previous behaviour.

Additionaly there are added two fields to NNCP Status: `NodeRunningUpdate`  and `NodeUpdateStart`, which contain node name and timestamp when a node started with configuration.

The `NodeRunningUpdate` serves as a mutex, nodes that fail to claim this field requeue the reconcile request and try again later.
When a node owning this field hasn't released it for longer than 5 minutes, other nodes assume that there is an issue and set their enactment state to `FailedToConfigure` (and don't attempt to retry anymore).

TODOs:
This PR introduces only basic functionality of the sequential rollout, there are several gaps that need to be covered in followup PRs:
- Webhook admitter, that prevents users to edit or create new NNCP, when existing NNCP is still in progress
- New enactment state `aborted`. In this PR, when a node doesn't manage to apply configuration in time, rest of the nodes set their state to `FailedToConfigure`, which is not accurate, because technically, they haven't tried. Instead, `aborted` state should be used to accurately inform user about the situation.
- There have been suggestions to allow user to define a number of nodes that are allowed to apply NNCP in parallel.
  We could instead of `Parallel` introduce something like a chunkSize, and instead of `nodeRunningUpdate` something like `nodeUpdateCapacity`which would effectively act as a semaphore instead of a mutex

**Release note**:

```release-note
Roll out policy configuration sequentially by default.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ export KUBEVIRT_PROVIDER ?= k8s-1.19
 export KUBEVIRT_NUM_NODES ?= 2 # 1 master, 1 worker needed for e2e tests
 export KUBEVIRT_NUM_SECONDARY_NICS ?= 2
 
-export E2E_TEST_TIMEOUT ?= 40m
+export E2E_TEST_TIMEOUT ?= 60m
 
 e2e_test_args = -v -timeout=$(E2E_TEST_TIMEOUT) -slowSpecThreshold=60 $(E2E_TEST_ARGS)
 
--- a/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -16,7 +16,6 @@ type NodeNetworkConfigurationPolicySpec struct {
 	// The desired configuration of the policy
 	DesiredState State `json:"desiredState,omitempty"`
 
-	// +kubebuilder:validation:XPreserveUnknownFields
 	// When set to true, changes are applied to all nodes in parallel
 	// +optional
 	Parallel bool `json:"parallel,omitempty"`
@@ -26,11 +25,11 @@ type NodeNetworkConfigurationPolicySpec struct {
 type NodeNetworkConfigurationPolicyStatus struct {
 	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
 
-	// NodeRunningUpdate field is used for serializing cluster nodes configuration
+	// NodeRunningUpdate field is used for serializing cluster nodes configuration when Parallel flag is false
 	// +optional
 	NodeRunningUpdate string `json:"nodeRunningUpdate,omitempty" optional:"true"`
 
-	// NodeUpdateStart marks starting time of a node on a policy configuration
+	// NodeUpdateStart marks starting time of a node on a policy configuration when Parallel flag is false
 	// +optional
 	NodeUpdateStart *metav1.Time `json:"nodeUpdateStart,omitempty" optional:"true"`
 }
--- a/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -1,5 +1,9 @@
 package shared
 
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
 // NodeNetworkConfigurationPolicySpec defines the desired state of NodeNetworkConfigurationPolicy
 type NodeNetworkConfigurationPolicySpec struct {
 	// NodeSelector is a selector which must be true for the policy to be applied to the node.
@@ -11,11 +15,24 @@ type NodeNetworkConfigurationPolicySpec struct {
 	// +kubebuilder:validation:XPreserveUnknownFields
 	// The desired configuration of the policy
 	DesiredState State `json:"desiredState,omitempty"`
+
+	// +kubebuilder:validation:XPreserveUnknownFields
+	// When set to true, changes are applied to all nodes in parallel
+	// +optional
+	Parallel bool `json:"parallel,omitempty"`
 }
 
 // NodeNetworkConfigurationPolicyStatus defines the observed state of NodeNetworkConfigurationPolicy
 type NodeNetworkConfigurationPolicyStatus struct {
 	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
+
+	// NodeRunningUpdate field is used for serializing cluster nodes configuration
+	// +optional
+	NodeRunningUpdate string `json:"nodeRunningUpdate,omitempty" optional:"true"`
+
+	// NodeUpdateStart marks starting time of a node on a policy configuration
+	// +optional
+	NodeUpdateStart *metav1.Time `json:"nodeUpdateStart,omitempty" optional:"true"`
 }
 
 const (
--- a/automation/check-patch.e2e-k8s.sh
+++ b/automation/check-patch.e2e-k8s.sh
@@ -31,7 +31,15 @@ main() {
     make cluster-up
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
     make cluster-sync
-    make E2E_TEST_TIMEOUT=1h E2E_TEST_ARGS="-noColor" E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml" test-e2e-handler
+
+    export E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml"
+    if [ $NMSTATE_PARALLEL_ROLLOUT == "true" ]; then
+       E2E_TEST_SUITE_ARGS="${E2E_TEST_SUITE_ARGS} -ginkgo.skip='user-guide|nns'"
+    else
+       E2E_TEST_SUITE_ARGS="${E2E_TEST_SUITE_ARGS} -ginkgo.skip='parallel'"
+    fi
+
+    make E2E_TEST_TIMEOUT=1h E2E_TEST_ARGS="-noColor" test-e2e-handler
 }
 
 [[ "${BASH_SOURCE[0]}" == "$0" ]] && main "$@"
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -20,11 +20,13 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -41,12 +43,15 @@ import (
 	nmstate "github.com/nmstate/kubernetes-nmstate/pkg/helper"
 	"github.com/nmstate/kubernetes-nmstate/pkg/policyconditions"
 	"github.com/nmstate/kubernetes-nmstate/pkg/selectors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
 var (
-	nodeName       string
-	watchPredicate = predicate.Funcs{
+	nodeName                   string
+	nodeConfigurationTimeout   = 5 * time.Minute
+	nodeRunningUpdateRetryTime = 5 * time.Second
+	watchPredicate             = predicate.Funcs{
 		CreateFunc: func(createEvent event.CreateEvent) bool {
 			return true
 		},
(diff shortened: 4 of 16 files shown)


api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -12,6 +14,15 @@ type NodeNetworkConfigurationPolicySpec struct {
...
 	DesiredState State `json:"desiredState,omitempty"`
+
+	// +kubebuilder:validation:XPreserveUnknownFields
+	// When Parallel is set to true, changes are applied to all nodes in parallel
+	// +optional
+	Parallel bool `json:"parallel,omitempty"`
+
+	NodeRunningUpdate string `json:"nodeRunningUpdate,omitempty" optional:"true"`
+
+	NodeUpdateStart metav1.Time `json:"nodeUpdateStarted,omitempty" optional:"true"`
```
phoracek, line 25: These two should be in PolicyState, not Spec, no?

rhrazdil: Yes, moved to PolicyState

api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -11,21 +13,36 @@ type NodeNetworkConfigurationPolicySpec struct {
 	// +kubebuilder:validation:XPreserveUnknownFields
 	// The desired configuration of the policy
 	DesiredState State `json:"desiredState,omitempty"`
+
+	// +kubebuilder:validation:XPreserveUnknownFields
```
qinqon, line 17: We don't need this

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -180,6 +239,30 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(request ctrl.Reques
...
+		// TODO:
+		//if metav1.Now().Sub(instance.Status.NodeUpdateStart.Time) > nodeConfigurationTimeout {
+		//	// a node has been running the update for too long
+		//	enactmentConditions.NotifyConfigurationAborted(fmt.Errorf("failed to configure in time"))
+		//	return reconcile.Result{}, nil
+		//}
+
+		err = r.setNodeRunningUpdate(request.NamespacedName)
+		if err != nil {
+			return reconcile.Result{RequeueAfter: nodeConfigurationRequeueTime}, nil
```
qinqon, line 262: Are we sure we don't want to block `Reconcile` ? I am a little concerned about nncp changed in the middle of rollout (We are suppose to not support multiple nncps).

pkg/enactmentstatus/conditions/conditions.go
```diff
@@ -123,6 +131,43 @@ func SetFailed(conditions *nmstate.ConditionList, reason nmstate.ConditionReason
...
+}
+
+func SetConfigurationAborted(conditions *nmstate.ConditionList, message string) {
+	SetAborted(conditions, nmstate.NodeNetworkConfigurationEnactmentConditionConfigurationAborted, message)
+}
+
+func SetAborted(conditions *nmstate.ConditionList, reason nmstate.ConditionReason, message string) {
+	conditions.Set(
+		nmstate.NodeNetworkConfigurationEnactmentConditionFailing,
+		corev1.ConditionTrue,
```
qinqon, line 149: Abort is not Failing, failing means that applying the configuration didn't work out.

qinqon (changes requested):
first pass and a big picture question.

pkg/enactmentstatus/conditions/conditions.go
rhrazdil: You're right, thanks!

api/shared/nodenetworkconfigurationpolicy_types.go
rhrazdil: Removed

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: I was thinking about blocking the clearing function that removes the runningNode field, I didn't realize this.
In this case, I think it makes sense to block at both setting and unsetting the runningNode field

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -128,6 +133,81 @@ func (r *NodeNetworkConfigurationPolicyReconciler) initializeEnactment(policy nm
...
+	ec *enactmentconditions.EnactmentConditions,
+) (ctrl.Result, error) {
+	if policy.Status.NodeRunningUpdate != "" && metav1.Now().Sub(policy.Status.NodeUpdateStart.Time) > nodeConfigurationTimeout {
+		// a node has been running the update for too long
+		errmsg := fmt.Errorf("A node has been configuring for too long, aborting")
+		ec.NotifyFailedToConfigure(errmsg)
+		return reconcile.Result{}, errmsg
+	}
+	err := r.setNodeRunningUpdate(types.NamespacedName{Name: policy.GetName(), Namespace: policy.GetNamespace()})
+	if err != nil {
```
qinqon, line 176: We have to retry only on conflict, also don't return `ctrl.Result` here just return the error the caller will check if it's a conflict and return return `ctrl.Result{RequeueAfter: nodeRunningUpdateRetryTime}, err` since this function is not blocking anymore.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -161,6 +241,9 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(request ctrl.Reques
 
 	enactmentConditions := enactmentconditions.New(r.Client, nmstateapi.EnactmentKey(nodeName, instance.Name))
 
+	if !instance.Spec.Parallel {
+		defer r.releaseNodeRunningUpdate(request.NamespacedName)
```
qinqon, line 245: We have to move this to the other Parallel if statement block

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -180,13 +263,21 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(request ctrl.Reques
 
 	enactmentConditions.NotifyMatching()
 
+	if !instance.Spec.Parallel {
+		res, err := r.claimNodeRunningUpdate(instance, &enactmentConditions)
+		if err != nil {
+			return res, err
+		}
+	}
```
qinqon, line 271: Let's do here something like
```golang
if !instance.Spec.Parallel {
		err := r.claimNodeRunningUpdate(instance, &enactmentConditions)
		if err != nil {
                        if errors.IsConflict(err) {
                            ctrl.Result{RequeueAfter: nodeRunningUpdateRetryTime}, nil
                        } else {
                             ctrl.Result{}, err
                        }
			return res, err
		}
                defer r.releaseNodeRunningUpdate(request.NamespacedName)
	}
```

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -128,6 +133,81 @@ func (r *NodeNetworkConfigurationPolicyReconciler) initializeEnactment(policy nm
...
+	_ = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		err := r.Client.Get(context.TODO(), policyKey, instance)
+		if err != nil {
+			r.Log.Info("Failed to get policy, retrying")
+			return err
+		}
+		// release only if we are the owner
+		if instance.Status.NodeRunningUpdate != nodeName {
+			return nil
+		}
```
qinqon, line 193: Remove this if we move `releaseNodeRunningUpdate` to the proper place withing Reconcile

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
```diff
@@ -48,6 +48,10 @@ spec:
                   type: string
                 description: 'NodeSelector is a selector which must be true for the policy to be applied to the node. Selector which must match a node''s labels for the policy to be scheduled on that node. More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/'
                 type: object
+              parallel:
+                description: When set to true, changes are applied to all nodes in parallel
+                type: boolean
+                x-kubernetes-preserve-unknown-fields: true
```
qinqon, line 54: Why do we need this ?

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
```diff
@@ -136,6 +151,13 @@ spec:
                   - type
                   type: object
                 type: array
+              nodeRunningUpdate:
+                description: NodeRunningUpdate field is used for serializing cluster nodes configuration
```
qinqon, line 155:
```suggestion
                description: NodeRunningUpdate field is used for serializing cluster nodes configuration when Parallel flag is false
```

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
```diff
@@ -136,6 +151,13 @@ spec:
...
+                type: string
+              nodeUpdateStart:
+                description: NodeUpdateStart marks starting time of a node on a policy configuration
```
qinqon, line 158:
```suggestion
                description: NodeUpdateStart marks starting time of a node on a policy configuration when Parallel flag is false
```

docs/user-guide/102-configuration.md
```diff
@@ -9,12 +9,14 @@ routing on cluster nodes. The configuration is driven by two main object types,
...
 A Policy describes what is the desired network configuration on cluster nodes.
-It is created by used and applies cluster-wide. On the other hand, an Enactment
+It is created by users and applies cluster-wide. On the other hand, an Enactment
 is a read-only object that carries execution state of a Policy per each Node.
 
-Policies are applied on node via NetworkManager. Thanks to this, the
+Policies are applied on a node via NetworkManager. Thanks to this, the
 configuration is persistent on the node and survives reboots.
 
+Only one policy can be in progress at a time.
```
qinqon, line 18:
```suggestion
:warning:  Only one policy can be in progress at a time.
```

Also even change NNCP description to say warning on applying a policy while there is another in progress is undefined behaviour.

qinqon (changes requested):
We have to think about a way to tests parallel: true at e2e tests.

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Done

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Done

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Done

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Done

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
rhrazdil: Forgot to remove this, Done

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
rhrazdil: Done

docs/user-guide/102-configuration.md
rhrazdil: Done

qinqon:
/hold
we are missing skip at example and nns tests for parallel lane and also add a test that runs only in parallel lane to check that enacments get created at the same time more or less.

test/e2e/handler/nncp_parallel_test.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		BeforeEach(func() {
+			node := nodes[0]
+			By("Create a policy")
+			updateDesiredState(linuxBrUp(bridge1))
+			By(fmt.Sprintf("Wait for %s progressing state reached", node))
+			enactmentConditionsStatusEventually(node).Should(ConsistOf(progressConditions))
+		})
+		AfterEach(func() {
+			By("Remove the policy")
```
qinqon, line 42: Test is failing because the bridge is not removed at AfterEach, first you neeed to apply the absent bridge policy then delete the policy.

Makefile
```diff
@@ -28,9 +28,13 @@ export KUBEVIRT_PROVIDER ?= k8s-1.19
 export KUBEVIRT_NUM_NODES ?= 2 # 1 master, 1 worker needed for e2e tests
 export KUBEVIRT_NUM_SECONDARY_NICS ?= 2
 
-export E2E_TEST_TIMEOUT ?= 40m
+export E2E_TEST_TIMEOUT ?= 60m
 
+ifeq ($(findstring true,$(NMSTATE_PARALLEL_ROLLOUT)),true)
```
qinqon, line 33: Let's do the skip just in automation, so it's only skipped at CI.

So by default a user will run all the test in sequential rollout and the small parallel test that check enactments.

But thinking about it what we really should do is to trait parallel and sequential rollout at different tests with DescriptionTable or the like, so users just run all of them and they are different test with different names and results, but that's a lot of work.

test/e2e/handler/nncp_parallel_test.go
```diff
@@ -0,0 +1,75 @@
...
+			deletePolicy(TestPolicy)
+			By("Reset desired state at all nodes")
+			resetDesiredStateForNodes()
+		})
+		It("should be progressing on multiple nodes in parallel", func() {
+			if !parallelRollout {
```
qinqon, line 48: We don't need this we are skipping at CI with -skip

qinqon (changes requested):
We are missing skipping the examples at parallel lane right ?

test/e2e/handler/nncp_parallel_test.go
rhrazdil: Right, removed

test/e2e/handler/nncp_parallel_test.go
rhrazdil: Thanks, fixed

automation/check-patch.e2e-k8s.sh
```diff
@@ -31,7 +31,13 @@ main() {
     make cluster-up
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
     make cluster-sync
-    make E2E_TEST_TIMEOUT=1h E2E_TEST_ARGS="-noColor" E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml" test-e2e-handler
+
+    if [ $NMSTATE_PARALLEL_ROLLOUT == "true" ]; then
+      export E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml -ginkgo.skip='user-guide'"
+    else
+      export E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml -ginkgo.skip='parallel'"
+    fi
```
qinqon, line 39: Let's concantenat E2E_TEST_SUITE_ARGS so we don't repeat ourselves.

test/e2e/handler/user_guide_test.go
```diff
@@ -44,13 +44,13 @@ var _ = Describe("Introduction", func() {
 			resetDesiredStateForNodes()
 		})
 
-		It("should succeed executing all the commands", func() {
+		It("[user-guide] should succeed executing all the commands", func() {
```
qinqon, line 47: Put [user-guide] at the `Describe` so we don't need to repeat it.

qinqon:
@rhrazdil I think we can skip the following tests from parallel lane too
https://github.com/nmstate/kubernetes-nmstate/blob/master/test/e2e/handler/nns_update_timestamp_test.go
https://github.com/nmstate/kubernetes-nmstate/blob/master/test/e2e/handler/nodes_test.go

They don't involve nncp or nnce

test/e2e/handler/user_guide_test.go
rhrazdil: Thanks, updated

automation/check-patch.e2e-k8s.sh
rhrazdil: Updated

Makefile
rhrazdil: > Let's do the skip just in automation, so it's only skipped at CI.

Done

> But thinking about it what we really should do is to trait parallel and sequential rollout at different tests with  DescriptionTable or the like, so users just run all of them and they are different test with different names and results, but that's a lot of work.

I agree, I'll add this to the list of followus in the PR description

automation/check-patch.e2e-k8s.sh
```diff
@@ -31,7 +31,15 @@ main() {
     make cluster-up
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
     make cluster-sync
-    make E2E_TEST_TIMEOUT=1h E2E_TEST_ARGS="-noColor" E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml" test-e2e-handler
+
+    export E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml"
+    if [ $NMSTATE_PARALLEL_ROLLOUT == "true" ]; then
+       E2E_TEST_SUITE_ARGS="${E2E_TEST_SUITE_ARGS} -ginkgo.skip='user-guide|nns'"
```
qinqon, line 37: This does not skip https://github.com/nmstate/kubernetes-nmstate/blob/master/test/e2e/handler/nodes_test.go

qinqon:
/hold cancel 

All good

PR #624, opened 2020-11-10
Disable metrics default behaviour

qinqon:

>
/kind bug
> /kind enhancement

By default the controller-runtime metrics will start up at
port 8080 for handler pod this is a problem since it's operating
on host networking so it's opening port 8080 directly at node, let's
just disable it and think about a solution in in the future in case
metrics are really needed.

**Release note**:

```release-note
Disable default metrics on port 8080
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,8 +64,7 @@ func init() {
 }
 
 func main() {
-	var metricsAddr, logType string
-	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	var logType string
 	flag.StringVar(&logType, "v", "production", "Log type (debug/production).")
 	flag.Parse()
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -85,7 +85,8 @@ func main() {
 	}
 
 	ctrlOptions := ctrl.Options{
-		Scheme: scheme,
```
phoracek, line 88: There is an attribute on like 68 called "metrics-addr'. Can we drop it now?

phoracek (changes requested):

main.go
```diff
@@ -85,7 +85,8 @@ func main() {
...
+		Scheme:             scheme,
+		MetricsBindAddress: "0", // disable metrics
```
oshoval, line 89: cant u pick other port on line 68 ?

main.go
qinqon: Done

main.go
qinqon: Please add the comment to the proper line, I have remove the argument all together so no need to do anything about it.

phoracek (approved):
/lgtm
/approve

Thanks.

PR #625, opened 2020-11-10
[release-0.31] Disable metrics default behaviour

qinqon:

>
/kind bug
> /kind enhancement

By default the controller-runtime metrics will start up at
port 8080 for handler pod this is a problem since it's operating
on host networking so it's opening port 8080 directly at node, let's
just disable it and think about a solution in in the future in case
metrics are really needed.

**Release note**:

```release-note
Disable default metrics 
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -107,8 +107,9 @@ func main() {
 	}
 
 	mgrOptions := manager.Options{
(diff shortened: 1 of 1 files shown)


phoracek (approved):
/lgtm
/approve

PR #630, opened 2020-11-11
Create OLM Bundle

bcrochet:
This creates the bundle necessary for integrating with Operator Lifecycle
Manager (OLM).

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -177,6 +187,33 @@ vendor: $(GO)
 tools: $(GO)
 	./hack/install-tools.sh
 
+# Generate bundle manifests and metadata, then validate generated files.
+bundle: gen-crds manifests
```
phoracek, line 191: The upgrade guide for 1.2.0 says this should use kustomize: https://github.com/operator-framework/operator-sdk/blob/master/website/content/en/docs/upgrading-sdk-version/v1.2.0.md#golang-based-operators-update-makefiles-bundle-target

phoracek (changes requested):

Makefile
bcrochet: That is planned. Just want to get to this sooner than later.

api/v1beta1/nmstate_types.go
```diff
@@ -30,15 +31,22 @@ type NMStateSpec struct {
...
 // +kubebuilder:storageversion
+// +kubebuilder:subresource=status
 
 // NMState is the Schema for the nmstates API
 type NMState struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              NMStateSpec `json:"spec,omitempty"`
+	Spec              NMStateSpec   `json:"spec,omitempty"`
+	Status            NMStateStatus `json:"status,omitempty"`
```
qinqon, line 49: I don't see the controller code that update the status.

qinqon (changes requested):
Looks good overall: a nit and a question.

api/v1beta1/nmstate_types.go
bcrochet: The controller doesn't do anything with that. OLM uses the Conditions for it's own operator.

PR #633, opened 2020-11-12
Remove code directories from .dockerignore

bcrochet:
Remove pkg cmd and vendor from .dockerignore, as it causes builds
    with Docker and the build/Dockerfile.*openshift builds to fail.
    This is due to them being multistage builds.

>
> /kind bug
> /kind enhancement

   Remove pkg cmd and vendor from .dockerignore, as it causes builds
    with Docker and the build/Dockerfile.*openshift builds to fail.
    This is due to them being multistage builds.
    
    The error is:
    
    Step 4/10 : RUN go build --mod=vendor -o build/_output/bin/manager main.go
    ---> Running in 946711c1028e
    go: inconsistent vendoring in /go/src/github.com/openshift/kubernetes-nmstate:
    
    This is caused by the vendor directory not being present. This does not affect
    Podman builds.
    
    The multistage builds need to have a complete picture, so we need
    to include these directories in the build context. This will allow the
    regular Dockerfiles to also be migrated to a multistage build approach,
    which is desirable.

**Release note**:

```release-note
none
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,9 +16,6 @@ _kubevirtci
 kubevirtci
 
 # No need for source code is already compiled
-pkg
-cmd
-vendor
 docs
 
 # Test and infra


phoracek:
Please be more specific. Why do those builds fail, what is the error message (for reference) and how does this change fix that?

/hold

bcrochet:
> Please be more specific. Why do those builds fail, what is the error message (for reference) and how does this change fix that?
> 
> /hold

See description.

phoracek:
Thanks.

How comes they don't match? And how is it related to podman vs docker?

.dockerignore
```diff
@@ -16,9 +16,6 @@ _kubevirtci
 kubevirtci
 
 # No need for source code is already compiled
-pkg
-cmd
-vendor
```
phoracek, line 21: Can we delete them from the container after the build is done?

phoracek: Sorry, brainfart, it is multistage and it never gets to the second stage of course.

phoracek:
Thanks!

/hold cancel

PR #635, opened 2020-11-13
knmstate, docs: Remove docker caching

### LINKED ISSUE #637: [Flake-CI] Job publish-kubernetes-nmstate of type postsubmit ended with state failure
**What happened**:
Build failed after ~6 minutes.
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/publish-kubernetes-nmstate/1328235972407595008
To Build and run tests.

**Environment**:

- Others:

/kind bug
/triage build-watcher

qinqon:

>
/kind bug
> /kind enhancement

knmstate, docs: Don't cache builder

The current docs builder caching mechanism is implemented as a
container that use the FROM clausule from the same container that's
pretty wrong since you get never ending layers, let's just remove that
cache and think about it if it's necessary, we are talking about mintues.

Closes #637 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/publish.sh
+++ b/automation/publish.sh
@@ -8,40 +8,33 @@
 
 image_registry=${IMAGE_REGISTRY:-quay.io}
 image_repo=${IMAGE_REPO:-nmstate}
-branch=${PULL_BASE_REF:-$(git rev-parse --abbrev-ref HEAD)}
-docs_source_branch=master
-docs_container=${image_registry}/${image_repo}/kubernetes-nmstate-docs:$docs_source_branch
 
-# Publish kubernets-nmstate containers
 source automation/check-patch.setup.sh
-(
+cd ${TMP_PROJECT_PATH}
+
+push_knmstate_containers() {
     cd ${TMP_PROJECT_PATH}
     make \
         IMAGE_REGISTRY=${image_registry}  \
         IMAGE_REPO=${image_repo} \
         push-handler \
         push-operator
-)
+}
 
-# We don't have a versioned docs webpage so we only update it from master and
-# not from release branches
-if [ "$branch" == $docs_source_branch ]; then
-
-    # Update kubernetes-nmstate documentation container
-    docker build docs --build-arg BRANCH=$docs_source_branch -t $docs_container
-    docker push $docs_container
-
-    # To pass user/password from automations, idea is taken from [1]
-    # [1] https://stackoverflow.com/questions/8536732/can-i-hold-git-credentials-in-environment-variables
-    git config credential.helper '!f() { sleep 1; echo "username=${GITHUB_USER}"; echo "password=${GITHUB_TOKEN}"; }; f'
-
+publish_docs() {
     # Update gh-pages branch with the generated documentation
-    git checkout gh-pages --quiet
-    git rm -r --quiet *
-    docker run -v $(pwd):/gh-pages $docs_container bash -c "cp -r docs/build/kubernetes-nmstate/* /gh-pages"
-    git add -A
-    git commit -m "updated $(date +"%d.%m.%Y %H:%M:%S")"
-    git push --quiet
+    docker run -v $(pwd)/docs:/docs/ ruby make -C /docs install build
+    rm -rf /tmp/gh-pages
+    git clone --single-branch http://github.com/nmstate/kubernetes-nmstate -b gh-pages /tmp/gh-pages
+    rsync -rt --links --cvs-exclude docs/build/kubernetes-nmstate/* /tmp/gh-pages
+    (
+        cd /tmp/gh-pages
+        git add -A
+        git commit -m "updated $(date +"%d.%m.%Y %H:%M:%S")"
+        git push https://${GITHUB_USER}github.com/nmstate/kubernetes-nmstate gh-pages
+    )
     echo -e "\033[0;32mdemo updated $(date +"%d.%m.%Y %H:%M:%S")\033[0m"
-    git checkout $docs_source_branch --quiet
-fi
+}
+
+push_knmstate_containers
+publish_docs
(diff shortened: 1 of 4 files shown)


automation/publish.sh
```diff
@@ -8,39 +8,43 @@
...
+    )
     echo -e "\033[0;32mdemo updated $(date +"%d.%m.%Y %H:%M:%S")\033[0m"
-    git checkout $docs_source_branch --quiet
+}
+
+#push_handler
+
+# We don't have a versioned docs webpage so we only update it from master and
+# not from release branches
+if [ "$branch" == $docs_source_branch ]; then
```
phoracek, line 48: But this will be always master, no?

automation/publish.sh
```diff
@@ -8,39 +8,43 @@
...
+        cd /tmp/gh-pages
+        git add -A
+        git commit -m "updated $(date +"%d.%m.%Y %H:%M:%S")"
+        git push https://kubevirt-bot@github.com/nmstate/kubernetes-nmstate gh-pages
+    )
     echo -e "\033[0;32mdemo updated $(date +"%d.%m.%Y %H:%M:%S")\033[0m"
-    git checkout $docs_source_branch --quiet
+}
+
+#push_handler
```
phoracek, line 44: This should not be commented

automation/publish.sh
```diff
@@ -8,39 +8,43 @@
...
+    make -C docs install build
+    rm -rf /tmp/gh-pages
+    git clone http://github.com/nmstate/kubernetes-nmstate -b gh-pages /tmp/gh-pages
+    git -C /tmp/gh-pages rm -r \* --quiet
+    cp -r docs/build/kubernetes-nmstate/* /tmp/gh-pages
+    (
+        cd /tmp/gh-pages
+        git add -A
+        git commit -m "updated $(date +"%d.%m.%Y %H:%M:%S")"
+        git push https://kubevirt-bot@github.com/nmstate/kubernetes-nmstate gh-pages
```
phoracek, line 39: Should not be the username obtained from an environment variable? Feel weird to hardcode it here

automation/publish.sh
```diff
@@ -8,39 +8,43 @@
...
 
+publish_docs() {
     # Update gh-pages branch with the generated documentation
-    git checkout gh-pages --quiet
-    git rm -r --quiet *
-    docker run -v $(pwd):/gh-pages $docs_container bash -c "cp -r docs/build/kubernetes-nmstate/* /gh-pages"
-    git add -A
-    git commit -m "updated $(date +"%d.%m.%Y %H:%M:%S")"
-    git push --quiet
+    make -C docs install build
```
phoracek, line 30: You should use `        cd ${TMP_PROJECT_PATH}`. Otherwise we just happen to be in the directory as a side effect of a previous function.

automation/publish.sh
```diff
@@ -8,39 +8,43 @@
...
+    rm -rf /tmp/gh-pages
+    git clone http://github.com/nmstate/kubernetes-nmstate -b gh-pages /tmp/gh-pages
```
phoracek, line 32: Why do we fetch it? We should have the sourcecode availble in the job, no?

phoracek (changes requested):
Please mention why are we doing this in the commit message

automation/publish.sh
qinqon: I was thinking that we should run publish.sh at release branches since we want to run knmstate from release- branch HEAD but I think we are just running publish.sh at master so you right.

automation/publish.sh
qinqon: I am following what project-infra does for kubevirtci release, they harcoded the user but I can change that.

automation/publish.sh
qinqon: I don't want to change where prow is pointing to in the clone, that's why I just clone the branch and push it.

automation/publish.sh
qinqon: Let me add and env var with it in the project-infra PR

docs/_config.yaml
```diff
@@ -4,8 +4,8 @@ locale: "en-US"
 title: "The Kubernetes NMState project"
 description: A declarative API for Kubernetes node network management
 name: "The Kubernetes NMState team"
-url: "https://nmstate.github.io"
-baseurl: "/kubernetes-nmstate"
+url: "https://storage.googleapis.com"
```
phoracek, line 7: Why do you change this? Would not it break the page?

automation/publish.sh
```diff
@@ -8,39 +8,32 @@
...
-    # Update kubernetes-nmstate documentation container
-    docker build docs --build-arg BRANCH=$docs_source_branch -t $docs_container
-    docker push $docs_container
-
-    # To pass user/password from automations, idea is taken from [1]
-    # [1] https://stackoverflow.com/questions/8536732/can-i-hold-git-credentials-in-environment-variables
-    git config credential.helper '!f() { sleep 1; echo "username=${GITHUB_USER}"; echo "password=${GITHUB_TOKEN}"; }; f'
+}
 
+publish_docs() {
```
phoracek, line 23: This should happen only on the doc branch, no?

phoracek (changes requested):

docs/_config.yaml
qinqon: This is from manual call to publish.sh, thanks !

automation/publish.sh
qinqon: we publish from `master` into `gh-pages` branch, is the way github pages works.

automation/publish.sh
qinqon: Done

phoracek (approved):
/lgtm
/approve

PR #645, opened 2020-11-27
Reduce default gw probe traces

qinqon:

>
> /kind bug
> /kind enhancement

Currectly kubernetes-nmstate prints the network state every time it does
not find the default gw when the probe is running, clearly this is too
verbose. To fix that this change in case of "-v=production" just print
the error and the gjson path used to find the default gw, in case of
"-v=debug" it also prints the network state.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/probe/probes.go
+++ b/pkg/probe/probes.go
@@ -124,10 +124,18 @@ func defaultGw() (string, error) {
 		if err != nil {
 			return false, errors.Wrap(err, "failed retrieving current state to retrieve default gw")
 		}
+		defaultGwGjsonPath := "routes.running.#(destination==\"0.0.0.0/0\").next-hop-address"
 		defaultGw = gjsonCurrentState.
-			Get("routes.running.#(destination==\"0.0.0.0/0\").next-hop-address").String()
+			Get(defaultGwGjsonPath).String()
 		if defaultGw == "" {
-			log.Info("default gw missing", "state", gjsonCurrentState.String())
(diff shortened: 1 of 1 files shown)


pkg/probe/probes.go
```diff
@@ -124,10 +124,18 @@ func defaultGw() (string, error) {
...
 		}
+		defaultGwGjsonPath := "routes.running.#(destination==\"0.0.0.0/0\").next-hop-address"
 		defaultGw = gjsonCurrentState.
-			Get("routes.running.#(destination==\"0.0.0.0/0\").next-hop-address").String()
+			Get(defaultGwGjsonPath).String()
 		if defaultGw == "" {
-			log.Info("default gw missing", "state", gjsonCurrentState.String())
+			msg := "default gw missing"
+			defaultGwLog := log.WithValues("path", defaultGwGjsonPath)
+			defaultGwLogDebug := defaultGwLog.V(1)
```
rhrazdil, line 133: I'm not sure I understand the verbosity levels - in main.go there is a debug msg with V(2), why is here `defaultWgLogDebug` with `V(1)`?

qinqon: Forgot to remove the debug trace at main.go I added it to check what level works fine, the levels are from zap library and 1 works if we are in dev mode.

rhrazdil: Thanks

phoracek (approved):
/approve

PR #648, opened 2020-11-30
Fix intermitent Status at NNCP sequencial rollout

rhrazdil:

>
/kind bug
> /kind enhancement

Reconcile always resets nncp status at the beginning.
This doesn't cause issue when applied in parallel, but
in case of sequential rollout, the nncp is missing status
information for the whole duration of a reconcile.

This PR addresses if by resetting only when not using webhook.
When using webhook, it handles status reset of nncp when new nncp 
is created/updated, so we don't need to reset in Reconcile.

The issue persists when not using webhook, but does not affect functionality.

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -50,6 +50,7 @@ import (
 	nmstate "github.com/nmstate/kubernetes-nmstate/pkg/helper"
 	"github.com/nmstate/kubernetes-nmstate/pkg/policyconditions"
 	"github.com/nmstate/kubernetes-nmstate/pkg/selectors"
+	nncp "github.com/nmstate/kubernetes-nmstate/pkg/webhook/nodenetworkconfigurationpolicy"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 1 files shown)


qinqon:
@rhrazdil Looks like this solution breaks CI very bad, we have to thing about something else.

rhrazdil:
@qinqon I think it's caused by the `should successfully move default IP address on top of the bond` test case, I see it in some cases leave node02 in NotReady state even without this change

qinqon:
@rhrazdil maybe we can just detect that the webhook is running by looking at the timestamp label at NNCP and not do the Reset if we do.

rhrazdil:
@qinqon Do you mean the webhook-mutating-timestamp? Isn't that label added on nncp creation?

qinqon:
@rhrazdil it's added in CREATION and UPDATE if webhook si in place, in that case I think we don't need the Reset at reconcile, so this will be something like "If you don't  configure the webhook Status will be intermitent at sequencial rollout"

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -218,7 +219,9 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(request ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
-	policyconditions.Reset(r.Client, request.NamespacedName)
+	if instance.ObjectMeta.Annotations[nncp.TimestampLabelKey] != "" {
+		policyconditions.Reset(r.Client, request.NamespacedName)
+	}
```
qinqon, line 224: This will crash if it's not present, this has to be:
```golang
_, hasWebhookLabel := instance.ObjectMeta.Annotations[nncp.TimestampLabelKey]
if !hasWebhookLabel {
    policyconditions.Reset(r.Client, request.NamespacedName)
}
```

qinqon (changes requested):

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: ouch thanks, updated

PR #652, opened 2020-12-02
Policy rollout in chunks of custom size

rhrazdil:

>
> /kind bug
/kind enhancement

This PR changes policy spec parallel to a more versatile
chunkSize, which allows user to define how many percent of nodes are allowed
to apply a policy configuration at a time.

chunkSize == 0: all (matching) cluster nodes apply policy in parallel
chinkSize == n ; n > 0:  up to `n` percent of matching nodes apply policy in parallel

If not set, policy is applied sequentially one node at a time.

the env variable `NMSTATE_PARALLEL_ROLLOUT` is changed to `NMSTATE_CHUNK_SIZE`
To get this PR through CI, we need to:
1. add `export NMSTATE_CHUNK_SIZE` (so both env vars are temporarily exported)
2. merge this PR
3. remove `export NMSTATE_PARALLEL_ROLLOUT` which is no longer used  

**Please ignore** commit `Add Aborted conditionType for policy enactments`, this is part of another PR and will be removed from this PR when it's merged

**Release note**:

```release-note
Changed policy field `parallel` to `chunkSize`
```

### DIFF
--- a/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -12,18 +12,21 @@ type NodeNetworkConfigurationPolicySpec struct {
 	// The desired configuration of the policy
 	DesiredState State `json:"desiredState,omitempty"`
 
-	// When set to true, changes are applied to all nodes in parallel
+	// Percentage of matching cluster nodes that may apply policy configuration concurrently.
+	// If not specified, policy is applied one node at a time.
+	// Value of 0 means all nodes apply policy in parallel,
+	// Value of n, n > 1 means that up to n% of cluster nodes may apply policy configuration concurrently.
 	// +optional
-	Parallel bool `json:"parallel,omitempty"`
+	ChunkSize *int `json:"chunkSize,omitempty"`
 }
 
 // NodeNetworkConfigurationPolicyStatus defines the observed state of NodeNetworkConfigurationPolicy
 type NodeNetworkConfigurationPolicyStatus struct {
 	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
 
-	// NodeRunningUpdate field is used for serializing cluster nodes configuration when Parallel flag is false
+	// NodeChunkCapacity is used as a semaphore to synchronize nodes applying policy configuration
 	// +optional
-	NodeRunningUpdate string `json:"nodeRunningUpdate,omitempty" optional:"true"`
+	NodeChunkCapacity int `json:"nodeChunkCapacity,omitempty" optional:"true"`
 }
 
 const (
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"math"
 	"reflect"
 	"time"
 
@@ -56,6 +57,7 @@ import (
 var (
 	nodeName                                string
 	nodeRunningUpdateRetryTime              = 5 * time.Second
+	defaultMaxNodeCapacity                  = 1
 	onCreateOrUpdateWithDifferentGeneration = predicate.Funcs{
 		CreateFunc: func(createEvent event.CreateEvent) bool {
 			return true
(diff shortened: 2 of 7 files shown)


rhrazdil:
cc @qinqon @phoracek

api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -12,18 +12,21 @@ type NodeNetworkConfigurationPolicySpec struct {
...
 	DesiredState State `json:"desiredState,omitempty"`
 
-	// When set to true, changes are applied to all nodes in parallel
+	// Percentage of matching cluster nodes that may apply policy configuration concurrently.
+	// If not specified, policy is applied one node at a time.
+	// Value of 0 means all nodes apply policy in parallel,
+	// Value of n, n > 1 means that up to n% of cluster nodes may apply policy configuration concurrently.
 	// +optional
-	Parallel bool `json:"parallel,omitempty"`
+	ChunkSize *int `json:"chunkSize,omitempty"`
```
qinqon, line 20: I am super worry about upgrades here I think we have release a versin with the `Parallel` parameter, we have to read operator-sdk doc but I think we have to add ChickSize and mark Parallel as obsolete and at the next release delete Parallel, also maybe even add migration code that in case of Parallel argument we conert it into ChucnkSize also we may need to move this tov1btea2 or something like this, we will need to test this with the upgrae test, but let's sync first.

I have an idea maybe both make sense, If Parallel is false is pure sequencial rollouthttps://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4617/pull-kubevirt-e2e-kind-1.17-sriov/1336545341402517504 and if Paralle is true it means we run the NNCP inparallel and in that case ChunkSize can be configured.

api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -12,18 +12,21 @@ type NodeNetworkConfigurationPolicySpec struct {
 	// The desired configuration of the policy
 	DesiredState State `json:"desiredState,omitempty"`
 
-	// When set to true, changes are applied to all nodes in parallel
+	// Percentage of matching cluster nodes that may apply policy configuration concurrently.
+	// If not specified, policy is applied one node at a time.
+	// Value of 0 means all nodes apply policy in parallel,
```
qinqon, line 17: Sinze `n` is percentage the full parallel has to be n = 100, it will means that the ChunkSize is 100% of the cluster.

api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -12,18 +12,21 @@ type NodeNetworkConfigurationPolicySpec struct {
...
 
 // NodeNetworkConfigurationPolicyStatus defines the observed state of NodeNetworkConfigurationPolicy
 type NodeNetworkConfigurationPolicyStatus struct {
 	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
 
-	// NodeRunningUpdate field is used for serializing cluster nodes configuration when Parallel flag is false
+	// NodeChunkCapacity is used as a semaphore to synchronize nodes applying policy configuration
 	// +optional
-	NodeRunningUpdate string `json:"nodeRunningUpdate,omitempty" optional:"true"`
+	NodeChunkCapacity int `json:"nodeChunkCapacity,omitempty" optional:"true"`
```
qinqon, line 29: Can we add kubbuilder validator annotations for negative values ? (I think they exists and will be implemented add schema validations)

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -56,6 +57,7 @@ import (
 var (
 	nodeName                                string
 	nodeRunningUpdateRetryTime              = 5 * time.Second
+	defaultMaxNodeCapacity                  = 1
```
qinqon, line 60: 1% is two nodes at a 200 cluster size, I think we have to use something else for really sequencial rollout.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -165,16 +167,33 @@ func (r *NodeNetworkConfigurationPolicyReconciler) enactmentsCountsByPolicy(poli
 	return enactmentCount, nil
 }
 
-func (r *NodeNetworkConfigurationPolicyReconciler) claimNodeRunningUpdate(policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) error {
+func (r *NodeNetworkConfigurationPolicyReconciler) calculateMaxCapacityFromPercentage(enactmentCounts enactmentconditions.ConditionCount, chunkSize int) int {
+	if chunkSize > 100 {
```
qinqon, line 171: We have to add this validations in the schema.

qinqon (changes requested):
first pass: Let's sync on what to do with params.

qinqon:
/hold

Let's first sync on the big picture with the CRD parameters.

rhrazdil:
Closing this in favor of #715 
/close

PR #653, opened 2020-12-04
Add Node webhook to reset NNCP Status

qinqon:

>
> /kind bug
> /kind enhancement

When the Node labels are changed the NNCPs get Reconciled again but the
NNCP Status is not reset until the reconcile cycle is reached. This
change add a webhook that reset the NNCP condition if node labels are
changed.

Code with Hooks has beeing moved to hooks.go and the Handlers to handler.go

**Release note**:

```release-note
NONE
```


qinqon:
push: rebased

PR #655, opened 2020-12-07
Add an example policy of removing a bridge with eth port

rhrazdil:

>
> /kind bug

/kind enhancement

Provide an example of a policy that removes linux bridge connected to an ethernet interface.
The example demonstrates that with removal of the bridge, we need to explixitly enable
dhcp on the ethernet interface (or set static IP) in the same policy. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/examples_test.go
+++ b/test/e2e/handler/examples_test.go
@@ -50,6 +50,12 @@ var _ = Describe("[user-guide] Examples", func() {
 			ifaceNames: []string{"br1"},
 		},
 		exampleSpec{
+			name:       "Detach bridge port and restore its configuration",
+			fileName:   "detach-bridge-port-and-restore-eth.yaml",
+			policyName: "detach-bridge-port-and-restore-eth",
+			ifaceNames: []string{"br1"},
+		},
+		exampleSpec{
 			name:       "OVS bridge",
 			fileName:   "ovs-bridge.yaml",
 			policyName: "ovs-bridge",
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -18,6 +18,12 @@ toc_sticky: true
 {% include_absolute 'examples/bond-vlan.yaml' %}
 ```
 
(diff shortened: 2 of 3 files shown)


test/e2e/handler/examples_test.go
```diff
@@ -49,6 +49,12 @@ var _ = Describe("[user-guide] Examples", func() {
 			policyName: "linux-bridge",
 			ifaceNames: []string{"br1"},
 		},
+		exampleSpec{
+			name:       "Linux bridge absent",
```
phoracek, line 53: Detach bridge port and restore its configuration

I think that would speak more for the intent

test/e2e/handler/examples_test.go
```diff
@@ -49,6 +49,12 @@ var _ = Describe("[user-guide] Examples", func() {
...
+			fileName:   "linux-bridge-absent.yaml",
```
phoracek, line 54: Reflecting the title (but making it somehow shorter) here would be nice

phoracek (changes requested):

phoracek:
We should add the example to the webpage https://github.com/nmstate/kubernetes-nmstate/blob/master/docs/examples.md

test/e2e/handler/examples_test.go
```diff
@@ -49,6 +49,12 @@ var _ = Describe("[user-guide] Examples", func() {
 			policyName: "linux-bridge",
 			ifaceNames: []string{"br1"},
 		},
+		exampleSpec{
+			name:       "Detach bridge port and restore its configuration",
+			fileName:   "detach-bridge-port-and-restore.yaml",
+			policyName: "detach-bridge-port-and-restore",
```
phoracek, line 55: This does not match the policy name used in the file

phoracek (changes requested):

phoracek (approved):
/lgtm
/approve

PR #656, opened 2020-12-08
Fix condition checking in helper/client.go UpdateCurrentState

rhrazdil:

>
/kind bug
> /kind enhancement

Error checking condition contains ! (negation) when checking for NotFound type of error,
which should not be there.
As a consequence, we're masking all other types of error as NotFound

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -122,9 +122,10 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1beta1.N
 
 	err = client.Status().Update(context.Background(), nodeNetworkState)
 	if err != nil {
-		// Request object not found, could have been deleted after reconcile request.
-		if !apierrors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			return errors.Wrap(err, "Request object not found, could have been deleted after reconcile request")
+		} else {
+			return errors.Wrap(err, "Error updating nodeNetworkState")
 		}
 	}
 


rhrazdil:
cc @qinqon

pkg/helper/client.go
```diff
@@ -122,9 +122,10 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1beta1.N
 
 	err = client.Status().Update(context.Background(), nodeNetworkState)
 	if err != nil {
-		// Request object not found, could have been deleted after reconcile request.
-		if !apierrors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			return errors.Wrap(err, "Request object not found, could have been deleted after reconcile request")
+		} else {
+			return errors.Wrap(err, "Error updating nodeNetworkState")
```
qinqon, line 128: This will break conflict resolution, that's why is done this way.

qinqon (changes requested):
/hold
This is done in purpose I think.

pkg/helper/client.go
rhrazdil: we check `!apierrors.IsNotFound(err)` (mind the **!**)  and if that condition is true, we
`return errors.Wrap(err, "Request object not found, could have been deleted after reconcile request")`.
But that is surely the only error that did not occur

qinqon: but are we wrapping this function on `RetryOnConflict` for this to work we cannot wrap errors.

rhrazdil: We are already wrapping the error though

rhrazdil: Maybe we don't want to return an error when the object is not found, in that case it's only the error message that isn't right

qinqon:
/lgtm
/approve

We are not using retryOnConflict also it ends up returning the error at Reconcile so it will be requeued.

PR #657, opened 2020-12-08
Fix bonding e2e test

rhrazdil:

>
/kind bug
> /kind enhancement

Based on https://www.kernel.org/doc/Documentation/networking/bonding.txt,
when adding a bond interface in active_backup mode, the
MAC address of the bond is derived from the first active slave iface
However, the first active slave iface is not the first in the list of
slaves defined in the policy and is random.

**Removed** one interface from the created bond and file BZ for nmstate:
https://bugzilla.redhat.com/show_bug.cgi?id=1906307

When resolved, we need to re-add the second iface to the bond

Reference from the doc above:
```
8.  Where does a bonding device get its MAC address from?
	When using slave devices that have fixed MAC addresses, or when
the fail_over_mac option is enabled, the bonding device's MAC address is
the MAC address of the active slave.
	For other configurations, if not explicitly configured (with
ifconfig or ip link), the MAC address of the bonding device is taken from
its first slave device.  This MAC address is then passed to all following
slaves and remains persistent (even if the first slave is removed) until
the bonding device is brought down or reconfigured.
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/bonding_default_interface_test.go
+++ b/test/e2e/handler/bonding_default_interface_test.go
@@ -10,6 +10,7 @@ import (
 	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
 )
 
+// TODO: When https://bugzilla.redhat.com/show_bug.cgi?id=1906307 is resolved, add firstSecondaryNic to the bond again
 func boundUpWithPrimaryAndSecondary(bondName string) nmstate.State {
 	return nmstate.NewState(fmt.Sprintf(`interfaces:
   - name: %s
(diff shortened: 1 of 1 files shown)


rhrazdil:
cc @qinqon

phoracek (commented):
This does not look like a proper fix. Say I am a customer facing this issue. Asking me to set it up in two steps sounds like a poor workaround. Shouldn't we fix this on nmstate side?

qinqon:
@rhrazdil so @phoracek is right let's create a minimal reproducer with nmstate infra https://github.com/nmstate/nmstate/tree/base/automation and open a BZ then we notify nmstate team.

I suppose that adding the slaves at specified order would be enough, but maybe they give us a proper workaround directly.

rhrazdil:
@phoracek @qinqon  Filed this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1906307
In the meantime, shall we leave the test as is? 
What do you think about crating the bond with only one slave, so that the bond API remains tested, but the test doesn't fail.

phoracek:
Thanks for the BZ! Having it with a single slave would be a fine workaround. But make sure to add a huge TODO linking the BZ on top of it.

PR #660, opened 2020-12-09
Add Node webhook to reset NNCP Status

qinqon:

>
> /kind bug
> /kind enhancement

When the Node labels are changed the NNCPs get Reconciled again but the
NNCP Status is not reset until the reconcile cycle is reached. This
change add a webhook that reset the NNCP condition if node labels are
changed.

Code with Hooks has beeing moved to hooks.go and the Handlers to handler.go

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -229,6 +229,18 @@ webhooks:
         apiGroups: ["*"]
         apiVersions: ["v1alpha1","v1beta1"]
         resources: ["nodenetworkconfigurationpolicies", "nodenetworkconfigurationpolicies/status"]
+  - name: nodenetworkconfigurationpolicies-mutate-all.nmstate.io
+    clientConfig:
+      service:
+        name: {{template "handlerPrefix" .}}nmstate-webhook
+        namespace: {{ .HandlerNamespace }}
+        path: "/nodenetworkconfigurationpolicies-mutate-all"
+    rules:
+      - operations: ["UPDATE"]
+        apiGroups: ["*"]
+        apiVersions: ["v1"]
+        resources: ["nodes"]
+
 ---
 apiVersion: policy/v1beta1
 kind: PodDisruptionBudget
--- a/pkg/webhook/nodenetworkconfigurationpolicy/handler.go
+++ b/pkg/webhook/nodenetworkconfigurationpolicy/handler.go
@@ -42,7 +42,3 @@ func mutatePolicyHandler(neededMutationFor func(nmstatev1beta1.NodeNetworkConfig
 		return response
 	}
 }
-
-func always(nmstatev1beta1.NodeNetworkConfigurationPolicy) bool {
-	return true
-}
--- a/pkg/webhook/nodenetworkconfigurationpolicy/handler.go
+++ b/pkg/webhook/nodenetworkconfigurationpolicy/handler.go
@@ -8,6 +8,11 @@ import (
 
 	"github.com/pkg/errors"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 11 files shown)


qinqon:
/retest

bond flaky test

deploy/handler/operator.yaml
```diff
@@ -229,6 +229,18 @@ webhooks:
         apiGroups: ["*"]
         apiVersions: ["v1alpha1","v1beta1"]
         resources: ["nodenetworkconfigurationpolicies", "nodenetworkconfigurationpolicies/status"]
+  - name: node-modified.nmstate.io
```
rhrazdil, line 232: I suggest naming the hook node-labels-modified, so the it's clear what is the trigger (and also it would better fit the current naming convention)

```suggestion
  - name: node-labels-modified.nmstate.io
```

deploy/handler/operator.yaml
```diff
@@ -229,6 +229,18 @@ webhooks:
...
+    clientConfig:
+      service:
+        name: {{template "handlerPrefix" .}}nmstate-webhook
+        namespace: {{ .HandlerNamespace }}
+        path: "/node-modified"
```
rhrazdil, line 237: ditto

rhrazdil (changes requested):

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/20ac64908acc7254e2e2c71e7adfb5ee51090fb5..da4c4709700f6cd7271a6e836ce99c99fdb3b468): rename webhook to `node-labels-modified`, add a timestamp when it runs and a e2e tests that check that the webhook is running.

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/da4c4709700f6cd7271a6e836ce99c99fdb3b468..a73916db034966ba1fafce2b83cd044fdcf377fb): add missing test/node/labels.go file

qinqon:
@phoracek @rhrazdil CI is passing all good ?

phoracek (changes requested):
/hold

I cannot review this PR. There are too many changes and it feels like I'm looking on a huge state machine. To be fair, it does not seem that all of that is introduced by this PR. Issue is that you did refactoring and moved function in the same commit as this fix.

Please split it in multiple commits moving the code, introducing new abstraction (if there is any) and finally adding this new fix.

Commit/PR comment describing the flow in which we are getting this hook and what then happens to the annotation would be most welcome.

pkg/webhook/nodenetworkconfigurationpolicy/server.go
```diff
@@ -26,6 +26,7 @@ func Add(mgr manager.Manager, o certificate.Options) error {
 		webhookserver.WithHook("/nodenetworkconfigurationpolicies-mutate", deleteConditionsHook()),
 		webhookserver.WithHook("/nodenetworkconfigurationpolicies-status-mutate", setConditionsUnknownHook()),
 		webhookserver.WithHook("/nodenetworkconfigurationpolicies-timestamp-mutate", setTimestampAnnotationHook()),
+		webhookserver.WithHook("/node-labels-modified", deleteConditionsOnNodeLabelsModifiedHook(mgr.GetClient())),
```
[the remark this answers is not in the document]
qinqon: we express what is the hook doing, in this case the hook delete the conditions at all NNCPs when the node labels has changed.

pkg/webhook/nodenetworkconfigurationpolicy/handler.go
```diff
@@ -42,3 +47,60 @@ func mutatePolicyHandler(neededMutationFor func(nmstatev1beta1.NodeNetworkConfig
 		return response
 	}
 }
+
+func mutateAllPoliciesHandler(cli client.Client, neededMutationFor func(oldNode, newNode corev1.Node) bool, mutators ...mutator) admission.HandlerFunc {
```
phoracek, line 51: Why do we need so much code for "reset the NNCP condition if node labels are changed"?

Would you please describe in a single paragraph in the commit message what is your strategy here to reset the condition?

qinqon: Expaneded commit message.

PR #677, opened 2021-01-07
Merge eth1 and eth2 NNCPS in user guide into one

rhrazdil:

>
/kind bug
> /kind enhancement

After introducing parallel and sequential policy application
we decided to allow only one sequentially applied policy to be
in progress at a time.
This example in user guide wasn't updated to follow this.

Instead of creating two policies, one for configuring `eth1` and one `eth2` interface, I'm
placing the configuration to one policy. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/user-guide/102-configuration.md
+++ b/docs/user-guide/102-configuration.md
@@ -343,7 +343,7 @@ kubectl apply -f eth1-eth2_up.yaml
 Wait for the Policy to get applied:
 
 ```shell
-kubectl wait nncp eth1 eth2 --for condition=Available --timeout 2m
+kubectl wait nncp eth1-eth2 --for condition=Available --timeout 2m
 ```
 
 Both NICs are now back up and with assigned IPs:
--- a/docs/user-guide/eth1-eth2_up.yaml
+++ b/docs/user-guide/eth1-eth2_up.yaml
@@ -2,7 +2,7 @@
 apiVersion: nmstate.io/v1beta1
 kind: NodeNetworkConfigurationPolicy
 metadata:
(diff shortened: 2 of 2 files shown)


phoracek:
The intent of this example is to show that users can apply multiple standalone policies to configure all the interfaces, i.e. they don't need to put all the (unrelated) configuration into a single file.

Is this file with two policies causing issues to our tests?

rhrazdil:
@phoracek We agreed that there should be only one sequentially applied policy at a time, because when there are more, we cannot guarantee the order in which the enactments are applied.
Which is why we decided to add a validation hook, that will check that no such policies are in progress. 

Now when we apply the file with 2 (or more policies), the second and subsequent policies are denied by the hook

phoracek:
> @phoracek We agreed that there should be only one sequentially applied policy at a time, because when there are more, we cannot guarantee the order in which the enactments are applied.
> Which is why we decided to add a validation hook, that checks that no such policy is in progress when users create/update policies.
> 
> Now when we apply the file with 2 (or more policies), the second and subsequent policies are denied by the hook

My understanding was that we are to prevent users from applying the same policy while it is being applied. That should make sure that enactments are completed before we start the second rollout.

Why do we prevent 2 different policies from being applied? I'm afraid that would get users into unexpected situations where they apply their directory full of policies and it would fail to.

rhrazdil:
> My understanding was that we are to prevent users from applying the same policy while it is being applied. That should make sure that enactments are completed before we start the second rollout.

You're totally right, I misunderstood.
Closing this PR
/close

PR #663, opened 2020-12-15
WIP: Use varlink to communicate with a nmstate running as a daemon

qinqon:
Signed-off-by: Quique Llorente <ellorent@redhat.com>

**Is this a BUG FIX or a FEATURE ?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind enhancement

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**Release note**:

```release-note

```

PR #673, opened 2020-12-22
Retry NNSs Status checking at e2e tests

qinqon:
**Is this a BUG FIX or a FEATURE ?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind enhancement

**What this PR does / why we need it**:
Sometimes the currentState get changed by the system so NNS status get
refreshed, there is a test checking that NNS Status is not updated if
not needed, so in those cases the test is retried.

Example of network changes detected 
```
[nns] NNS LastSuccessfulUpdateTime
/tmp/knmstate/kubernetes-nmstate/test/e2e/handler/nns_update_timestamp_test.go:17
  when network configuration hasn't change
  /tmp/knmstate/kubernetes-nmstate/test/e2e/handler/nns_update_timestamp_test.go:28
    should not be updated [It]
    /tmp/knmstate/kubernetes-nmstate/test/e2e/handler/nns_update_timestamp_test.go:29

    Failed after 0.008s.
    Expected
        <string>: NodeNetworkStateStatus
    to match fields: {
    .CurrentState:
    	Expected
    	    <string>: "...on: fe80::/..."
    	to equal               |
    	    <string>: "...on: fd10:24..."
    .LastSuccessfulUpdateTime:
    	Expected
    	    <v1.Time>: {
    	        Time: 2021-01-14T13:36:40Z,
    	    }
    	to equal
    	    <v1.Time>: {
    	        Time: 2021-01-14T13:35:40Z,
    	    }
    }
    
```

**Special notes for your reviewer**:
Closes #670 

**Release note**:

```release-note
NONE
```

PR #682, opened 2021-01-12
Decrease NNS refresh frequency and remove dyn attr

qinqon:

>
> /kind bug
> /kind enhancement

Trying to refresh NNS every 5 seconds makes nmstatectl to call python
everytime consuming CPU resources, decreasing the frequency will reduce
CPU consumption.

Also increasing the refresh period to 1 minute make CI fail on timeout,
to overcome that we force NNS refresh if a NNCP is applied.

And filter gc-timer and hello-timer hey are already not reflecting
reality since we remove them before
compare them so the value that is present at NNS is not accurate, let's
just remove them from NNS so we can use FilterOut to compare states too,
so we don't compare filtered stuff like veth or calico.

Proper solution is to use varlink so python is already started up at
different container, this is done at #663.

**Release note**:

```release-note
Decrease NNS refresh frequency and remove dyn attr
```


qinqon:
Just one flaky test to fix, looks like some force refresh is being done in the middle of taking the NNSs possibly at Reconcile cycle pending from another test.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -326,6 +329,26 @@ func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Man
...
+	if err != nil {
+		log.Error(err, "failed retrieving NodeNetworkState to force refresh")
+		return
+	}
+	if nns.Labels == nil {
+		nns.Labels = map[string]string{}
+	}
+	nns.Labels[state.ForceRefreshLabel] = fmt.Sprintf("%d", time.Now().UnixNano())
+
+	err = r.Client.Update(context.Background(), nns)
```
rhrazdil, line 346: Shouldn't we care about conflicts? Or does a conflict mean that the nns was refreshed and thus no need to do so?

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -326,6 +329,26 @@ func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Man
...
+	nns := &nmstatev1beta1.NodeNetworkState{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: name}, nns)
+	if err != nil {
+		log.Error(err, "failed retrieving NodeNetworkState to force refresh")
+		return
+	}
+	if nns.Labels == nil {
+		nns.Labels = map[string]string{}
+	}
+	nns.Labels[state.ForceRefreshLabel] = fmt.Sprintf("%d", time.Now().UnixNano())
```
rhrazdil, line 344: this label should be removed when the nns is updated?

controllers/nodenetworkstate_controller.go
```diff
@@ -102,3 +104,17 @@ func (r *NodeNetworkStateReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		WithEventFilter(onDeleteForThisNode).
 		Complete(r)
 }
+
+func isForceRefresh(updateEvent event.UpdateEvent) bool {
```
rhrazdil, line 108: Just an idea, what do you think about calling this `shouldForceRefresh`, as there is additional logic and not just simple label check

rhrazdil (changes requested):
The second commit `Force NNS refresh on NNCP update` would deserve a commit msg

controllers/nodenetworkconfigurationpolicy_controller.go
qinqon: Kind of, since in a minute is going to be refreshed I didn't want to add conflicts here, but I can try.

controllers/nodenetworkconfigurationpolicy_controller.go
qinqon: The first idea was about a label without value and remove at the end of Reconcile on NNS but if the Reconcile fail the label is going to be there and there is not going to be a way to detect force refresh if we compare only the label existence, that's why I just keep it with a timestamp so we know whene the NNS was force refreshed last time (not a bad thing to have though).

RamLavi (changes requested):
I don't get the motivation for the last 2 commits:
`Force NNS refresh on NNCP update`  and `Ignore hello-timer field`, can you elaborate on the commit message?

qinqon:
> I don't get the motivation for the last 2 commits:
> `Force NNS refresh on NNCP update` and `Ignore hello-timer field`, can you elaborate on the commit message?

Originally I just increased to 1 minute the frequency period for NNS refresh but doing so make CI quite slow and ends up with timeouts and we don't want to wait more than one hour there (think also on CNAO tier1 tests).

Most of the e2e tests that wait on NNS state changes are also applying a NNCP so forcing a NNS refresh when NNCP is applied make those test has similar latencies than before, also it make sense to refresh NNS after a success NNCP.

controllers/nodenetworkstate_controller.go
qinqon: Good idea, I will change it after fixing flakiness at test.

qinqon:
Flaky appears, should be fixed by https://github.com/nmstate/kubernetes-nmstate/pull/673.

qinqon:
Looks like some calico interface are intermitent (maybe a pod is restarting or the like) we should totally skip filtered interfaces from comparation and also filter routes from those interfaces too.

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/7ce6e84e7ec9946754a317430c0ffdfa1126dbee..9cab7cc245bfd7cef96007acb5b6aaa061456e07): Squashed the 4 commits since force NNS refresh is ok with reviewers.

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/9cab7cc245bfd7cef96007acb5b6aaa061456e07..72338ca48312c080ea37c7a4eeb3cd04e6a595f0): Address @rhrazdil comments

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Gotcha, thanks

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Undrestood, it's probably not necessary then

pkg/state/filter.go
```diff
@@ -1,21 +1,73 @@
...
-	gcTimerRexp = regexp.MustCompile(` *gc-timer: *[0-9]*\n`)
+	timerRexp            = regexp.MustCompile(` *(gc|hello)-timer: *[0-9]*\n`)
+	interfacesFilterGlob glob.Glob
 )
 
+func init() {
+	if !environment.IsHandler() {
+		return
+	}
+	interfacesFilter, isSet := os.LookupEnv("INTERFACES_FILTER")
```
rhrazdil, line 23: I think you mentioned using a configMap instead of envVar?

qinqon:
There is still something changing in the network configuration at nodes, maybe we can just ensure that we test this at unit test and remove the e2e tests, we don't know what else can change on the future.

phoracek:
> There is still something changing in the network configuration at nodes, maybe we can just ensure that we test this at unit test and remove the e2e tests, we don't know what else can change on the future.

More we can cover with unit tests, the better. However, we should know what is the issue you are seeing here to make sure that we do not refresh more than necessary in prod.

qinqon:
/let's try to flaky to show so we get the diff.

pkg/state/filter_test.go
```diff
@@ -0,0 +1,340 @@
...
+	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
+)
+
+var _ = Describe("FilterOut", func() {
+	var (
+		state, filteredState nmstate.State
+		interfacesFilterGlob glob.Glob
+	)
+
+	Context("whete there is a linux bridge with gc-timer and hello-time", func() {
```
rhrazdil, line 17:
```suggestion
	Context("when there is a linux bridge with gc-timer and hello-time", func() {
```

rhrazdil (approved):
Looks good, if an issue appears, we'll see the nns diff and can act accordingly.
/lgtm

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/1ec26c199f09cc748da7dbc575e5cbe60d48c5d8..27d5ebdca7c7a9568f2e052d7d8b3d0eaebe8799): Addressed @rhrazdil comments

pkg/helper/client.go
```diff
@@ -116,6 +117,11 @@ func UpdateCurrentState(client client.Client, nodeNetworkState *nmstatev1beta1.N
 		stateToReport = observedState
 	}
 
+	if state.RemoveDynamicAttributesFromStruct(stateToReport) == state.RemoveDynamicAttributesFromStruct(nodeNetworkState.Status.CurrentState) {
+		log.Info("Skipping NNS update, stateToReport == currentState")
```
phoracek: Let's speak human. The node network state has not changed?

pkg/state/labels.go
```diff
@@ -0,0 +1,5 @@
+package state
+
+const (
+	ForceRefreshLabel = "nmstate.io/force-nns-refresh"
```
phoracek, line 4: IMO this should live in the module of the feature. Having a module to catch all labels, no matter what is their purpose, is not much different from having utils.go. I'd move it to where it is used and make sure it is not exposed to underlying implementation - only to modules concerned about controllers

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -42,8 +42,9 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
 			}
 		})
 	})
-	Context("when network configuration changed", func() {
+	Context("when network configuration is changed by a NNCP", func() {
 		BeforeEach(func() {
+			// We want to test all the NNS so we apply poclies to masters and workes
```
phoracek, line 51: policies
workers

pkg/helper/client.go
phoracek: Use NodeNetworkState instead of NNS. Shorts are bad

pkg/state/filter.go
```diff
@@ -38,6 +38,7 @@ func RemoveDynamicAttributesFromStruct(state shared.State) string {
 }
 
 func FilterOut(currentState shared.State) (shared.State, error) {
+
```
phoracek, line 39: Redundant

pkg/state/filter.go
```diff
@@ -68,5 +69,5 @@ func filterOut(currentState shared.State, interfacesFilterGlob glob.Glob) (share
 		return currentState, err
 	}
 
-	return shared.State{Raw: filteredState}, nil
+	return shared.NewState(RemoveDynamicAttributes(string(filteredState))), nil
```
phoracek: The commit message does not match the content here, does it?

pkg/state/filter_test.go
```diff
@@ -14,23 +14,172 @@ var _ = Describe("FilterOut", func() {
 		interfacesFilterGlob glob.Glob
 	)
 
+	Context("when there is a linux bridge with gc-timer and hello-timer", func() {
```
phoracek, line 17: States you have used there have way more stuff in them than these two attributes. That makes it terrifying to read through. Any reason to keep all the options there?

pkg/helper/client.go
```diff
@@ -86,27 +85,20 @@ func CreateOrUpdateNodeNetworkState(client client.Client, node *corev1.Node, nam
...
-	stateToReport, err := state.FilterOut(observedState)
-	if err != nil {
-		log.Error(err, "failed filtering out interfaces from NodeNetworkState, keeping orignal content, please fix the glob")
-		stateToReport = observedState
-	}
-
-	if state.RemoveDynamicAttributesFromStruct(stateToReport) == state.RemoveDynamicAttributesFromStruct(nodeNetworkState.Status.CurrentState) {
-		log.Info("Skipping NNS update, stateToReport == currentState")
+	if observedState.String() == nodeNetworkState.Status.CurrentState.String() {
+		log.Info("Skipping NNS update, observedState == currentState")
```
phoracek, line 94: Please use human language and do not use shorts

phoracek (changes requested):

controllers/nodenetworkstate_controller.go
```diff
@@ -102,3 +104,17 @@ func (r *NodeNetworkStateReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		WithEventFilter(onDeleteForThisNode).
 		Complete(r)
 }
+
+func shouldForceRefresh(updateEvent event.UpdateEvent) bool {
+	newForceRefresh, ok := updateEvent.MetaNew.GetLabels()[state.ForceRefreshLabel]
+	if !ok {
+		// There is no force refresh label so we don't force it
```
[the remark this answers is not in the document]
qinqon: Renaming the "ok" vars so if statement is clear.

pkg/state/labels.go
qinqon: Moved to `controllers/labels.go` since there are two controllers that use it.

pkg/state/filter.go
qinqon: The commit message from this line is `Filter out routes and dynamic attributes` I would say that it's correct.

pkg/state/filter_test.go
qinqon: Just copy paste from real linux-bridge, I will clean that up, at the end I just one to keep some to be sure that only the specific ones are deleted.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -335,17 +334,17 @@ func (r *NodeNetworkConfigurationPolicyReconciler) forceNNSRefresh(name string)
...
 	if nns.Labels == nil {
 		nns.Labels = map[string]string{}
 	}
-	nns.Labels[state.ForceRefreshLabel] = fmt.Sprintf("%d", time.Now().UnixNano())
+	nns.Labels[forceRefreshLabel] = fmt.Sprintf("%d", time.Now().UnixNano())
 
 	err = r.Client.Update(context.Background(), nns)
 	if err != nil {
-		log.Error(err, "failed adding force refresh label to NodeNetworkState")
+		log.Error(err, "WARNING: failed forcing NNS refresh, it will be refreshed after regular period")
```
phoracek, line 347: This would still apprear in logs prefixed with ERROR

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -335,17 +334,17 @@ func (r *NodeNetworkConfigurationPolicyReconciler) forceNNSRefresh(name string)
 	nns := &nmstatev1beta1.NodeNetworkState{}
 	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: name}, nns)
 	if err != nil {
-		log.Error(err, "failed retrieving NodeNetworkState to force refresh")
+		log.Error(err, "WARNING: failed retrieving NodeNetworkState to force refresh, it will be refreshed after regular period")
```
phoracek, line 337: Again, this should not be seen as `ERROR: WARNING: ` in the logs, that would be misleading.

When somebody opens a BZ and shares nmstate logs. I want to be able to search it for "ERROR" and the first line that pops up must be the cause of the bug.

controllers/labels.go
```diff
@@ -0,0 +1,5 @@
+package controllers
```
phoracek, line 1: Please don't keep a commit with my name on it. These fixes should be distributed in their relevant commits.

phoracek (changes requested):

qinqon:
Squashed commits and addressed @phoracek comments

phoracek (approved):
/lgtm
/approve

Issue #684, opened 2021-01-22
Node controller predicate does not filter at watcher


qinqon:
**What happened**:
At big clusters every handler is retrieving all the nodes and that hits hard the kube-apiserver nodes.
**What you expected to happen**:
controller-runtime filtering using predicates so it retrieves only the ones filetered.

**How to reproduce it (as minimally and precisely as possible)**:
Counting the number of Node Get operations at running kubernetes-nmstate.

**Anything else we need to know?**:
This is a bug at controller-runtime  this is a bug at controller-runtime https://github.com/kubernetes-sigs/controller-runtime/issues/244.

**Environment**:

- `NodeNetworkState` on affected nodes (use `kubectl get nodenetworkstate <node_name> -o yaml`):
- Problematic `NodeNetworkConfigurationPolicy`:
- kubernetes-nmstate image (use `kubectl get pods --all-namespaces -l app=kubernetes-nmstate -o jsonpath='{.items[0].spec.containers[0].image}'`):
- NetworkManager version (use `nmcli --version`)
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Others:

phoracek:
It does filter it there now!

PR #685, opened 2021-01-22
Jitter the nns refresh period

qinqon:

>
/kind bug
> /kind enhancement

Refreshing NNS at the same time on all the nodes can hit apiserver hard,
let's jitter it a little so refreshing is distributed over time.

This will reduce the impact of https://github.com/nmstate/kubernetes-nmstate/issues/684

**Release note**:

```release-note
Jitter the nns refresh period
```

### DIFF
--- a/controllers/node_controller_test.go
+++ b/controllers/node_controller_test.go
@@ -44,6 +44,9 @@ var _ = Describe("Node controller reconcile", func() {
 				Name: existingNodeName,
 			},
 		}
+		expectRequeueAfterIsSetWithNetworkStateRefresh = func(result ctrl.Result) {
+			ExpectWithOffset(1, result.RequeueAfter).To(BeNumerically("~", nmstatenode.NetworkStateRefresh, float64(nmstatenode.NetworkStateRefresh)*nmstatenode.NetworkStateRefreshMaxFactor))
+		}
 	)
 	BeforeEach(func() {
 		reconciler = NodeReconciler{}
@@ -109,7 +112,7 @@ routes:
 		It("should not call nmstateUpdater and return a Result with RequeueAfter set", func() {
 			result, err := reconciler.Reconcile(request)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(result).To(Equal(reconcile.Result{RequeueAfter: nmstatenode.NetworkStateRefresh}))
+			expectRequeueAfterIsSetWithNetworkStateRefresh(result)
 		})
 	})
 	Context("when node is not found", func() {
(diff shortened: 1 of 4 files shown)


pkg/node/refresh.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	networkStateRefresh = time.Minute
+)
+
+// NodeNetworkStateRefresh add some jitter to to the refresh rate so it does
+// not hit apiserver at the same time, the jittering is implemented using
```
phoracek, line 14: It is clear how it's implemented from the body

controllers/node_controller_test.go
```diff
@@ -44,6 +44,10 @@ var _ = Describe("Node controller reconcile", func() {
 				Name: existingNodeName,
 			},
 		}
+		expectRequeueAfterIsSet = func(result ctrl.Result) {
+			ExpectWithOffset(1, result.RequeueAfter).To(BeNumerically(">=", time.Minute))
```
phoracek, line 48: Does not jitter go in both ways?

phoracek (changes requested):

pkg/node/refresh.go
qinqon: But you don't see the body at godocs.

controllers/node_controller_test.go
```diff
@@ -44,6 +44,10 @@ var _ = Describe("Node controller reconcile", func() {
 				Name: existingNodeName,
 			},
 		}
+		expectRequeueAfterIsSet = func(result ctrl.Result) {
```
rhrazdil, line 47: The func name might be more descriptive, it makes me ask after **what** is set?

controllers/node_controller_test.go
```diff
@@ -44,6 +44,10 @@ var _ = Describe("Node controller reconcile", func() {
...
+			ExpectWithOffset(1, result.RequeueAfter).To(BeNumerically(">=", time.Minute))
+			ExpectWithOffset(1, result.RequeueAfter).To(BeNumerically("<=", float64(time.Minute)+(float64(time.Minute)*0.1)))
```
rhrazdil, line 49: I think it would be safer to use variable for the jitter value, if we decided to increase the jitter interval, then the test may 'get lucky' and flake later randomly.

rhrazdil (changes requested):

pkg/node/refresh.go
phoracek: Do we apply this rule to other bigger functions?

I won't fight for this, but linking documentation of the implementation feels like a burden on maintenance.

qinqon: I will copy paste what Jitter does.

qinqon:
[force-refresh](https://github.com/nmstate/kubernetes-nmstate/compare/ba04ef02a2234f209c436b4e7282bb6d967741f1..dab4c7b542667d39db5e6f334e22140e27258db7): Address @phoracek comments

pkg/node/refresh.go
```diff
@@ -0,0 +1,18 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	networkStateRefresh = time.Minute
+)
+
+// NodeNetworkStateRefresh add some jitter to to the refresh rate so it does
+// not hit apiserver at the same time. it returns random time.Duration in range
+// [60-66] seconds.
```
phoracek, line 15: The refresh interval is taken from a constant. It should not be documented in the function. It would get out of sync and frankly, the default value should be irrelevant while you are coding. Things should work no matter if it is a second on an hour.

rhrazdil (approved):
/lgmt

rhrazdil (approved):
/lgtm

PR #690, opened 2021-02-01
Skip linux-bridge filtering if no "bridge" section

qinqon:
We have found that at some envs the "bridge" section from linux-bridge
is not present in case the bridge is down, this change ignore the
filtering in this is the case to preveng a segmentation fault.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/state/filter.go
+++ b/pkg/state/filter.go
@@ -30,7 +30,16 @@ func FilterOut(currentState shared.State) (shared.State, error) {
 }
 
 func filterOutRoutes(kind string, state map[string]interface{}, interfacesFilterGlob glob.Glob) {
-	routes := state["routes"].(map[string]interface{})
+	routesRaw, hasRoutes := state["routes"]
+	if !hasRoutes {
+		return
+	}
+
+	routes, ok := routesRaw.(map[string]interface{})
+	if !ok {
+		return
+	}
+
 	routesByKind := routes[kind].([]interface{})
 
 	if routesByKind == nil {
@@ -58,9 +67,24 @@ func filterOutDynamicAttributes(iface map[string]interface{}) {
 		return
 	}
 
-	bridge := iface["bridge"].(map[string]interface{})
+	bridgeRaw, hasBridge := iface["bridge"]
+	if !hasBridge {
+		return
+	}
+	bridge, ok := bridgeRaw.(map[string]interface{})
+	if !ok {
+		return
+	}
 
-	options := bridge["options"].(map[string]interface{})
+	optionsRaw, hasOptions := bridge["options"]
+	if !hasOptions {
+		return
+	}
+	options, ok := optionsRaw.(map[string]interface{})
+	if !ok {
+		return
+	}
+
 	delete(options, "gc-timer")
 	delete(options, "hello-timer")
 }
--- a/pkg/state/filter_test.go
+++ b/pkg/state/filter_test.go
@@ -307,4 +307,28 @@ routes:
 			Expect(returnedState).To(MatchYAML(filteredState))
 		})
 	})
+	Context("when there is a linux bridge without 'bridge' options because is down", func() {
+		BeforeEach(func() {
+			state = nmstate.NewState(`
+interfaces:
+- name: br1
+  type: linux-bridge
+  state: down
+`)
+
+			filteredState = nmstate.NewState(`
+interfaces:
+- name: br1
+  type: linux-bridge
+  state: down
+`)
+			interfacesFilterGlob = glob.MustCompile("")
+		})
+		It("should keep the bridge as it is", func() {
+			returnedState, err := filterOut(state, interfacesFilterGlob)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(returnedState).To(MatchYAML(filteredState))
+		})
+	})
+
 })


pkg/nmstatectl/remote.go
```diff
@@ -0,0 +1,68 @@
+package nmstatectl
```
phoracek, line 1: This should not be a part of this PR, right?

qinqon: Removed, this is from the PoC.

pkg/state/filter.go
```diff
@@ -58,9 +58,24 @@ func filterOutDynamicAttributes(iface map[string]interface{}) {
 		return
 	}
 
-	bridge := iface["bridge"].(map[string]interface{})
+	bridgeIface, ok := iface["bridge"]
```
phoracek, line 61: use `hasBridge` instead of `ok`. Then the following condition makes more sense.

phoracek, line 61: How is this `iface`? We just know that it has the type of an empty interface. Maybe `bridgeRaw` would be slightly better?

pkg/state/filter_test.go
```diff
@@ -307,4 +307,66 @@ routes:
...
+  state: down
+  ipv4:
+    enabled: false
+  ipv6:
+    enabled: false
+  lldp:
+    enabled: false
+  mac-address: A2:EE:84:7B:42:4C
+  mtu: 1500
+routes:
```
phoracek, line 328: Please drop everything that is not necessary for this test. We do not care about routes, do we?

phoracek (changes requested):

pkg/state/filter_test.go
```diff
@@ -307,4 +307,66 @@ routes:
 			Expect(returnedState).To(MatchYAML(filteredState))
 		})
 	})
+	Context("when there is a linux bridge without 'bridge' options because is down", func() {
+		BeforeEach(func() {
+			state = nmstate.NewState(`
+interfaces:
+- name: eth1
```
rhrazdil, line 314: the `ethernet` interface seems not relevant for the test case either

pkg/state/filter_test.go
```diff
@@ -307,4 +307,44 @@ routes:
...
+  type: linux-bridge
+  state: down
+  ipv4:
+    enabled: false
+  ipv6:
+    enabled: false
+  lldp:
+    enabled: false
+  mac-address: A2:EE:84:7B:42:4C
+  mtu: 1500
```
phoracek, line 324: None of this is needed to test that you fixed the issue, is it?

qinqon: I have replicate what we have in the env that is failing the bridge is like this.

pkg/state/filter.go
```diff
@@ -30,7 +30,11 @@ func FilterOut(currentState shared.State) (shared.State, error) {
 }
 
 func filterOutRoutes(kind string, state map[string]interface{}, interfacesFilterGlob glob.Glob) {
-	routes := state["routes"].(map[string]interface{})
+	routesRaw, hasRoutes := state["routes"]
+	if !hasRoutes {
+		return
+	}
+	routes := routesRaw.(map[string]interface{})
```
phoracek, line 37: Why don't you test whether the casting went right here, but you do check it further down?

phoracek (changes requested):

pkg/state/filter.go
qinqon: In the case of routes there is always "config" and "running"

pkg/state/filter_test.go
phoracek: But then it is not a test for missing bridge. It is a test for missing bridge *and* another issues that we haven't thought about and may be in the code. We may catch a bug by chance here, but it feels really random and non-fitting for this specific test. You may argue that we should include snapshot of the whole state of a very complicated node to catch more potential issues

qinqon: But we want to test with sane data right ? this attributes are present at bridge like this when a bridge is down.

pkg/state/filter.go
phoracek: How is it relevant to casting? Do you mean that in case of "bridge" for instance we can get nil?

qinqon: Ahh right, misunderstood, added it.

pkg/state/filter_test.go
qinqon: Ok, will just keep name, type and state.

phoracek: You are writting a unit test. The test should be concerned only about a specific subpart of the system. If we need the whole state to be populated to get one specific function running, then we have an issue in our architecture and end-up with ton of unmaintainable yamls.

We had a bug with missing bridge. You should write a minimal reproducer and then a fix.

If you expect that e.g. "lldp" may affect run of the function, then we need a test for invalid lldp.

phoracek (commented):
/approve

Looks good. Would you lgtm @rhrazdil if all seems fine to you?

rhrazdil (commented):
/lgtm

enp0s3:
@phoracek @qinqon sorry but it's not clear to me what do you mean by "bridge is down" - was that administratively down or was the link of the bridge down? Perhaps I am not familiar with the states

PR #701, opened 2021-02-24
Explicitly unmanage veth interfaces

phoracek:

/kind bug

nmstate 0.3 introduces a bug where all veths attached to nmstatectl
configured bridge turn into "managed" by NetworkManager. This was not
the case in 0.2.

Due to this regression, NetworkManager deliberatelly detaches veth
ifaces from the bridge and by doing that it disconnectecs Pods/VMs from
the network.

With this change, we explicitly set veth interfaces as unmanaged.

Important: With this change, it is no longer possible to change
configuration of bridges that have veths attached.

This should be reverted once https://bugzilla.redhat.com/show_bug.cgi?id=1932247 becomes available.

**Release note**:

```release-note
Explicitly set host veths as unmanaged
```

### DIFF
--- a/build/bin/unmanaged-veth
+++ b/build/bin/unmanaged-veth
@@ -1,0 +1,6 @@
+#!/bin/bash -xe
+
+echo '[keyfile]
+unmanaged-devices=interface-name:veth*
+' > /host/etc/NetworkManager/conf.d/001-cnv-unmanaged-veth.conf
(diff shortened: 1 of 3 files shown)


qinqon:
/lgtm
/approve

Ugly hack that will unblock upgrades, this will be removed after fix at nmstate.

PR #708, opened 2021-03-01
state, filter: Treat numeric-only iface names

EdDev:

/kind bug

When processing the filtering of the interfaces status, the interfaces
with names that include only numeric characters are interpreted as
float64.

In order to (partially) support such interfaces, make sure to detect them
and avoid their filtering. I.E. these interfaces cannot be filtered-out.

Note: This solution is partial due to the improper interpretation of
special numeric (only) names such as (123.0 and 0xff). These will
wrongly get interpreted as 123 and 255.

**Release note**:
```release-note
Support interface names with numeric-only characters.
Names with special characters, like `1.0` or `0xff` are not supported.
```

### DIFF
--- a/pkg/state/filter.go
+++ b/pkg/state/filter.go
@@ -7,11 +7,14 @@ import (
 	"github.com/nmstate/kubernetes-nmstate/api/shared"
 	"github.com/nmstate/kubernetes-nmstate/pkg/environment"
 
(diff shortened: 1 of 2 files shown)


EdDev:
I do not really like this solution, it looks like a workaround. Trying something else...
Putting this one on hold for now, lets see if I manage to put something better on the table.

EdDev:
See #709 for a better alternative to this one.

PR #709, opened 2021-03-02
state, filter: Fix the interpretation of ifaces names

EdDev:
/kind bug

In scenarios where an interface has only numeric characters, the
application panics with a failure to extract a string type from a Go
interface (which is actually a float64).

In order to fix the problem, the parsing of the interface name is performed
using a dedicated interface-state structure.

In addition, the representation of special numeric values that an interface name can have is also fixed, such that `0xfe` can be correctly represented back and not shown as an integer (`254`).

---

In order to support this, a type structure that represents the basic state root items has been used.
This structure includes:
- interfaces
- routes:
  - config
  - running

In order to keep the generic nature of these items content,
they use as value a generic Go interface{}.

**Release note**:
```release-note
Support interface names with numeric values (e.g. 123, 1.0, 0xfe).
```

### DIFF
--- a/pkg/state/filter.go
+++ b/pkg/state/filter.go
@@ -29,32 +29,16 @@ func FilterOut(currentState shared.State) (shared.State, error) {
 	return filterOut(currentState, interfacesFilterGlobFromEnv)
 }
 
-func filterOutRoutes(kind string, state map[string]interface{}, interfacesFilterGlob glob.Glob) {
-	routesRaw, hasRoutes := state["routes"]
-	if !hasRoutes {
-		return
-	}
-
-	routes, ok := routesRaw.(map[string]interface{})
-	if !ok {
-		return
-	}
-
-	routesByKind := routes[kind].([]interface{})
-
-	if routesByKind == nil {
-		return
-	}
-
+func filterOutRoutes(routes []interface{}, interfacesFilterGlob glob.Glob) []interface{} {
 	filteredRoutes := []interface{}{}
-	for _, route := range routesByKind {
+	for _, route := range routes {
 		name := route.(map[string]interface{})["next-hop-interface"]
 		if !interfacesFilterGlob.Match(name.(string)) {
 			filteredRoutes = append(filteredRoutes, route)
 		}
 	}
 
-	state["routes"].(map[string]interface{})[kind] = filteredRoutes
+	return filteredRoutes
 }
 
 func filterOutDynamicAttributes(iface map[string]interface{}) {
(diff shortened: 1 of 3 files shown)


EdDev:
change: Fixed gofmt

pkg/state/type.go
```diff
@@ -0,0 +1,19 @@
...
+	Running []interface{} `json:"running"`
+}
+
+type InterfacesState struct {
+	Interfaces []Interface `json:"interfaces"`
+}
+
+type Interface struct {
+	Name string `json:"name"`
+}
```
qinqon, line 19: We don't need this to be public "yet"

EdDev: You mean the members or the structures?

EdDev: I guess you meant the struct. I changed it.

EdDev:
change: Do not expose the new state types

qinqon (commented):
Just a question

pkg/state/filter.go
```diff
@@ -89,31 +73,37 @@ func filterOutDynamicAttributes(iface map[string]interface{}) {
...
 		return currentState, err
 	}
 
-	filterOutInterfaces(state, interfacesFilterGlob)
-	filterOutRoutes("running", state, interfacesFilterGlob)
-	filterOutRoutes("config", state, interfacesFilterGlob)
+	var ifacesState interfacesState
+	if err := yaml.Unmarshal(currentState.Raw, &ifacesState); err != nil {
+		return currentState, err
+	}
```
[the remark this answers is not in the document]
qinqon: Looks like the yaml conunterpart works fine https://play.golang.org/p/ocI6B0FhHLw

pkg/state/filter.go
```diff
@@ -89,31 +73,37 @@ func filterOutDynamicAttributes(iface map[string]interface{}) {
 	delete(options, "hello-timer")
 }
 
-func filterOutInterfaces(state map[string]interface{}, interfacesFilterGlob glob.Glob) {
-	interfaces := state["interfaces"]
+func filterOutInterfaces(ifaces []interface{}, ifacesNames []interfaceState, interfacesFilterGlob glob.Glob) []interface{} {
```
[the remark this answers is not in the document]
qinqon: Using the yaml representation of []byte make it work https://play.golang.org/p/XxFTjYlJNra, this way we don't need to pass two slices and we have transparent placeholder for the attributes we don't care.

pkg/state/filter.go
```diff
@@ -102,3 +107,18 @@ func filterOut(currentState shared.State, interfacesFilterGlob glob.Glob) (share
...
+// Numeric values, including the ones with a base prefix (e.g. 0x123) should be stringify.
+func normalizeInterfacesNames(rawState []byte, state *rootState) error {
+	var stateForNormalization rootState
+	if err := goyaml.Unmarshal(rawState, &stateForNormalization); err != nil {
+		return err
+	}
+	for i, iface := range stateForNormalization.Interfaces {
+		if isNumeric(iface.Name) {
+			state.Interfaces[i].Name = iface.Name
+		}
```
qinqon, line 121: I think checking `isNumeric` has the same cost as the Name copy, so we can just copy all the names and then we don't need `isNumeric`

pkg/state/filter.go
```diff
@@ -89,35 +74,51 @@ func filterOutDynamicAttributes(iface map[string]interface{}) {
...
+	if err := yaml.Unmarshal(currentState.Raw, &state); err != nil {
 		return currentState, err
 	}
 
-	filterOutInterfaces(state, interfacesFilterGlob)
-	filterOutRoutes("running", state, interfacesFilterGlob)
-	filterOutRoutes("config", state, interfacesFilterGlob)
+	if err := normalizeInterfacesNames(currentState.Raw, &state); err != nil {
+		return currentState, err
+	}
```
qinqon, line 96: Looks like the `gopkg.in/yaml.v2` has similar functionalit to `MarshalJSON` and `UnmarshalJSON` 

https://pkg.go.dev/gopkg.in/yaml.v2#Marshaler
https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshaler

So using those we can replace totally `sigs.k8s.io/yaml` with `gopkg.in/yaml.v2` so we have the normalization for free and we can also copy the rest of the attributes on those functions.

qinqon (changes requested):
Good work I know how much time to end with this kind of solutions, added some coments to simplify stuff.

pkg/state/filter.go
EdDev: It has different behaviors and produces in some cases `null`. But the main problem is that we need to replace it all over the project.
I tried to do this under filtering but got some nasty issues.

I suggest exploring this on a separate PR.

pkg/state/filter.go
EdDev: Yes, right. I had a problem with some double quoting, but with the latest version it is ok.
Done

pkg/state/type.go
```diff
@@ -1,11 +1,47 @@
...
+	if err := yaml.Unmarshal(b, &i.Data); err != nil {
+		return fmt.Errorf("failed Unmarshaling b: %w", err)
+	}
+
+	var ifaceFields interfaceFields
+	if err := yaml.Unmarshal(b, &ifaceFields); err != nil {
+		return fmt.Errorf("failed Unmarshaling raw: %w", err)
+	}
+	i.Data["name"] = ifaceFields.Name
+	i.Name = ifaceFields.Name
```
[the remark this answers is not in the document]
EdDev: Done

qinqon (approved):
/hold

First I want to see how so nmstate team is reporting that quotes are included and why we don't see it here.

qinqon:
Looks like problem is just with float64

```
ip link add 60e+02 type dummy
name: 60e+02
```

@EdDev can you add a unit test for that ?

EdDev:
> Looks like problem is just with float64
> 
> ```
> ip link add 60e+02 type dummy
> name: 60e+02
> ```
> 
> @EdDev can you add a unit test for that ?

I have changed a bit the commits to show better what has been done and what each fixes.

qinqon:
/lgtm
/approve
/unhold

We will do a follow up to try to use just one yaml parser.

Issue #670, opened 2020-12-22
[flaky CI] [nns] NNS LastSuccessfulUpdateTime when network configuration hasn't change should not be updated


qinqon:
**What happened**:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/661/pull-kubernetes-nmstate-e2e-handler-k8s/1341025954180370432

```

[nns] NNS LastSuccessfulUpdateTime when network configuration hasn't change should not be updated expand_less | 15s
-- | --
/tmp/knmstate/kubernetes-nmstate/test/e2e/handler/nns_update_timestamp_test.go:29 Failed after 0.006s. Expected     <string>: NodeNetworkStateStatus to match fields: { .CurrentState: 	Expected 	    <string>: "...on: fd10:24..." 	to equal               \| 	    <string>: "...on: fe80::/..." .LastSuccessfulUpdateTime: 	Expected 	    <v1.Time>: { 	        Time: 2020-12-21T14:32:38Z, 	    } 	to equal 	    <v1.Time>: { 	        Time: 2020-12-21T14:32:33Z, 	    } }


```

/kind bug
/triage build-watcher

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- `NodeNetworkState` on affected nodes (use `kubectl get nodenetworkstate <node_name> -o yaml`):
- Problematic `NodeNetworkConfigurationPolicy`:
- kubernetes-nmstate image (use `kubectl get pods --all-namespaces -l app=kubernetes-nmstate -o jsonpath='{.items[0].spec.containers[0].image}'`):
- NetworkManager version (use `nmcli --version`)
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Others:

nelsonspbr:
Again: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/692/pull-kubernetes-nmstate-e2e-handler-k8s-future/1356945962311880705

qinqon:
The diff on states
https://github.com/nmstate/kubernetes-nmstate/pull/688

xpivarc:
[...]

PR #711, opened 2021-03-03
Check nns timestamp depend on currnetState

qinqon:

>
/kind bug
> /kind enhancement

Closes https://github.com/nmstate/kubernetes-nmstate/issues/670

The NNS timestamp e2e tests that check the update timestamp is not
changed if no NNCP is applied is failing since node network
configuration changes sometimes without manual intervention. This change
check if update timestamp is different if current state has change and
if is the same if current state has not changed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/nns_update_timestamp_test.go
+++ b/test/e2e/handler/nns_update_timestamp_test.go
@@ -6,13 +6,9 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	. "github.com/onsi/gomega/gstruct"
-
-	"github.com/andreyvit/diff"
 
 	"k8s.io/apimachinery/pkg/types"
 
-	"github.com/nmstate/kubernetes-nmstate/api/shared"
 	nmstatev1beta1 "github.com/nmstate/kubernetes-nmstate/api/v1beta1"
 	nmstatenode "github.com/nmstate/kubernetes-nmstate/pkg/node"
 )
@@ -28,24 +24,46 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
 			originalNNSs[node] = nodeNetworkState(key)
 		}
 	})
-	Context("when network configuration hasn't change", func() {
-		It("should not be updated", func() {
-			for node, originalNNS := range originalNNSs {
-				// Give enough time for the NNS to be updated (3 interval times)
-				timeout := 3 * nmstatenode.NetworkStateRefresh
-				key := types.NamespacedName{Name: node}
+	Context("when network configuration is not changed by a NNCP", func() {
+		It("nns should not be updated after reconcile", func() {
+			// Give enough time for the NNS to be reconcile(2 interval times)
+			interval := 2 * nmstatenode.NetworkStateRefresh
+			timeout := 4 * interval
+			Eventually(func() error {
+				for node, originalNNS := range originalNNSs {
+					key := types.NamespacedName{Name: node}
+					currentStatus := nodeNetworkState(key).Status
+					originalStatus := originalNNS.Status
+					if currentStatus.CurrentState.String() == originalStatus.CurrentState.String() {
+						By(fmt.Sprintf("Check LastSuccessfulUpdateTime changed at %s", node))
+						Expect(currentStatus.LastSuccessfulUpdateTime).To(Equal(originalStatus.LastSuccessfulUpdateTime))
+					} else {
+						return fmt.Errorf("Network configuration changed, sending and error to retry")
+					}
+				}
+				return nil
+			}, timeout, interval).Should(Succeed())
+		})
+	})
+	Context("when network configuration is changed externally", func() {
+		expectedDummyName := "dummy0"
 
-				Consistently(func() shared.NodeNetworkStateStatus {
-					return nodeNetworkState(key).Status
-				}, timeout, time.Second).Should(MatchAllFields(Fields{
-					"CurrentState":             WithTransform(shared.State.String, Equal(originalNNS.Status.CurrentState.String())),
-					"LastSuccessfulUpdateTime": Equal(originalNNS.Status.LastSuccessfulUpdateTime),
-					"Conditions":               Equal(originalNNS.Status.Conditions),
-				}), func() string {
-					return fmt.Sprintf("currentState diff: \n%s", diff.LineDiff(originalNNS.Status.CurrentState.String(), nodeNetworkState(key).Status.CurrentState.String()))
-				})
+		BeforeEach(func() {
+			createDummyConnectionAtAllNodes(expectedDummyName)
+		})
+		AfterEach(func() {
+			deleteConnectionAndWait(allNodes, expectedDummyName)
+		})
+		It("should update it with according to network state refresh duration", func() {
+			for node, originalNNS := range originalNNSs {
+				By(fmt.Sprintf("Checking timestamp against original one %s", originalNNS.Status.LastSuccessfulUpdateTime))
+				Eventually(func() time.Time {
+					currentNNS := nodeNetworkState(types.NamespacedName{Name: node})
+					return currentNNS.Status.LastSuccessfulUpdateTime.Time
+				}, 2*nmstatenode.NetworkStateRefresh, 10*time.Second).Should(BeTemporally(">", originalNNS.Status.LastSuccessfulUpdateTime.Time), "should update it at %s", node)
 			}
 		})
+
 	})
 	Context("when network configuration is changed by a NNCP", func() {
 		BeforeEach(func() {
--- a/test/e2e/handler/nodes_test.go
+++ b/test/e2e/handler/nodes_test.go
@@ -27,28 +27,19 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
 			})
 		})
 		Context("and new interface is configured", func() {
-			var (
-				expectedDummyName = "dummy0"
-			)
+			expectedDummyName := "dummy0"
+
 			BeforeEach(func() {
-				createDummyAtNodes(expectedDummyName)
+				createDummyConnectionAtNodes(expectedDummyName)
 			})
 			AfterEach(func() {
-				deleteConnectionAtNodes(expectedDummyName)
-				By("Make sure the dummy interface gets deleted")
-				for _, node := range nodes {
-					for _, iface := range interfacesNameForNode(node) {
-						if iface == expectedDummyName {
-							deleteDeviceAtNode(node, expectedDummyName)
-						}
-					}
-				}
+				deleteConnectionAndWait(nodes, expectedDummyName)
 			})
 			It("[test_id:3794]should update node network state with it", func() {
 				for _, nodeName := range nodes {
 					Eventually(func() []string {
 						return interfacesNameForNode(nodeName)
-					}, node.NetworkStateRefresh, time.Second).Should(ContainElement(expectedDummyName))
+					}, 2*node.NetworkStateRefresh, time.Second).Should(ContainElement(expectedDummyName))
 				}
 			})
 		})
--- a/test/e2e/handler/utils.go
+++ b/test/e2e/handler/utils.go
@@ -21,6 +21,7 @@ import (
 
 	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
 	nmstatev1beta1 "github.com/nmstate/kubernetes-nmstate/api/v1beta1"
+	nmstatenode "github.com/nmstate/kubernetes-nmstate/pkg/node"
 	"github.com/nmstate/kubernetes-nmstate/test/cmd"
 	"github.com/nmstate/kubernetes-nmstate/test/e2e/handler/linuxbridge"
 	testenv "github.com/nmstate/kubernetes-nmstate/test/env"
@@ -221,24 +222,53 @@ func deleteBridgeAtNodes(bridgeName string, ports ...string) []error {
 	return errs
 }
 
-func createDummyAtNodes(dummyName string) []error {
+func createDummyConnection(nodesToModify []string, dummyName string) []error {
 	By(fmt.Sprintf("Creating dummy %s", dummyName))
-	_, errs := runner.RunAtNodes(nodes, "sudo", "nmcli", "con", "add", "type", "dummy", "con-name", dummyName, "ifname", dummyName, "ip4", "192.169.1.50/24")
-	_, upErrs := runner.RunAtNodes(nodes, "sudo", "nmcli", "con", "up", dummyName)
+	_, errs := runner.RunAtNodes(nodesToModify, "sudo", "nmcli", "con", "add", "type", "dummy", "con-name", dummyName, "ifname", dummyName, "ip4", "192.169.1.50/24")
+	_, upErrs := runner.RunAtNodes(nodesToModify, "sudo", "nmcli", "con", "up", dummyName)
 	errs = append(errs, upErrs...)
 	return errs
 }
 
-func deleteConnectionAtNodes(name string) []error {
+func createDummyConnectionAtNodes(dummyName string) []error {
+	return createDummyConnection(nodes, dummyName)
+}
+
+func createDummyConnectionAtAllNodes(dummyName string) []error {
+	return createDummyConnection(allNodes, dummyName)
+}
+
+func deleteConnection(nodesToModify []string, name string) []error {
 	By(fmt.Sprintf("Delete connection %s", name))
-	_, errs := runner.RunAtNodes(nodes, "sudo", "nmcli", "con", "delete", name)
+	_, errs := runner.RunAtNodes(nodesToModify, "sudo", "nmcli", "con", "delete", name)
 	return errs
 }
 
-func deleteDeviceAtNode(node string, name string) error {
-	By(fmt.Sprintf("Delete device %s  at node %s", name, node))
-	_, err := runner.RunAtNode(node, "sudo", "nmcli", "device", "delete", name)
-	return err
+func deleteConnectionAtNodes(name string) []error {
+	return deleteConnection(nodes, name)
+}
+func deleteConnectionAtAllNodes(name string) []error {
+	return deleteConnection(allNodes, name)
+}
+
+func deleteDevice(nodesToModify []string, name string) []error {
+	By(fmt.Sprintf("Delete device %s  at nodes %v", name, nodesToModify))
+	_, errs := runner.RunAtNodes(nodesToModify, "sudo", "nmcli", "device", "delete", name)
+	return errs
+}
+
+func waitForInterfaceDeletion(nodesToCheck []string, interfaceName string) {
+	for _, nodeName := range nodesToCheck {
+		Eventually(func() []string {
+			return interfacesNameForNode(nodeName)
+		}, 2*nmstatenode.NetworkStateRefresh, time.Second).ShouldNot(ContainElement(interfaceName))
+	}
+}
+
+func deleteConnectionAndWait(nodesToModify []string, interfaceName string) {
+	deleteConnection(nodesToModify, interfaceName)
+	deleteDevice(nodesToModify, interfaceName)
+	waitForInterfaceDeletion(nodesToModify, interfaceName)
 }
 
 func interfaces(state nmstate.State) []interface{} {


RamLavi (changes requested):
Some comments. Also could you rephrase the commit message? I'm not sure I got what you meant there

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,22 +26,28 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
 			originalNNSs[node] = nodeNetworkState(key)
 		}
 	})
-	Context("when network configuration hasn't change", func() {
-		It("should not be updated", func() {
+	Context("when network configuration is not changed by a NNCP", func() {
+		FIt("should be updated only if configuration changed", func() {
+			timeout := 2 * nmstatenode.NetworkStateRefresh
+			By(fmt.Sprintf("Sleeping %s to wait for nns refresh", timeout))
```
phoracek, line 32: Can't you use Consistently instead? So we have an early exit in case the functionality does not work as expected.

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,22 +26,28 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
-				}, timeout, time.Second).Should(MatchAllFields(Fields{
-					"CurrentState":             WithTransform(shared.State.String, Equal(originalNNS.Status.CurrentState.String())),
-					"LastSuccessfulUpdateTime": Equal(originalNNS.Status.LastSuccessfulUpdateTime),
-					"Conditions":               Equal(originalNNS.Status.Conditions),
-				}), func() string {
-					return fmt.Sprintf("currentState diff: \n%s", diff.LineDiff(originalNNS.Status.CurrentState.String(), nodeNetworkState(key).Status.CurrentState.String()))
-				})
+				currentStatus := nodeNetworkState(key).Status
+				originalStatus := originalNNS.Status
+				if currentStatus.CurrentState.String() != originalStatus.CurrentState.String() {
```
phoracek, line 39: Isn't this given? There is nothing applying NNCP, so this should stay intact. And it is not in the scope of the test t check that current status stays the same, is it?

phoracek (changes requested):

rhrazdil (commented):
The code is looking good when Ram's comments are addressed, I'm just wondering if we really need the test.
Isn't it enough to test that timestamp is updated when we really update the state?
Alternatively, what do you think about splitting this into two tests:

  1) test that the timestamp is updated when we change state
  2) test that the timestamp hasn't changed if we didn't change the state, add a retry block that would wait couple of times for 2 * nmstatenode.NetworkStateRefresh and only when the state really hasn't changed, check the timestamp?
This way the tests wouldn't be doing different things, but the issue that the state may be changed by someone else is still there though

EDIT: now I noticed we already have test for checking the timestamp after we change nns

test/e2e/handler/nns_update_timestamp_test.go
qinqon: Good idea, done.

test/e2e/handler/nns_update_timestamp_test.go
qinqon: Looks like network configuration changes without human intervention at kubevirtci, some ipv6 routes disappear sometimes, also I have see at openshift it happends a lot when I was debugging the big cluster, so the tests is prepare for it at the end what we want is to test that updatetimestamp has the expected value.

qinqon:
> The code is looking good when Ram's comments are addressed, I'm just wondering if we really need the test.
> Isn't it enough to test that timestamp is updated when we really update the state?

The difference is that when we apply a NNCP we don't wait for RequeueAfter and we just force refresh nns, while if a change appears at network configuration without NNCP it appears at NNS after the RequeueAfter so it's difference scenario.

> Alternatively, what do you think about splitting this into two tests:
> 
> 1. test that the timestamp is updated when we change state

Good idea I change the network configuration directly at node and wait for state to be changed.

> 2. test that the timestamp hasn't changed if we didn't change the state, add a retry block that would wait couple of times for `2 * nmstatenode.NetworkStateRefresh` and only when the state really hasn't changed, check the timestamp?
> 

Problem with that is we don't know exactly how many retries we have to do, but we can try, I remember we have something similar after a few iterations and the issue was still present.

> This way the tests wouldn't be doing different things based on what the nns happens to be, but the issue that the state may be changed by someone else is still there though

test/e2e/handler/nns_update_timestamp_test.go
phoracek: I'm not sure I understand the test there. It tests two thing and it depends on some random changes on the host which tests gets triggered?

qinqon: I have splitet the test as @rhrazdil suggested, please take a look

test/reporter/reporter.go
```diff
@@ -141,6 +142,7 @@ func (r *KubernetesNMStateReporter) logPods(testName string, sinceTime time.Time
 }
 
 func (r *KubernetesNMStateReporter) OpenTestLogFile(logType string, testName string, cb func(f io.Writer), extraWriters ...io.Writer) {
+	syscall.Umask(0)
```
rhrazdil, line 145: why do we need this? Shouldn't we change it back to the original mask?

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,49 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
-					return fmt.Sprintf("currentState diff: \n%s", diff.LineDiff(originalNNS.Status.CurrentState.String(), nodeNetworkState(key).Status.CurrentState.String()))
-				})
+				By(fmt.Sprintf("Checking timestamp agains original one %s", originalNNS.Status.LastSuccessfulUpdateTime))
+				Eventually(func() time.Time {
+					currentNNS := nodeNetworkState(types.NamespacedName{Name: node})
+					return currentNNS.Status.LastSuccessfulUpdateTime.Time
+				}, 2*nmstatenode.NetworkStateRefresh, 10*time.Second).Should(BeTemporally(">=", originalNNS.Status.LastSuccessfulUpdateTime.Add(nmstatenode.NetworkStateRefresh)), "should update it at %s", node)
 			}
 		})
+
```
rhrazdil, line 69: Extra new line

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,49 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
+				for node, originalNNS := range originalNNSs {
+					key := types.NamespacedName{Name: node}
+					currentStatus := nodeNetworkState(key).Status
+					originalStatus := originalNNS.Status
+					if currentStatus.CurrentState.String() == originalStatus.CurrentState.String() {
+						By(fmt.Sprintf("Check LastSuccessfulUpdateTime changed at %s", node))
+						Expect(currentStatus.LastSuccessfulUpdateTime).To(Equal(originalStatus.LastSuccessfulUpdateTime))
+						retry = 0
+					} else {
+						By("Check LastSuccessfulUpdateTime again network configuration changed")
```
rhrazdil, line 43: Let's hope nothing would change the networking too often, still, we may want to fail here if we run out or attemts don't we?

rhrazdil (changes requested):

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,49 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
 			originalNNSs[node] = nodeNetworkState(key)
 		}
 	})
-	Context("when network configuration hasn't change", func() {
-		It("should not be updated", func() {
+	Context("when network configuration is not changed by a NNCP", func() {
```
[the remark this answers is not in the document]
qinqon: we have to different test one that checks changes done by NNCP and other that check on external changes, the first one checks that NNS refresh is instantanious while the external network configuration takes some time.

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,49 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
+		It("should be updated only if configuration changed", func() {
```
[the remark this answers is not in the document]
qinqon: Done

test/e2e/handler/nns_update_timestamp_test.go
qinqon: Refactored using eventually playing with the fact that if we use an interval of 2 * refresh and a timeout of 4 * interval we have the same behaviour and also fails on retry limit.

test/e2e/handler/nns_update_timestamp_test.go
qinqon: Done

test/reporter/reporter.go
qinqon: Removed, it was an experiment.

test/e2e/handler/utils.go
```diff
@@ -221,24 +221,47 @@ func deleteBridgeAtNodes(bridgeName string, ports ...string) []error {
...
-	_, errs := runner.RunAtNodes(nodes, "sudo", "nmcli", "con", "add", "type", "dummy", "con-name", dummyName, "ifname", dummyName, "ip4", "192.169.1.50/24")
-	_, upErrs := runner.RunAtNodes(nodes, "sudo", "nmcli", "con", "up", dummyName)
+	_, errs := runner.RunAtNodes(nodesToModify, "sudo", "nmcli", "con", "add", "type", "dummy", "con-name", dummyName, "ifname", dummyName, "ip4", "192.169.1.50/24")
+	_, upErrs := runner.RunAtNodes(nodesToModify, "sudo", "nmcli", "con", "up", dummyName)
 	errs = append(errs, upErrs...)
 	return errs
 }
 
-func deleteConnectionAtNodes(name string) []error {
+func createDummyAtNodes(dummyName string) []error {
```
RamLavi, line 232: ditto

test/e2e/handler/utils.go
```diff
@@ -221,24 +221,47 @@ func deleteBridgeAtNodes(bridgeName string, ports ...string) []error {
...
+	return createDummy(nodes, dummyName)
+}
+
+func createDummyAtAllNodes(dummyName string) []error {
```
RamLavi, line 236: ditto

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,48 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
+			Eventually(func() error {
+				for node, originalNNS := range originalNNSs {
+					key := types.NamespacedName{Name: node}
+					currentStatus := nodeNetworkState(key).Status
+					originalStatus := originalNNS.Status
+					if currentStatus.CurrentState.String() == originalStatus.CurrentState.String() {
+						By(fmt.Sprintf("Check LastSuccessfulUpdateTime changed at %s", node))
+						Expect(currentStatus.LastSuccessfulUpdateTime).To(Equal(originalStatus.LastSuccessfulUpdateTime))
+					} else {
+						return fmt.Errorf("Network configuration changed, sending and error to retry")
```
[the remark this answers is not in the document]
RamLavi: also, why do we retry and not simple check that in this case (i.e. currentStatus != originalStatus) that the LastSuccessfulUpdateTime
 are different?

RamLavi (changes requested):

test/e2e/handler/utils.go
```diff
@@ -221,24 +221,47 @@ func deleteBridgeAtNodes(bridgeName string, ports ...string) []error {
 	return errs
 }
 
-func createDummyAtNodes(dummyName string) []error {
+func createDummy(nodesToModify []string, dummyName string) []error {
```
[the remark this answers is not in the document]
qinqon: Renamed to `createDummyConnection`, it feels better.

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,48 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
+				By(fmt.Sprintf("Checking timestamp agains original one %s", originalNNS.Status.LastSuccessfulUpdateTime))
+				Eventually(func() time.Time {
+					currentNNS := nodeNetworkState(types.NamespacedName{Name: node})
+					return currentNNS.Status.LastSuccessfulUpdateTime.Time
+				}, 2*nmstatenode.NetworkStateRefresh, 10*time.Second).Should(BeTemporally(">=", originalNNS.Status.LastSuccessfulUpdateTime.Add(nmstatenode.NetworkStateRefresh)), "should update it at %s", node)
 			}
 		})
+
 	})
 	Context("when network configuration is changed by a NNCP", func() {
```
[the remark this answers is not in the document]
qinqon: You know how controllers works, there is always some latencies.

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,48 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
+						return fmt.Errorf("Network configuration changed, sending and error to retry")
+					}
+				}
+				return nil
+			}, timeout, interval).Should(Succeed())
+		})
+	})
+	Context("when network configuration is changed externally", func() {
+		var (
+			expectedDummyName = "dummy0"
```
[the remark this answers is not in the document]
qinqon: Done

test/e2e/handler/nns_update_timestamp_test.go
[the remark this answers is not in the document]
qinqon: Fore that we have the other test that create the dummy, this one has to verify that if network configuration has not changed the LastSuccessfulUpdateTime is not changed so we retry until network configuration does not change is the best we can do.

test/e2e/handler/nns_update_timestamp_test.go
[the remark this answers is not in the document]
rhrazdil: ack, thanks

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -28,24 +24,47 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
+		BeforeEach(func() {
+			createDummyConnectionAtAllNodes(expectedDummyName)
+		})
+		AfterEach(func() {
+			deleteConnectionAtAllNodes(expectedDummyName)
+			deleteDeviceAtAllNodes(expectedDummyName)
+		})
+		It("should update it with according to network state refresh duration", func() {
+			for node, originalNNS := range originalNNSs {
+				By(fmt.Sprintf("Checking timestamp agains original one %s", originalNNS.Status.LastSuccessfulUpdateTime))
```
rhrazdil, line 60: nit: `against`

test/e2e/handler/nns_update_timestamp_test.go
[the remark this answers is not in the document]
RamLavi: could you change the phrasing to `It("should eventually be updated"`? I'm afraid `immediately` would imply we are expecting a synchronous action. could also be in a later separate PR, as it's not directly related.

test/e2e/handler/nns_update_timestamp_test.go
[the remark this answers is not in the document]
RamLavi: but if the reoccurring "non human intervention" changes start being more frequent, then we might fail even then..

I kinda think that if we can't isolate the scenario more tightly, then maybe we should just drop this test, and find a proper substitute for it in unit tests (I think we already have something like this, no?)... 
what do you think?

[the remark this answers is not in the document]
qinqon: Let's give the test a try merghing it, in case of network configuration not stable even after 4 retries let's just move it to unit test.

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/d1d84317feb34865470c0dbf681963c3a1d2dace..721e94e241fa484fb08df96f7103fecb0684eaad): Fix race at teardown when dummy interface is deleted but we don't wait for NNS to reflect that.

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/721e94e241fa484fb08df96f7103fecb0684eaad..0e9ca68651d87fc7a19027e8f6265f7789118df1): fix FIt

RamLavi (approved):
/lgtm

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -22,30 +18,59 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
-				}), func() string {
-					return fmt.Sprintf("currentState diff: \n%s", diff.LineDiff(originalNNS.Status.CurrentState.String(), nodeNetworkState(key).Status.CurrentState.String()))
-				})
+		BeforeEach(func() {
+			createDummyConnectionAtAllNodes(expectedDummyName)
+		})
+		AfterEach(func() {
+			deleteConnectionAtAllNodes(expectedDummyName)
+			deleteDeviceAtAllNodes(expectedDummyName)
+			for _, nodeName := range allNodes {
```
rhrazdil, line 58: This is used at two places, could you define a function for it please?

test/e2e/handler/nodes_test.go
```diff
@@ -27,28 +27,25 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 			AfterEach(func() {
 				deleteConnectionAtNodes(expectedDummyName)
-				By("Make sure the dummy interface gets deleted")
-				for _, node := range nodes {
-					for _, iface := range interfacesNameForNode(node) {
-						if iface == expectedDummyName {
-							deleteDeviceAtNode(node, expectedDummyName)
-						}
-					}
+				deleteDeviceAtNodes(expectedDummyName)
```
rhrazdil, line 37: ditto

rhrazdil (changes requested):

rhrazdil:
Thanks!
/lgtm

phoracek (commented):
/approve

qinqon:
@RamLavi @rhrazdil @phoracek, there were some flakiness related to NNCP applied from another tests and force update NNS, so there is a race between the tests checking NNS timestamp and force refresh caling reconcile and update the state with the created dummy, I have add 5 seconds sleep before the test suite.

test/e2e/handler/nodes_test.go
```diff
@@ -27,28 +27,19 @@ var _ = Describe("[rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
-					for _, iface := range interfacesNameForNode(node) {
-						if iface == expectedDummyName {
-							deleteDeviceAtNode(node, expectedDummyName)
-						}
-					}
-				}
+				deleteConnectionAndWait(nodes, expectedDummyName)
 			})
-			It("[test_id:3794]should update node network state with it", func() {
+			FIt("[test_id:3794]should update node network state with it", func() {
```
rhrazdil, line 38: FIt -> It

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -22,30 +18,55 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
...
-				}), func() string {
-					return fmt.Sprintf("currentState diff: \n%s", diff.LineDiff(originalNNS.Status.CurrentState.String(), nodeNetworkState(key).Status.CurrentState.String()))
-				})
+		BeforeEach(func() {
+			createDummyConnectionAtAllNodes(expectedDummyName)
+		})
+		AfterEach(func() {
+			deleteConnectionAndWait(allNodes, expectedDummyName)
+		})
+		FIt("should update it with according to network state refresh duration", func() {
```
rhrazdil, line 60: FIt -> It

test/e2e/handler/nns_update_timestamp_test.go
```diff
@@ -22,30 +18,55 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
 		originalNNSs map[string]nmstatev1beta1.NodeNetworkState
 	)
 	BeforeEach(func() {
+		By("Sleep 5 seconds to wait for force refresh applied nncp at other tests")
```
rhrazdil, line 21: Do you think it would make sense to add a function that would wait for nns to not change for 2-3 intervals?
Something similar as they do in k8s here
https://github.com/kubernetes/kubernetes/blob/02b9dd6a6e7d2928945afa4b8aedd7650d6d4cab/test/e2e/apimachinery/resource_quota.go#L301

qinqon: Sure, let's do it

qinqon: I think I am going to do this different, I am going to just compare the NNS update timstamp to be > to the original one, since we are testing the enqueueafter at unit test.

qinqon:
[force-push](https://github.com/nmstate/kubernetes-nmstate/compare/e8f42a4849d7d5e82987785ddae1248706f1c4c1..1855f3638813ca2020806e348c2f8113a3621ab1): Compare to the original NNS timestamp without adding the NetworkStateRefresh

If it changes for other reasons we are still ok with it, the RequeueAfter is alredy unit tested.

rhrazdil (approved):
/lgtm

PR #715, opened 2021-03-15
Change spec.Parallel field with spec.MaxUnavailable

rhrazdil:

Updated used guide and added an example policy
that specifies maxUnavailable.

>
> /kind bug
/kind enhancement

Change parallel field with maxUnavailable, which allows
more granular configuration for concurrent policy
configuration.

Default behaviour remains unchanged as maxUnavailable
is by default set to 1.

MaxUnavailable can be either set with an int value or
percentage in string:

```
spec:
  maxUnavailable: "50%"
```

or

```
spec:
  maxUnavailable: 3
```

**Release note**:
```release-note
spec.parallel field is dropped from the API, spec.maxUnavailable is added as s substitute
```


qinqon (changes requested):
Before reviewing this I think we have to put default to "50%" and remove parallel lane, what do you think ?

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
```diff
@@ -49,17 +49,20 @@ spec:
                 description: The desired configuration of the policy
                 type: object
                 x-kubernetes-preserve-unknown-fields: true
+              maxUnavailable:
+                anyOf:
+                - type: integer
+                - type: string
+                description: MaxUnavailable specifies percentage or number of machines
+                  that can be updating at a time. Default is 1.
+                x-kubernetes-int-or-string: true
```
qinqon, line 58: Is this generated ?

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -154,6 +159,25 @@ func (r *NodeNetworkConfigurationPolicyReconciler) initializeEnactment(policy nm
...
+func (r *NodeNetworkConfigurationPolicyReconciler) maxUnavailableNodesCount(policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) (int, error) {
+	nmstateNodes, err := policyconditions.NodesRunningNmstate(r.Client)
+	if err != nil {
+		return 0, err
+	}
+	intOrPercent := intstr.FromInt(DEFAULT_MAXUNAVAILABLE)
+	if policy.Spec.MaxUnavailable != nil {
+		intOrPercent = *policy.Spec.MaxUnavailable
+	}
+	maxUnavailable, err := intstr.GetValueFromIntOrPercent(&intOrPercent, len(nmstateNodes), true)
```
qinqon, line 171: Looks like we have to use `GetScaledValueFromIntOrPercent` since `GetValueFromIntOrPercent` is being deprecated, 
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go#L141-L168

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -245,28 +277,28 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
...
+		log.Error(err, "")
+		enactmentConditions.NotifyAborted(err)
+		return ctrl.Result{}, nil
+	}
+
+	err = r.claimNodeRunningUpdate(instance)
+	if err != nil {
+		if apierrors.IsConflict(err) {
+			return ctrl.Result{RequeueAfter: nodeRunningUpdateRetryTime}, err
+		} else {
```
qinqon, line 297: we don't need this else since previous is a return

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -245,28 +277,28 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
...
-			if apierrors.IsConflict(err) {
-				return ctrl.Result{RequeueAfter: nodeRunningUpdateRetryTime}, err
-			} else {
-				return ctrl.Result{}, err
-			}
+
+	enactmentCount, err := r.enactmentsCountsByPolicy(instance)
+	if err != nil {
+		log.Error(err, "Error getting enactment counts")
+		return ctrl.Result{}, nil
```
qinqon, line 284: We don't return the error ? It will call Reconcile again.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -171,25 +195,33 @@ func (r *NodeNetworkConfigurationPolicyReconciler) claimNodeRunningUpdate(policy
...
 		return err
 	}
 	return nil
 }
 
-func (r *NodeNetworkConfigurationPolicyReconciler) releaseNodeRunningUpdate(policyKey types.NamespacedName) {
+func (r *NodeNetworkConfigurationPolicyReconciler) releaseNodeRunningUpdate(policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) {
+	policyKey := types.NamespacedName{Name: policy.GetName(), Namespace: policy.GetNamespace()}
 	instance := &nmstatev1beta1.NodeNetworkConfigurationPolicy{}
 	_ = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
qinqon, line 216: Let at least print the error

qinqon (changes requested):
First pass, skipped tests.

deploy/crds/nmstate.io_nodenetworkconfigurationpolicies.yaml
rhrazdil: yes it is

api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -12,18 +14,20 @@ type NodeNetworkConfigurationPolicySpec struct {
 	// The desired configuration of the policy
 	DesiredState State `json:"desiredState,omitempty"`
 
-	// When set to true, changes are applied to all nodes in parallel
+	// MaxUnavailable specifies percentage or number
+	// of machines that can be updating at a time. Default is 1.
```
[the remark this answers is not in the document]
rhrazdil: I can set default to 50%, but in CI we only have 3 nodes, so we'll need to force parallel lane to 75% to have 2 nodes working in parallel

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: AFAIK, returning error triggers reconcile again
https://cluster-api.sigs.k8s.io/developer/providers/implementers-guide/controllers_and_reconciliation.html#reconciliation

api/shared/nodenetworkconfigurationpolicy_types.go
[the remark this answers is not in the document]
rhrazdil: Done:
- normal lane: NMSTATE_MAXUNAVAILABLE=1
- parallel lane: NMSTATE_MAXUNAVAILABLE=2 (for the reason above, 50% would only be 1 node)

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -53,6 +54,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	DEFAULT_MAXUNAVAILABLE = 1
```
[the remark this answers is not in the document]
rhrazdil: done

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: removed

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Right! I didn't see that because I had old vendor packages. Done

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Done

controllers/nodenetworkconfigurationpolicy_controller.go
qinqon: But we really don't want Reconcile to be called again ?

api/shared/nodenetworkconfigurationpolicy_types.go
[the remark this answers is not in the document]
qinqon: Ack, let's do a follow up PR with KUBEVIRT_NUM_NODES=5 (1 master + 4 workers) since most of the test NNCP are applied towards workers and keep just one lane.

Since we will have just one lane even with 5 nodes we will consume less resources (2 lanes * 3 nodes = 6 nodes)

docs/user-guide/102-configuration.md
```diff
@@ -462,6 +462,38 @@ status:
...
+
+By default, Policy configuration is applied sequentially, one node at a time.
+This configuration strategy is safe and prevents the entire cluster from being
+temporarily unavailable, if the applied configuration breaks network connectivity.
+
+For big clusters however, it may take too much time for a configuration to finish.
+In such a case, `maxUnavailable` can be used to define portion size of a cluster
+that can apply a policy configuration concurrently.
+MaxUnavailable specifies percentage or a constant number of nodes that
+can be progressing a policy at a time. Default is 1.
```
qinqon, line 475:
```suggestion
can be progressing a policy at a time. Default is "50%".
```

pkg/policyconditions/conditions.go
```diff
@@ -86,7 +86,7 @@ func SetPolicyFailedToConfigure(conditions *nmstate.ConditionList, message strin
 	)
 }
 
-func nodesRunningNmstate(cli client.Client) ([]corev1.Node, error) {
+func NodesRunningNmstate(cli client.Client) ([]corev1.Node, error) {
```
qinqon, line 89: Can we do some boy scout here and mov this to `pkg/node` with something like `node.FindNmstateRunning` since  this has nothing to do with policyconditions

qinqon (commented):
Second pass, just nits, e2e tests looks fine.

docs/user-guide/102-configuration.md
rhrazdil: Thanks, misssed that. Fixed

pkg/policyconditions/conditions.go
rhrazdil: Done

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Hmm, indeed we should reconcile.

rhrazdil: Fixed

qinqon (changes requested):
Small nit at function name.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -245,28 +281,27 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
...
+		return ctrl.Result{}, err
+	}
+	if enactmentCount.Failed() > 0 {
+		err = fmt.Errorf("policy has failing enactments, aborting")
+		log.Error(err, "")
+		enactmentConditions.NotifyAborted(err)
+		return ctrl.Result{}, nil
+	}
+
+	err = r.claimNodeRunningUpdate(instance)
```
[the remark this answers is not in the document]
rhrazdil: [force-pushed](https://github.com/nmstate/kubernetes-nmstate/compare/79736c6282c6fd9db84673d17c64865dfe9f8343..955c66c1ca8ae8bdf79796be5ffd9b37fcd8128b) changes to the said function name, also updated related error messages to be aligned with the code

qinqon (approved):
Let's do the CI cleanup at follow up

PR #716, opened 2021-03-17
Move cert-manager execution to a new pod

qinqon:

>
> /kind bug
> /kind enhancement

Currectly the cert-manager from kube-admission-webhook is running
whithing the webhook server itself, this means that we need to create a
placeholder secret and also a very specific readiness probe has to be
used to ensure that secret is properly rotated at start up. This change
introduces a new pod runing the cert-manager code so we don't need to
create the placeholder secret (the webhook pod will not start until the
secret is not created) and we don't need a readiness probess since
secret will be correct form the beginning.

Also start up is faster since webhook pods start at the very moment the
secret is created by cert-manager pod.

**Release note**:

```release-note
Move cert-manager code to a cert-manager pod 
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -56,6 +56,75 @@ spec:
               value: "False"
             - name: PROFILER_PORT
               value: "6060"
+          ports:
+          - containerPort: 9443
+            name: webhook-server
+            protocol: TCP
+          volumeMounts:
+          - name: tls-key-pair
+            readOnly: true
+            mountPath: /tmp/k8s-webhook-server/serving-certs/
+      volumes:
+        - name: tls-key-pair
+          secret:
+            secretName: {{template "handlerPrefix" .}}nmstate-webhook
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{template "handlerPrefix" .}}nmstate-cert-manager
+  namespace: {{ .HandlerNamespace }}
+  labels:
+    app: kubernetes-nmstate
+    component: kubernetes-nmstate-cert-manager
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      name: {{template "handlerPrefix" .}}nmstate-cert-manager
+  template:
+    metadata:
+      labels:
+        app: kubernetes-nmstate
+        component: kubernetes-nmstate-cert-manager
+        name: {{template "handlerPrefix" .}}nmstate-cert-manager
+      annotations:
+        description: kubernetes-nmstate-webhook rotate webhook certs
+    spec:
+      serviceAccountName: {{template "handlerPrefix" .}}nmstate-handler
+      nodeSelector: {{ toYaml .WebhookNodeSelector | nindent 8 }}
+      tolerations: {{ toYaml .WebhookTolerations | nindent 8 }}
+      affinity: {{ toYaml .WebhookAffinity | nindent 8 }}
+      containers:
+        - name: nmstate-cert-manager
+          args:
+          - --v=production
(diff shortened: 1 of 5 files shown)


deploy/handler/operator.yaml
```diff
@@ -56,38 +56,85 @@ spec:
...
-              value: {{ .CARotateInterval | default "8760h0m0s" }}
-            - name: CA_OVERLAP_INTERVAL
-              value: {{ .CAOverlapInterval | default "24h0m0s" }}
-            - name: CERT_ROTATE_INTERVAL
-              value: {{ .CertRotateInterval | default "4380h0m0s" }}
-            - name: CERT_OVERLAP_INTERVAL
-              value: {{ .CertOverlapInterval | default "24h0m0s" }}
           ports:
-          - containerPort: 8443
+          - containerPort: 9443
```
rhrazdil, line 60: Why are we changing this?

rhrazdil: Ack, Thanks

qinqon:
/hold

Let's do a test PR at CNAO to verify that it's happy about this.

deploy/handler/operator.yaml
RamLavi, line 60: will it have future value to put this in a env var (prevent future collisions)?

deploy/handler/operator.yaml
```diff
@@ -56,38 +56,85 @@ spec:
...
-              httpHeaders:
-              - name: Content-Type
-                value: application/json
-            initialDelaySeconds: 10
-            periodSeconds: 10
           volumeMounts:
           - name: tls-key-pair
             readOnly: true
-            mountPath: /etc/webhook/certs
+            mountPath: /tmp/k8s-webhook-server/serving-certs/
```
RamLavi, line 66: genereally speaking, why in `/tmp/...`?

RamLavi (commented):
first brush, 
Genereally speaking, don't we want to move it to a pod from kaw itself?

deploy/handler/operator.yaml
qinqon: we don't run it with network=host so is not going to collision, I doubt we are going to have multiple containers at the same pod with the same port.

deploy/handler/operator.yaml
qinqon: Is the default value from controller-runtime webhook server

qinqon:
> first brush,
> Genereally speaking, don't we want to move it to a pod from kaw itself?

kaw is just a little cert-manager specialized on webhooks, where or how do you deploy is user's responsability.

qinqon:
Testing it a CNAO here https://github.com/kubevirt/cluster-network-addons-operator/pull/806

qinqon:
/hold cancel

Test PR at CNAO is working fine https://github.com/kubevirt/cluster-network-addons-operator/pull/806

deploy/handler/operator.yaml
```diff
@@ -56,38 +56,85 @@ spec:
...
 apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{template "handlerPrefix" .}}nmstate-cert-manager
+  namespace: {{ .HandlerNamespace }}
+  labels:
+    app: kubernetes-nmstate
+    component: kubernetes-nmstate-cert-manager
+spec:
+  replicas: 2
```
phoracek, line 81: nmstate already has quite a lot of moving parts. We don't mind downtime of couple of minutes in case one of the nodes running this goes down, do we? If not, could we keep it on a single replica? To save memory and CPU.

phoracek (changes requested):

phoracek (approved):
/lgtm
/approve

PR #717, opened 2021-03-23
Consolidate parallel and sequential lanes

rhrazdil:
Run the e2e handler test lane only once, configured
to use default 50% of worker nodes to process
NNCPs concurrently.

For that reason, setting number of nodes to 5
so that 2 nodes run concurrently (5/2 == 2.5 => 2)

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

**Is this a BUG FIX or a FEATURE ?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
/kind enhancement

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**Release note**:

```release-note
run e2e-handler tests in parallel with 50% of worker nodes
```

PR #719, opened 2021-03-25
Fix maxUnavailable nodes calculation from percentages to only consider matching nodes

rhrazdil:
Change how maxUnavailable node cound is computed from percentages
In the original commit that introduced maxUnavailable field, the number of nodes
that can run in parallel is computed from the entire cluster node count, when percentage
value is specified. This commit changes that to use only label matching nodes.

So if a nodeSelector is specified that selects half of cluster nodes, then specifying 50% maxUnavailable
will only allow half of the nodes that have the matching label to run in parallel.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
maxUnavailable node count is calculated from mathing nodes only when percentage is specified
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -40,11 +40,12 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	nmstateapi "github.com/nmstate/kubernetes-nmstate/api/shared"
 	nmstatev1beta1 "github.com/nmstate/kubernetes-nmstate/api/v1beta1"
+	"github.com/nmstate/kubernetes-nmstate/pkg/enactment"
 	"github.com/nmstate/kubernetes-nmstate/pkg/enactmentstatus"
 	enactmentconditions "github.com/nmstate/kubernetes-nmstate/pkg/enactmentstatus/conditions"
 	"github.com/nmstate/kubernetes-nmstate/pkg/environment"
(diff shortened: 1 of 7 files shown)


rhrazdil:
/hold
waiting for #717 to land

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -160,16 +159,12 @@ func (r *NodeNetworkConfigurationPolicyReconciler) initializeEnactment(policy nm
 	})
 }
 
-func (r *NodeNetworkConfigurationPolicyReconciler) maxUnavailableNodeCount(policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) (int, error) {
-	nmstateNodes, err := node.NodesRunningNmstate(r.Client)
-	if err != nil {
-		return 0, err
-	}
+func (r *NodeNetworkConfigurationPolicyReconciler) maxUnavailableNodeCount(policy *nmstatev1beta1.NodeNetworkConfigurationPolicy, matchingNodes int) (int, error) {
```
qinqon, line 162: Remove the `matchingNodes` argument and call `enactmentCount.Matching()` inside at the end this is what we want to do, we don't need to parameterize it also remove the `(r *NodeNetworkConfigurationPolicyReconciler)` receiver since is not used, we can even move this undier ./pkg and make it public to use it at integratin tests.

test/e2e/handler/utils.go
```diff
@@ -37,7 +37,7 @@ const TestPolicy = "test-policy"
 var (
 	bridgeCounter  = 0
 	bondConunter   = 0
-	maxUnavailable = environment.GetIntVarWithDefault("NMSTATE_MAX_UNAVAILABLE", 1)
+	maxUnavailable = environment.GetVarWithDefault("NMSTATE_MAX_UNAVAILABLE", "50%")
```
qinqon, line 40: Let's make the default value public somewhere at `pkg` so we use it at tests too.

test/e2e/handler/utils.go
```diff
@@ -539,3 +539,15 @@ func skipIfNotKubernetes() {
 		Skip("Tutorials use interface naming that is available only on Kubernetes providers")
 	}
 }
+
+func maxUnavailableNodeCount() int {
```
qinqon, line 543: As said let's try to use the same logic from porduct moving the function under `pkg` and usint it on tests.

qinqon (changes requested):
Looks good, let's try to remove duplication and encapsulate things propertly.

rhrazdil:
@qinqon sorry forgot to replace the 50% string in utils, [fixed](https://github.com/nmstate/kubernetes-nmstate/compare/28be74c2d76d9c75c51c3e6b265d1d835c50f1f4..b2f374b0acc1b6da7d7daf672cb6ee0d02fc4de9)

qinqon:
/hold cancel

This no longer depends on the lanes aggregation.

PR #721, opened 2021-03-26
Put back the readiness probe

qinqon:

>
> /kind bug
> /kind enhancement

Although the mounted secret is good from the beginning, webhook will be
only ready when webhook server is started up.

**Release note**:

```release-note
Add readiness probe back
```


rhrazdil (approved):
Makes sense, only report webhook ready when the server s running.
/approve
/lgtm

PR #724, opened 2021-04-09
Add readiness probe to handler

qinqon:

>
> /kind bug
/kind enhancement

If kubernetes-nmstate is deployed at a cluster with not compatible nodes
(like NetworkManager not started, bad version or not installed) the
kubernetes-nmstate-handler start without issue. This change add a
readiness probe that runs "nmstatectl show" to check if nmstatectl is
functional at a basic level.

**Release note**:

```release-note
Added readiness probe running nmstatectl show to handler
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -216,6 +216,14 @@ spec:
               mountPath: /var/k8s_nmstate
           securityContext:
             privileged: true
+          readinessProbe:
+            exec:
+              command:
+              - nmstatectl
+              - show
+            initialDelaySeconds: 5
+            periodSeconds: 5
+            timeoutSeconds: 1
       volumes:
         - name: dbus-socket
           hostPath:


deploy/handler/operator.yaml
```diff
@@ -216,6 +216,13 @@ spec:
               mountPath: /var/k8s_nmstate
           securityContext:
             privileged: true
+          readinessProbe:
+            exec:
+              command:
+              - nmstatectl
+              - show
+            initialDelaySeconds: 5
+            periodSeconds: 5
```
phoracek, line 225: This means we will stop retrying this once it succeeds, since it is readiness, right?

deploy/handler/operator.yaml
```diff
@@ -216,6 +216,13 @@ spec:
               mountPath: /var/k8s_nmstate
           securityContext:
             privileged: true
+          readinessProbe:
```
rhrazdil, line 219: IIUC what the [docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) say, we should specify `timeoutSeconds` explicitly, because in pre 1.20 version, the default 1 second timeout doesn't work and the probe processes run indefinitely until a command returns.

Maybe not a big risk in case of nmstatectl, but better be safe. wdyt @qinqon ?

deploy/handler/operator.yaml
qinqon: Yep.

deploy/handler/operator.yaml
qinqon: Good catch!!

rhrazdil:
Thank you
/approve
/lgtm

Issue #729, opened 2021-04-20
Handler lockfile mechanism does not work


bcrochet:
**What happened**:
Installed handler via CNAO, then installed via operator as well. Both sets of handlers appears to think they had the lock.

**What you expected to happen**:
Only the first handlers should have controlled the lock.

**How to reproduce it (as minimally and precisely as possible)**:
Install handler via CNAO, then install via operator. 

**Anything else we need to know?**:
NAMESPACE                NAME                                              READY   STATUS    RESTARTS   AGE
cluster-network-addons   cluster-network-addons-operator-966677658-hshtz   1/1     Running   0          17m
cluster-network-addons   nmstate-cert-manager-d6fb64b4f-gj2qx              1/1     Running   2          17m
cluster-network-addons   nmstate-handler-nlv6m                             1/1     Running   0          17m
cluster-network-addons   nmstate-handler-tcn7g                             1/1     Running   0          17m
cluster-network-addons   nmstate-webhook-5b779dd5c9-9mmtg                  1/1     Running   2          17m
cluster-network-addons   nmstate-webhook-5b779dd5c9-qljsb                  1/1     Running   0          17m
...
nmstate                  nmstate-cert-manager-bff5b8695-jxwrq              1/1     Running   1          9m10s
nmstate                  nmstate-handler-9qxs8                             1/1     Running   0          9m10s
nmstate                  nmstate-handler-srfgm                             1/1     Running   0          9m10s
nmstate                  nmstate-operator-8646844d48-nddm4                 1/1     Running   0          12m
nmstate                  nmstate-webhook-868d486d94-pwx4t                  1/1     Running   1          9m10s
nmstate                  nmstate-webhook-868d486d94-w8j25                  1/1     Running   0          9m10s

[bcrochet@bcrochet-test01 kubernetes-nmstate]$ oc logs -n cluster-network-addons nmstate-handler-nlv6m | head -2
{"level":"info","ts":1618951199.669442,"logger":"setup","msg":"Try to take exclusive lock on file: /var/k8s_nmstate/handler_lock"}
{"level":"info","ts":1618951199.6698868,"logger":"setup","msg":"Successfully took nmstate exclusive lock"}

[bcrochet@bcrochet-test01 kubernetes-nmstate]$ oc logs -n nmstate nmstate-handler-9qxs8 | head -2
[...]

PR #731, opened 2021-04-23
Use github.com/gofrs/flock to lock handler

qinqon:

>
/kind bug
Closes https://github.com/nmstate/kubernetes-nmstate/issues/729
> /kind enhancement

The nmstate-handler has a lock that make it wait if another handler is
already running in the system, but the library used for that does not
work well with containers. This change replace that library with
different one that uses golang syscall.Flock behind the scine [1] and
that works fine with containers sharing node volume.

[1] github.com/nightlyone/lockfile.

**Release note**:

```release-note
Fix handler locking using github.com/gofrs/flock.
```


qinqon:
/hold
wip

main.go
```diff
@@ -186,6 +188,25 @@ func main() {
...
+			os.Exit(1)
+			setupLog.Error(err, "failed checking nmstatectl health")
+		}
+
+		// Handler runs with host networking so opening ports is problematic
+		// they will collide with node ports so to ensure that we reach this
+		// point (we have the handler lock and nmstatectl show is working) a
+		// file is touched and the file is checked at readinessProbe field.
+		healthyFile := "/tmp/healthy"
+		setupLog.Info("Marking handler as healthy touching healthy file", "healthyFile", healthyFile)
```
rhrazdil, line 204: This message sounds odd, we don't mark handler healthy here, do we? That's what the probe will do once it tries to cat that file

test/e2e/operator/nmstate_install_test.go
```diff
@@ -85,24 +74,55 @@ var _ = Describe("NMState operator", func() {
...
+		)
+		BeforeEach(func() {
+			installNMState(defaultNMState)
+			daemonset.GetEventually(handlerKey).Should(daemonset.BeReady())
+			installOperator(operatorNamespace)
+		})
+		AfterEach(func() {
+			uninstallNMState(defaultNMState)
+			uninstallOperator(operatorNamespace)
+			installOperator("nmstate")
```
rhrazdil, line 89: Why is the operator installed in "nmstate" namespace at the end?
If it's done to test that operator can be installed and uninstalled at one namespace and then installed in another, wouldn't it be better to have another spec for it?

main.go
qinqon: well we mark the container by touching the file

rhrazdil: Ahh, I understand now

test/e2e/operator/nmstate_install_test.go
qinqon: The test-e2e-operator test suite spec the operator to be installed, but the test we are adding here remove the nmstate operator to check that the handler installed from different namespace is blocked until the "nmstate" one is delete. 

The install here ensure that after that test the nmstate operator is there as expected at the other test cases.

rhrazdil:
Thank you
/approve
/lgtm

PR #732, opened 2021-04-27
Refactor nncp controller

rhrazdil:

>
> /kind bug
/kind enhancement

The first thing that reader is presumably interested in when opening
nodenetworkconfigurationpolicy_controller.go file is the Reconcile loop.
In this PR, I'm moving the Reconcile loop to the top of the file,
and reordering the helper functions so that the reader can see the
details in the same order in which these helper functions are called in Reconcile.

Additionally, an unused method `enactmentsCountByPolicy` is removed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -89,6 +89,13 @@ var (
 	}
 )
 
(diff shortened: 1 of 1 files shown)


qinqon:
/lgtm
/approve
Thanks!

PR #736, opened 2021-04-28
Bump kube-admission-webhook to 0.15.0

qinqon:

>
> /kind bug
> /kind enhancement

Bump kube-admission-webhook to [v0.15.0](https://github.com/qinqon/kube-admission-webhook/releases/tag/v0.15.0)

**Release note**:

```release-note
Fix certificate rotation issue related to tls: pub key not matching private key
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.8.0
 	github.com/kelseyhightower/envconfig v1.4.0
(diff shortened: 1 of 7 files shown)


rhrazdil (commented):
Would it make sense to split to 2 commits? One to remove lockfile and one to bump kaw?

RamLavi (approved):
/lgtm

qinqon:
> Would it make sense to split to 2 commits? One to remove lockfile and one to bump kaw?

Is the result of calling `make vendor` that's why they are in the same commit, I don't htink it's worth it to split it.

PR #740, opened 2021-05-03
cluster-sync, Extend eventually timeout

oshoval:
/kind bug

The first deployment after make cluster-up
might take few more seconds.
Extend the timeout a bit, so the make cluster-sync
would be more stable.

```release-note
None
```

### DIFF
--- a/cluster/sync-common.sh
+++ b/cluster/sync-common.sh
@@ -1,5 +1,5 @@
 function eventually {
-    timeout=$(( $KUBEVIRT_NUM_NODES * 10 ))
+    timeout=$(( $KUBEVIRT_NUM_NODES * 30 ))
     interval=5
     cmd=$@
     echo "Checking eventually $cmd"


qinqon:
/lgtm
/approve

Thanks!

oshoval:
Thanks @qinqon 
I need please ok to test

cluster/sync-common.sh
```diff
@@ -1,5 +1,5 @@
 function eventually {
-    timeout=$(( $KUBEVIRT_NUM_NODES * 10 ))
+    timeout=$(( $KUBEVIRT_NUM_NODES * 30 ))
```
oshoval, line 2: btw there is a little bug i think on line 9 `if [ $timeout -le 0 ]; then`
lets say the timeout is 5, and the interval is 5, it means it will try one time, sleep and then exit
it won't try the 2nd time,
should i change it to `-lt` instead `-le` ?

qinqon: Right, can you a prepare a PR for that ?

oshoval: sure thx

oshoval: https://github.com/nmstate/kubernetes-nmstate/pull/741

PR #742, opened 2021-05-06
Use non cached client at handler for nodes/nns

qinqon:

>
> /kind bug
> /kind enhancement

The cache is configured to select only the Node and NNS that has the
pod's node as key, but update the NNCP Status need to read all the Nodes
and NNSs. This change pass a client.Reader that bypass cache to do those
operations.

**Release note**:

```release-note
```


qinqon:
/hold

It's using a controller-runtime beta version

PR #744, opened 2021-05-07
Cr cache selectors

qinqon:

>
/kind bug
> /kind enhancement

The controller-runtime cache retrieves all the instances of the
instructed type to reconcile, this does not scale at big cluster and in
the case of handler that it's running as daemonset it's even worse, to
fix that se use the controller-runtime cache selectors to filter the
cache queries only to resources that affects the node.
    
The cache is configured to select only the Node and NNS that has the
pod's node as key, but update the NNCP Status need to read all the Nodes
and NNSs. This change pass a client.New instance that bypass cache to do those
operations.

Depends on https://github.com/nmstate/kubernetes-nmstate/pull/761

**Release note**:

```release-note
Filter handler queries for Node and NodeNetworkState to running node name.
```


qinqon:
/hold

It's using a beta controller-runtime version.

alonSadan (commented):
Thank you Quique. Added some minor questions/comments

main.go
```diff
@@ -98,8 +101,20 @@ func main() {
 	} else if environment.IsCertManager() {
 		ctrlOptions.LeaderElection = true
 		ctrlOptions.LeaderElectionID = "5d2e944b.nmstate.io"
+	} else if environment.IsHandler() {
```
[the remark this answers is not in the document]
qinqon: One sections sets the ctrlOptions and the other do the manager/controller setup.

pkg/policyconditions/conditions_test.go
```diff
@@ -161,7 +161,7 @@ var _ = Describe("Policy Conditions", func() {
 
 			client := fake.NewFakeClientWithScheme(s, objs...)
 			key := types.NamespacedName{Name: updatedPolicy.Name}
-			err := Update(client, key)
+			err := Update(client, client, key)
```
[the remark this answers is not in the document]
alonSadan: oh. got it

alonSadan (approved):
Thank you Quique!
~/lgtm~

Edit: as we discus offline there are some other places ([for example](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L75))  that might need  also to use the APIReader. Also we need to make sure that we are not trying to retrieve other resources (besides nodes and nodenetwrokstates) from the cache, as it could also lead to massive resources usage,(for example when poling from the cache fails and the client creates an api -watcher for this object) 

I will try and add inline comments for suspicious places.

alonSadan:
> Thank you Quique!
> ~/lgtm~
> 
> Edit: as we discus offline there are some other places ([for example](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L75)) that might need also to use the APIReader. Also we need to make sure that we are not trying to retrieve other resources (besides nodes and nodenetwrokstates) from the cache, as it could also lead to massive resources usage,(for example when poling from the cache fails and the client creates an api -watcher for this object)
> 
> I will try and add inline comments for suspicious places.

I don't know how to add comments to files not included ion the PR, so execpt for the inline comments ,  I'll add the comments here: 

1) [probes]( https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L93 ): 
   There we create a costume client to bypass the global cache. The apiReader can be used there instead of creating the custom client.
2) [enactmentstatus]( https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/enactmentstatus/status.go#L29): 
we are caching on the global client cache. When the handler pod will run as a daemonset, on large scale, this could be a large memory consumer. Also, if the cache won't have the resource the an api watcher will be created and I think that with a lot od nodes in the cluster this could lead to a lot of watchers. 
3) helper/client: [1](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/helper/client.go#L47) (IN here the function GetNodeNetworkState is never used we might should delete it), [2](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/helper/client.go#L75) same reasons as (2)
4) [slectors/labels](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/selectors/labels.go#L23): same as (2)
5) also I think the webhook is using the global client cache: [1](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/webhook/nodenetworkconfigurationpolicy/handler.go#L60), 
6) [the nmstate controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nmstate_controller.go#L68)
7) [node controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/node_controller.go#L81)
8) [nodenetwrokstate controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nodenetworkstate_controller.go#L56)

qinqon:
> > Thank you Quique!
> > ~/lgtm~
> > Edit: as we discus offline there are some other places ([for example](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L75)) that might need also to use the APIReader. Also we need to make sure that we are not trying to retrieve other resources (besides nodes and nodenetwrokstates) from the cache, as it could also lead to massive resources usage,(for example when poling from the cache fails and the client creates an api -watcher for this object)
> > I will try and add inline comments for suspicious places.
> 

Thanks Alon! that's exactly what we need to check.

> I don't know how to add comments to files not included ion the PR, so execpt for the inline comments , I'll add the comments here:
> 
> 1. [probes](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L93):
>    There we create a costume client to bypass the global cache. The apiReader can be used there instead of creating the custom client.

Although the cache lifetime is going to match the function's scope, is better to just use the apiReader so we don't start any watcher at all and since this probes are executed at NNCP time is not a frequent hit on api server.

> 2. [enactmentstatus](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/enactmentstatus/status.go#L29):
>    we are caching on the global client cache. When the handler pod will run as a daemonset, on large scale, this could be a large memory consumer. Also, if the cache won't have the resource the an api watcher will be created and I think that with a lot od nodes in the cluster this could lead to a lot of watchers.

I will create a new client here so the cache is deleted after the update, using apiReader is not an option since we cannot update it.

> 3. helper/client: [1](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/helper/client.go#L47) (IN here the function GetNodeNetworkState is never used we might should delete it), [2](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/helper/client.go#L75) same reasons as (2)

I will delete GetNodeNetworkState, about the o ther since NodeNetworkState is restricted at global cache we don't have an issue there.

> 4. [slectors/labels](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/selectors/labels.go#L23): same as (2)

Nodes are restricted to at global cache, we don't have an issue here.

> 5. also I think the webhook is using the global client cache: [1](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/webhook/nodenetworkconfigurationpolicy/handler.go#L60),

This is not a problem we don't have a massive number of NodeNetworkConfigurationPolicy also we have just a pair of webhook pods so they will not hit massively the api server.

> 6. [the nmstate controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nmstate_controller.go#L68)

This has nothing to do with the issue since it's the operator that watches the NMState CR.

> 7. [node controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/node_controller.go#L81)

We are restricting Node so we have no issue.

> 8. [nodenetwrokstate controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nodenetworkstate_controller.go#L56)

We are restricting Node so we have no issue.

> 9. several places on [nodenetworkconfigurationpolicy_controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nodenetworkconfigurationpolicy_controller.go)

Those places use restriscted resources Node and NodeNetworkState.

> 10. [policyconditions/conditions](https://github.com/nmstate/kubernetes-nmstate/blob/5f948c287fdaee55048dbcabebfca050234441b8/pkg/policyconditions/conditions.go#L97)
> 

This is NodeNetworkConfigurationPolicy and is not a massive resource.

> I am not all needs to be addressed. @qinqon for your information

qinqon:
> > > Thank you Quique!
> > > ~/lgtm~
> > > Edit: as we discus offline there are some other places ([for example](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L75)) that might need also to use the APIReader. Also we need to make sure that we are not trying to retrieve other resources (besides nodes and nodenetwrokstates) from the cache, as it could also lead to massive resources usage,(for example when poling from the cache fails and the client creates an api -watcher for this object)
> > > I will try and add inline comments for suspicious places.
> 
> Thanks Alon! that's exactly what we need to check.
> 
> > I don't know how to add comments to files not included ion the PR, so execpt for the inline comments , I'll add the comments here:
> > 
> > 1. [probes](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/probe/probes.go#L93):
> >    There we create a costume client to bypass the global cache. The apiReader can be used there instead of creating the custom client.
> 
> Although the cache lifetime is going to match the function's scope, is better to just use the apiReader so we don't start any watcher at all and since this probes are executed at NNCP time is not a frequent hit on api server.
> 

I was wrong `client.New(config, clientOptions) ` creates a client without cache so we are good here.

> > 1. [enactmentstatus](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/enactmentstatus/status.go#L29):
> >    we are caching on the global client cache. When the handler pod will run as a daemonset, on large scale, this could be a large memory consumer. Also, if the cache won't have the resource the an api watcher will be created and I think that with a lot od nodes in the cluster this could lead to a lot of watchers.
> 
> I will create a new client here so the cache is deleted after the update, using apiReader is not an option since we cannot update it.
> 
> > 1. helper/client: [1](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/helper/client.go#L47) (IN here the function GetNodeNetworkState is never used we might should delete it), [2](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/helper/client.go#L75) same reasons as (2)
> 
> I will delete GetNodeNetworkState, about the o ther since NodeNetworkState is restricted at global cache we don't have an issue there.
> 
> > 1. [slectors/labels](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/selectors/labels.go#L23): same as (2)
> 
> Nodes are restricted to at global cache, we don't have an issue here.
> 
> > 1. also I think the webhook is using the global client cache: [1](https://github.com/nmstate/kubernetes-nmstate/blob/main/pkg/webhook/nodenetworkconfigurationpolicy/handler.go#L60),
> 
> This is not a problem we don't have a massive number of NodeNetworkConfigurationPolicy also we have just a pair of webhook pods so they will not hit massively the api server.
> 
> > 1. [the nmstate controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nmstate_controller.go#L68)
> 
> This has nothing to do with the issue since it's the operator that watches the NMState CR.
> 
> > 1. [node controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/node_controller.go#L81)
> 
> We are restricting Node so we have no issue.
> 
> > 1. [nodenetwrokstate controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nodenetworkstate_controller.go#L56)
> 
> We are restricting Node so we have no issue.
> 
> > 1. several places on [nodenetworkconfigurationpolicy_controller](https://github.com/nmstate/kubernetes-nmstate/blob/main/controllers/nodenetworkconfigurationpolicy_controller.go)
> 
> Those places use restriscted resources Node and NodeNetworkState.
> 
> > 1. [policyconditions/conditions](https://github.com/nmstate/kubernetes-nmstate/blob/5f948c287fdaee55048dbcabebfca050234441b8/pkg/policyconditions/conditions.go#L97)
> 
> This is NodeNetworkConfigurationPolicy and is not a massive resource.
> 
> > I am not all needs to be addressed. @qinqon for your information

alonSadan:
I think the second commit now needs to be modified or dropped, because of the new client that was introduced.

main.go
```diff
@@ -98,8 +102,20 @@ func main() {
...
+		// Handler runs as a daemonset and we want that each handler pod will cache/reconcile only resources belongs the node it runs on.
```
alonSadan, line 106: is this still needed now that the client asks the api server for the resources?

pkg/enactment/enactment.go
```diff
@@ -10,7 +10,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func CountByPolicy(cli client.Client, policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) (int, enactmentconditions.ConditionCount, error) {
+func CountByPolicy(cli client.Reader, policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) (int, enactmentconditions.ConditionCount, error) {
```
alonSadan, line 13: needs to be updated to client.APIclient

pkg/node/nodes.go
```diff
@@ -16,7 +16,7 @@ const (
 	DEFAULT_MAXUNAVAILABLE = "50%"
 )
 
-func NodesRunningNmstate(cli client.Client, nodeSelector map[string]string) ([]corev1.Node, error) {
+func NodesRunningNmstate(cli client.Reader, nodeSelector map[string]string) ([]corev1.Node, error) {
```
alonSadan, line 19: ditto

pkg/node/nodes.go
```diff
@@ -42,7 +42,7 @@ func NodesRunningNmstate(cli client.Client, nodeSelector map[string]string) ([]c
 	return filteredNodes, nil
 }
 
-func MaxUnavailableNodeCount(cli client.Client, policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) (int, error) {
+func MaxUnavailableNodeCount(cli client.Reader, policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) (int, error) {
```
alonSadan, line 45: ditto

pkg/policyconditions/conditions.go
```diff
@@ -87,7 +87,7 @@ func SetPolicyFailedToConfigure(conditions *nmstate.ConditionList, message strin
 	)
 }
 
-func Update(cli client.Client, policyKey types.NamespacedName) error {
+func Update(cli client.Client, apiReader client.Reader, policyKey types.NamespacedName) error {
```
alonSadan, line 90: ditto

alonSadan (changes requested):

pkg/enactment/enactment.go
qinqon: We expect this function to only read so we pass a reader.

pkg/node/nodes.go
qinqon: https://github.com/nmstate/kubernetes-nmstate/pull/744#discussion_r638639149

pkg/node/nodes.go
qinqon: https://github.com/nmstate/kubernetes-nmstate/pull/744#discussion_r638639149

main.go
qinqon: What do you mean ?

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -93,8 +93,9 @@ var (
 // NodeNetworkConfigurationPolicyReconciler reconciles a NodeNetworkConfigurationPolicy object
 type NodeNetworkConfigurationPolicyReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	APIClient client.Client
```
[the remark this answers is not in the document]
qinqon: Done

main.go
alonSadan: My bad I confused the controller client which has cache and the other client

pkg/enactment/enactment.go
alonSadan: OK. understand. Thanks

controllers/nodenetworkconfigurationpolicy_controller.go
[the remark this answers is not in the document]
alonSadan: Awesome. Thanks

alonSadan:
Thank you Quique

\lgtm

qinqon:
/hold cancel
This use final controller-runtime version.

rhrazdil (commented):
@qinqon Is there a way to measure the impact of this PR? Could it be tested to prevent us from using a client where as a consequence such use would increase resource consumption?

controllers/nodenetworkconfigurationpolicy_controller.go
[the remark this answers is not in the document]
qinqon: Done

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -261,11 +262,11 @@ func (r *NodeNetworkConfigurationPolicyReconciler) initializeEnactment(policy nm
...
-		enactmentConditions := enactmentconditions.New(r.Client, enactmentKey)
+		enactmentConditions := enactmentconditions.New(r.APIClient, enactmentKey)
 		enactmentConditions.Reset()
 	}
 
-	return enactmentstatus.Update(r.Client, enactmentKey, func(status *nmstateapi.NodeNetworkConfigurationEnactmentStatus) {
+	return enactmentstatus.Update(r.APIClient, enactmentKey, func(status *nmstateapi.NodeNetworkConfigurationEnactmentStatus) {
 		status.DesiredState = policy.Spec.DesiredState
 		status.PolicyGeneration = policy.Generation
 	})
```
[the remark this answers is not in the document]
qinqon: 278/300 they should since we don't want to cache all the enactments at all the handlers, I am doing the change.
351 not needed since we are not going to have a mssive number of NNCPs on the cluster
361 NNSs are restricted by the cache selector so we have no problem there.

qinqon:
> @qinqon Is there a way to measure the impact of this PR? Could it be tested to prevent us from using a client where as a consequence such use would increase resource consumption?

The main problem were that controller-runtime was installing a watcher for the type at every handler so all the handler were watching all the Nodes and nodes are usually frequenly updated with labels and the like, in the case of kubevirt they are updated every minute.

Maybe we can check what watchers we are installing at apiservers, I know for sure that installing the metrics-server allow you to get the number of watchers per resource but it does not give you if the watcher as a field or label selector.

rhrazdil:
@qinqon Thanks for clearing it up.

Unit tests need to be fixed now, because the APIClient is nil. I suppose for the sake of unit tests, we can just used the same client for both
```
reconciler.APIClient = cl
```
at https://github.com/nmstate/kubernetes-nmstate/blob/29432b94743320a48227809f568aadc53b2edcf0/controllers/nodenetworkconfigurationpolicy_controller_test.go#L124

qinqon:
> @qinqon Thanks for clearing it up.
> 
> Unit tests need to be fixed now, because the APIClient is nil. I suppose for the sake of unit tests, we can just used the same client for both
> 
> ```
> reconciler.APIClient = cl
> ```
> 
> at
> 
> https://github.com/nmstate/kubernetes-nmstate/blob/29432b94743320a48227809f568aadc53b2edcf0/controllers/nodenetworkconfigurationpolicy_controller_test.go#L124

Done

rhrazdil:
@qinqon Sorry, totally missclicked

/retest
/lgtm
/approve

PR #748, opened 2021-05-11
Add "Pending" condition to nnce status

qinqon:

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
Add "Pending" condition to NNCE status and change STATUS column to show condition.Type.
```

### DIFF
--- a/api/v1alpha1/nodenetworkconfigurationenactment_types.go
+++ b/api/v1alpha1/nodenetworkconfigurationenactment_types.go
@@ -20,7 +20,7 @@ type NodeNetworkConfigurationEnactmentList struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=nodenetworkconfigurationenactments,shortName=nnce,scope=Cluster
-// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.type==\"Available\")].reason",description="Status"
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.status==\"True\")].type",description="Status"
 
 // NodeNetworkConfigurationEnactment is the Schema for the nodenetworkconfigurationenactments API
 type NodeNetworkConfigurationEnactment struct {
--- a/docs/user-guide/102-configuration.md
+++ b/docs/user-guide/102-configuration.md
@@ -60,7 +60,7 @@ kubectl get nodenetworkconfigurationpolicies
 
 ```
 NAME              STATUS
-bond0-eth1-eth2   SuccessfullyConfigured
+bond0-eth1-eth2   Available
 ```
 
 We can also use short name `nncp` to reach the same effect:
@@ -71,7 +71,7 @@ kubectl get nncp
 
 ```
 NAME              STATUS
-bond0-eth1-eth2   SuccessfullyConfigured
+bond0-eth1-eth2   Available
 ```
 
 By using `-o yaml` we obtain the full Policy with its current state:
@@ -111,8 +111,8 @@ kubectl get nodenetworkconfigurationenactments
 
 ```
 NAME                     STATUS
-node01.bond0-eth1-eth2   SuccessfullyConfigured
-node02.bond0-eth1-eth2   SuccessfullyConfigured
+node01.bond0-eth1-eth2   Available
+node02.bond0-eth1-eth2   Available
 ```
 
 We can also use short name `nnce` to reach the same effect:
@@ -123,8 +123,8 @@ kubectl get nnce
 
 ```
 NAME                     STATUS
-node01.bond0-eth1-eth2   SuccessfullyConfigured
-node02.bond0-eth1-eth2   SuccessfullyConfigured
+node01.bond0-eth1-eth2   Available
+node02.bond0-eth1-eth2   Available
 ```
 
 By using `-o yaml` we obtain the full status of given Enactment:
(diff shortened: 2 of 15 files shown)


pkg/enactmentstatus/conditions/conditions.go
```diff
@@ -199,10 +225,55 @@ func SetProgressing(conditions *nmstate.ConditionList, message string) {
...
+		"",
+	)
+	conditions.Set(
+		nmstate.NodeNetworkConfigurationEnactmentConditionAvailable,
+		corev1.ConditionUnknown,
+		nmstate.NodeNetworkConfigurationEnactmentConditionMaxUnavailableLimitReached,
+		"",
+	)
+	conditions.Set(
+		nmstate.NodeNetworkConfigurationEnactmentConditionAborted,
```
phoracek, line 274: you set this twice, is that intended?

qinqon: Done

docs/user-guide/102-configuration.md
```diff
@@ -176,8 +175,9 @@ status:
 
 The output contains the `desiredState` applied by the Policy for the given Node.
 It also contains a list of conditions. This list is more detailed than the one
-in Policy. It shows whether the Policy matched given Node
-(`AllSelectorsMatching`), if the `desiredState` is currently being applied on
+in Policy. It shows whether the max unavailable node limits has reached so the
+node has to wait for the networking node to be applied ('Pending'),
```
rhrazdil, line 179: we should use return quotes here: ('Pending') -> (\`Pending\`)

docs/user-guide/102-configuration.md
```diff
@@ -176,8 +175,9 @@ status:
 
 The output contains the `desiredState` applied by the Policy for the given Node.
 It also contains a list of conditions. This list is more detailed than the one
-in Policy. It shows whether the Policy matched given Node
-(`AllSelectorsMatching`), if the `desiredState` is currently being applied on
+in Policy. It shows whether the max unavailable node limits has reached so the
```
rhrazdil, line 178: Maybe we could reformulate this s little, wdyt?
`It shows whether the max unavailable node limits has reached` -> `It shows whether the limit of unavailable nods was reached so the node has to wait for other nodes to finish applying network configuration ('Pending')`

rhrazdil (commented):
Just documentation nits

docs/user-guide/102-configuration.md
qinqon: Done

docs/user-guide/102-configuration.md
qinqon: Done

test/e2e/handler/conditions.go
```diff
@@ -171,3 +158,22 @@ func waitForDegradedPolicy(policy string) {
 func waitForPolicy(policy string, matcher GomegaMatcher) {
 	policyConditionsStatusForPolicyEventually(policy).Should(matcher, "should reach expected status at NNCP '%s', \n current enactments statuses:\n%s", policy, enactmentsStatusToYaml())
 }
+
+func cleanConditionMessageAndTimestamp(receivedConditions shared.ConditionList) shared.ConditionList {
```
rhrazdil, line 162: This function is quite general, is there a reason to name the parameter `receivedConditions` instead of just `conditions`?

Also it may be good to reflect that the parameter is a list in the function name - `cleanConditionsMessageAndTimestamp`, because the next function does that - `matchConditionsFrom`.
But these are a cosmetic things

rhrazdil (commented):
Thanks, this looks very good. Just one comment
/approve

test/e2e/handler/conditions.go
qinqon: done

rhrazdil (approved):
Nice, thanks!

api/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -20,7 +20,7 @@ type NodeNetworkConfigurationEnactmentList struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=nodenetworkconfigurationenactments,shortName=nnce,scope=Cluster
-// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.type==\"Available\")].reason",description="Status"
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.status==\"True\")].type",description="Status"
```
phoracek, line 23: Isn't there a case when two conditions are True? I recall we originally considered allowing Progressing and Degraded being True at the same time, signalizing that the configuration (without rollout) is progressing, but some nodes have already failed. Is this still a possibility and what happens when we have 2x True?

phoracek (commented):
/hold

One small question I'd like to clarify

api/v1alpha1/nodenetworkconfigurationenactment_types.go
qinqon: Only one is true at the same time https://github.com/nmstate/kubernetes-nmstate/blob/master/pkg/enactmentstatus/conditions/conditions.go, if it's more than one is true it will show the first one that contain Status=true.

I can leave this as it was before, the issue was that "MaxUnavailableNodeLimitReached" does not look good to state that the NNCE is kind of queued/pending.

phoracek: I like it better this way. Just making sure we won't end up in inconsistent and confusing state. If there is never double true, this should be fine.

PR #752, opened 2021-05-19
Remove nmstate-config ConfigMap

qinqon:

>
> /kind bug
> /kind enhancement

The current way to filter interfaces from NNS is using a variable at
nmstate-config we are not propagating this to NMState CR or CNAO and
since nmstate-handler is deployed at every node it does not scale well.
This change remove the ConfigMap and keep the pod environment variable.

**Release note**:

```release-note
Remove nmstate-config ConfigMap
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -199,10 +199,7 @@ spec:
                 fieldRef:
                   fieldPath: spec.nodeName
             - name: INTERFACES_FILTER
-              valueFrom:
-                configMapKeyRef:
-                  name: {{template "handlerPrefix" .}}nmstate-config
-                  key: interfaces_filter
+              value: "{veth*,cali*}"
             - name: ENABLE_PROFILER
               value: "False"
             - name: PROFILER_PORT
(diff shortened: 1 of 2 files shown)


docs/user-guide/101-reporting.md
```diff
@@ -177,12 +177,11 @@ clutter the output with all Pod connections. This filtering can be adjusted via
 | `"{veth*,vnet*}"` | Omit all interfaces starting with either `veth` or `vnet`          |
 
 
-To override the default configuration, use `patch` command. Additionally, in
-order for the config to take effect, all handlers need to be restarted:
+To override the default configuration, use `set env` command. Additionally, in
```
RamLavi, line 180: not sure I understand this: to override, don't we need to change the nmstate-handler daemonset's variable? how does `set env` do this?

RamLavi: ah.. you mentioned it 2 lines below.. sorry :)

RamLavi (approved):
/lgtm

deploy/handler/operator.yaml
```diff
@@ -232,12 +229,6 @@ spec:
         - name: nmstate-lock
           hostPath:
             path: /var/k8s_nmstate
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: {{template "handlerPrefix" .}}nmstate-config
-  namespace: {{ .HandlerNamespace }}
 data:
```
rhrazdil, line 232: Why is `data` moved to the daemonset?

qinqon: Badly removed the ConfigMap, fixed now.

rhrazdil (approved):
Thanks!
/lgtm
/approve

Issue #753, opened 2021-05-20
node selector labels cannot be more than 63 characters


qinqon:
**What happened**:
When an NNCP is populated with nodeSelector labels bigger or equal to 63 characters the NNCE's dos not get created and NNCP not applied.

**What you expected to happen**:
Webhook validating the NNCP node selector or at least NNCP conditions showing the error.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- `NodeNetworkState` on affected nodes (use `kubectl get nodenetworkstate <node_name> -o yaml`):
- Problematic `NodeNetworkConfigurationPolicy`:
- kubernetes-nmstate image (use `kubectl get pods --all-namespaces -l app=kubernetes-nmstate -o jsonpath='{.items[0].spec.containers[0].image}'`):
- NetworkManager version (use `nmcli --version`)
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Others:

phoracek:
Failing early is definitely and improvement. However, it does not solve the root of this issue. If somebody uses very long node names, it would be impossible to use knmstate. Should we revisit the naming of NNCE and use a prefix taken from NNCP and complete it with a random suffix (similar to pods that are a part of a deployment)? To keep the NNCE list nice, we can keep NNCE and Node names as metadata and print them in separate columns of `kubectl get nnce`.

phoracek:
Oh my bad, never mind.

dankenigsberg:
What does other resources do in case they have a `nodeSelector` field? I personally dislike validations - it forces knmstate to know what is the limitation of the current k8s version. It's a violation of layering. I would rather see a status message explaining as much as we can guess why the operation does not progress.

qinqon:
> What does other resources do in case they have a `nodeSelector` field? I personally dislike validations - it forces knmstate to know what is the limitation of the current k8s version. It's a violation of layering. I would rather see a status message explaining as much as we can guess why the operation does not progress.

Well we are using the k8s golang lib for the validation os kubernetes-nmstate is not copy-pasting the validations done at k8s, also doing failfast is better IMHO.

dankenigsberg:
Even if you vendor k8s's validation, it would break the layering. If next k8s version changes the max to 64 chars, knmstate would become over-restrictive until you re-vendor and re-compile it.

[...]

PR #754, opened 2021-05-20
Validate NodeSelector with the same rules as label

qinqon:

>
> /kind bug
/kind enhancement

When a List operation is done passing label selector the labels are
validated before execute the query, at kubernetes-nmstate the
NodeSelector field is passed as a label selector to check for the nodes
that matches, if the NodeSelector is not valid the enactments will not
be created user will not know the state of the NNCP. This change
validate the NodeSelector at the webhook to fail fast on its errors.

Closes https://github.com/nmstate/kubernetes-nmstate/issues/753

**Release note**:

```release-note
Validate NodeSelector at the validation webhook.
```

### DIFF
--- a/pkg/webhook/nodenetworkconfigurationpolicy/validation_test.go
+++ b/pkg/webhook/nodenetworkconfigurationpolicy/validation_test.go
@@ -12,12 +12,15 @@ import (
 	"github.com/nmstate/kubernetes-nmstate/pkg/policyconditions"
 )
 
-func p(conditionsSetter func(*shared.ConditionList, string), message string) nmstatev1beta1.NodeNetworkConfigurationPolicy {
+func p(nodeSelector map[string]string, conditionsSetter func(*shared.ConditionList, string), message string) nmstatev1beta1.NodeNetworkConfigurationPolicy {
 	conditions := shared.ConditionList{}
 	conditionsSetter(&conditions, message)
 	return nmstatev1beta1.NodeNetworkConfigurationPolicy{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "testPolicy",
+		},
+		Spec: shared.NodeNetworkConfigurationPolicySpec{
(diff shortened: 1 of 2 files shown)


pkg/webhook/nodenetworkconfigurationpolicy/validation_test.go
```diff
@@ -55,21 +59,53 @@ var _ = Describe("NNCP Conditions Validation Admission Webhook", func() {
...
+			policy:       p(map[string]string{"bad key": "bar"}, policyconditions.SetPolicySuccess, ""),
+			validationFn: validatePolicyNodeSelector,
+			validationResult: []metav1.StatusCause{{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: "invalid label key: \"bad key\": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')",
+				Field:   "spec.nodeSelector",
+			}},
+		}),
+		Entry("policy has node selector value with length beyond the limit", ValidationWebhookCase{
+			policy:       p(map[string]string{"kubernetes.io/hostname": "jumbo-frame-bridge-cnv-qe-infra-23.cnvqe2.lab.eng.rdu2.redhat.com"}, policyconditions.SetPolicySuccess, ""),
```
rhrazdil, line 88: Maybe a generic long FQDN would be better, wdyt?

qinqon: Done

qinqon:
/hold
Looks like enactments size is a problem too.

qinqon:
/hold cancel
Let's merge this, then we do a NNCP name validation.

PR #755, opened 2021-05-20
Change handler toleration to "operator: exists"

cybertron:

>
/kind bug
> /kind enhancement

Since nmstate is to be used for configuration of the network
infrastructure, we want the handler to run on all nodes regardless
of taint. If a network config should not run on a given node, the
NNCP nodeSelector field can be used to accomplish that.

For example, a node with a taint of:

[map[effect:NoSchedule key:node.ocs.openshift.io/storage value:true]]

will currently keep the nmstate handler pod from running there.
However, that is not desirable since it prevents the use of nmstate
to configure networking on that storage node.

This changes the handler toleration to "operator: exists", which
will allow the handler to run on all nodes. The webhook toleration
is left alone since there is no need for that to be running on
nodes with a NoSchedule taint.

**Release note**:

```release-note
Handler pods now have the "operator: exists" toleration so they will run on all nodes. This allows the use of nmstate when configuring networking on NoSchedule tainted nodes.
```

### DIFF
--- a/controllers/nmstate_controller.go
+++ b/controllers/nmstate_controller.go
@@ -154,6 +154,10 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1beta1.NMState) error
 		Operator: corev1.TolerationOpExists,
 		Effect:   corev1.TaintEffectNoSchedule,
 	}
+	operatorExistsToleration := corev1.Toleration{
+		Key:      "",
+		Operator: corev1.TolerationOpExists,
+	}
 	amd64ArchOnMasterNodeSelector := map[string]string{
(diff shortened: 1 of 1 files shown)


cybertron:
/retest

I'm not able to see the logs for some reason, so I'm not sure why the jobs failed. e2e tests were working for me locally.

qinqon:
/retest

failing test feels related
```
E2E Test Suite: [rfe_id:3503][crit:medium][vendor:cnv-qe@redhat.com][level:component]NodeSelector when policy is set with node selector not matching any nodes and we add the label to the node and remove the label again should remove the not matching enactment
```

PR #763, opened 2021-06-04
Don't confict with nodeRunningUpdate same node

qinqon:

/kind bug

When "parallel" option is false if the NNCP status field
nodeRunningUpdate is not empty it means that a node is progressing so
Reconcile return a conflict, in the case of handler restarting before
"unclaiming" the NNCP will be reconcile again and a conflict will be
return. This change fix that by not returning conflict if the handler
restarting is the one that has claim the NNCP.

Depends on https://github.com/nmstate/kubernetes-nmstate/pull/764

**Release note**:

```release-note
Don't conflict a NNCP if progressing handler is restarted.
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -20,9 +20,10 @@ package controllers
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


qinqon:
It was already tested with nmstate and NM pin I am going to force merge this.

PR #770, opened 2021-06-08
Relationship labels, Label self deployed resources

oshoval:
In order to have relationship (app.kubernetes.io/*)
on all deployed resources,
this commit handles the resources that are deployed by itself
such as:
1. Node Network Configuration Enactments
2. Node Network State

The labels are inherited from the pod, if exists.
The pod will have those labels if deployed by OLM / HCO,
otherwise it won't and the labels won't be added to the above
resources.

```release-note
None
```

### DIFF
--- a/api/v1alpha1/nodenetworkconfigurationenactment_types.go
+++ b/api/v1alpha1/nodenetworkconfigurationenactment_types.go
@@ -5,6 +5,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/nmstate/kubernetes-nmstate/api/shared"
+	"github.com/nmstate/kubernetes-nmstate/pkg/names"
 )
 
 // +kubebuilder:object:root=true
@@ -38,9 +39,7 @@ func NewEnactment(nodeName string, policy NodeNetworkConfigurationPolicy) NodeNe
 				{Name: policy.Name, Kind: policy.TypeMeta.Kind, APIVersion: policy.TypeMeta.APIVersion, UID: policy.UID},
 			},
 			// Associate policy with the enactment using labels
-			Labels: map[string]string{
-				shared.EnactmentPolicyLabel: policy.Name,
-			},
+			Labels: names.IncludeRelationshipLabels(map[string]string{shared.EnactmentPolicyLabel: policy.Name}),
 		},
 		Status: shared.NodeNetworkConfigurationEnactmentStatus{
 			DesiredState: shared.NewState(""),
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -1,7 +1,39 @@
 package names
 
+import (
+	"os"
+)
+
 // ManifestDir is the directory where manifests are located.
 var ManifestDir = "./bindata"
 
 // NMStateResourceName is the name of the CR that the operator will reconcile
 const NMStateResourceName = "nmstate"
+
+// Relationship labels
+const COMPONENT_LABEL_KEY = "app.kubernetes.io/component"
+const PART_OF_LABEL_KEY = "app.kubernetes.io/part-of"
+const VERSION_LABEL_KEY = "app.kubernetes.io/version"
+const MANAGED_BY_LABEL_KEY = "app.kubernetes.io/managed-by"
+
+func IncludeRelationshipLabels(labels map[string]string) map[string]string {
+	if labels == nil {
+		labels = map[string]string{}
+	}
+
+	mapLabelKeys := map[string]string{
+		"COMPONENT":  COMPONENT_LABEL_KEY,
+		"PART_OF":    PART_OF_LABEL_KEY,
+		"VERSION":    VERSION_LABEL_KEY,
+		"MANAGED_BY": MANAGED_BY_LABEL_KEY,
+	}
+
+	for key, label := range mapLabelKeys {
+		envVar := os.Getenv(key)
+		if envVar != "" {
+			labels[label] = envVar
+		}
+	}
+
+	return labels
+}
--- a/pkg/names/names_test.go
+++ b/pkg/names/names_test.go
@@ -1,0 +1,90 @@
+package names
+
+import (
+	"os"
+	"reflect"
+
+	. "github.com/onsi/ginkgo"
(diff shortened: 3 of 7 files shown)


oshoval:
@RamLavi 
need please ok-to-test

trying to figure out how to add tests, since e2e are without HCO/OLM
and for unit tests it will need mocking the os.env

pkg/names/names.go
```diff
@@ -1,7 +1,36 @@
...
+
+	for key, label := range mapLabelKeys {
+		envVar := os.Getenv(key)
+		if envVar != "" {
+			labels[label] = envVar
+		}
+	}
+
+	return labels
+}
```
qinqon, line 36: Doing this everything is not needed since env vars are inmutable on pods, let's just labels the map at startup and use it in the code.

qinqon (changes requested):
Init the labels map at startup.

pkg/names/names.go
oshoval: is it fine to do this with a function closure ?
this way it will init it just the first time the function run and on the sequential runs it will use the cached values

api/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -31,16 +32,16 @@ type NodeNetworkConfigurationEnactment struct {
 }
 
 func NewEnactment(nodeName string, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
+	labels := names.GetRelationshipLabels()
```
RamLavi, line 35: @qinqon do we want that set on v1alpha1?
i don't think it's braking anything but not sure it's necessary

RamLavi (commented):
@oshoval, related to your question about unit tests, I think GetRelationshipLabelscan be unit-tested by playing with `os.Setenv()`.

oshoval:
Addressed https://github.com/nmstate/kubernetes-nmstate/pull/770#discussion_r648042017
will work on unit tests

api/v1alpha1/nodenetworkconfigurationenactment_types.go
```diff
@@ -31,16 +32,16 @@ type NodeNetworkConfigurationEnactment struct {
 }
 
 func NewEnactment(nodeName string, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
+	labels := names.RelationshipLabels
+	labels[shared.EnactmentPolicyLabel] = policy.Name
```
oshoval, line 36: this might change the names.RelationshipLabels
checking

oshoval:
Added unit tests

pkg/names/names.go
```diff
@@ -1,7 +1,36 @@
...
+		"MANAGED_BY": MANAGED_BY_LABEL_KEY,
+	}
+
+	for key, label := range mapLabelKeys {
+		envVar := os.Getenv(key)
+		if envVar != "" {
+			RelationshipLabels[label] = envVar
+		}
+	}
+}
```
qinqon, line 36: I think I was wrong, since we are just reading env vars we have no need to do this at startup time, let's just read the env vars in this function and update a map instead of storing at a global var with is a worse solution, sorry about the back and forth.

Also since we have use cases where a label is already defined (enactments) is better to update a map instead of creating a new one `func IncludeRelationshipLabels(map[string]string) map[string]string`, so we can use it declaratively with a little functional programming.  

```golang
Labels: IncludeRelationshipLabels(map[string]string{shared.EnactmentPolicyLabel: policy.Name}), 
```

pkg/names/names_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("InitRelationshipLabels", func() {
+	Context("When env vars are empty", func() {
```
qinqon, line 11: We need to add unit test for env vars not present to ensure that is still working and partial empty and partial not empty.

qinqon (changes requested):
Let's remove the global var, it was a bad idea from myself.

oshoval:
Addressed comments

pkg/names/names_test.go
```diff
@@ -0,0 +1,76 @@
...
+			os.Unsetenv("MANAGED_BY")
+		})
+
+		It("Should return labels with the right values", func() {
+			labels := IncludeRelationshipLabels(map[string]string{"foo": "bar"})
+			Expect(labels["foo"]).To(Equal("bar"))
+			Expect(labels[COMPONENT_LABEL_KEY]).To(Equal(expectedComponentLabel))
+			Expect(labels[PART_OF_LABEL_KEY]).To(Equal(expectedPartOfLabel))
+			Expect(labels[VERSION_LABEL_KEY]).To(Equal(""))
+			Expect(labels[MANAGED_BY_LABEL_KEY]).To(Equal(""))
```
RamLavi, line 73: if `MANAGED_BY_LABEL_KEY` is not set, don't we expect the output label to not include it at all?

pkg/names/names_test.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"os"
+	"reflect"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("IncludeRelationshipLabels", func() {
+	Context("When env vars don't exist", func() {
```
RamLavi, line 12: we should also include the case were the labels are already set (upgrade scenario)

RamLavi (changes requested):
looks good overall, added comments

pkg/names/names_test.go
oshoval: it doesn't exists, however even element that doesn't exist return `""`
i changed it to something more explicit

nice catch

oshoval:
addressed comments

api/v1alpha1/nodenetworkconfigurationenactment_types.go
oshoval: @qinqon can you see this one please ?

PR #771, opened 2021-06-08
Avoid incrementing NRU when NNCE is in progress

rhrazdil:
At the beginning, check if NNCE exists for a policy and is in
progressing state.

If it is, then it means that handler was killed while applying the
desired state - do not increment unavailableNodeCount in such case.

There is still a window where killing handler would cause
inconsistency - between status is set to Failed or Success and the
unavailableNodeCount is decremented.

Deferring the abovementioned status updates to minimize the window
during which killing handler causes inconsistency.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
Fix issue with nodeRunningUpdate being incremented multiple times when hander is killed
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -152,7 +152,7 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 		return ctrl.Result{}, err
 	}
 
-	err = r.initializeEnactment(*instance)
+	previousConditions, err := r.initializeEnactment(*instance)
 	if err != nil {
 		log.Error(err, "Error initializing enactment")
 	}
@@ -171,13 +171,15 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 		return ctrl.Result{}, nil
 	}
 
-	err = r.incrementUnavailableNodeCount(instance)
-	if err != nil {
-		if apierrors.IsConflict(err) {
-			enactmentConditions.NotifyPending()
-			return ctrl.Result{RequeueAfter: nodeRunningUpdateRetryTime}, err
+	if r.shouldIncrementUnavailableNodeCount(previousConditions) {
+		err = r.incrementUnavailableNodeCount(instance)
+		if err != nil {
+			if apierrors.IsConflict(err) {
+				enactmentConditions.NotifyPending()
+				return ctrl.Result{RequeueAfter: nodeRunningUpdateRetryTime}, err
+			}
+			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, err
 	}
 	defer r.decrementUnavailableNodeCount(instance)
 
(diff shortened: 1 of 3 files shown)


controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -152,6 +152,12 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 		return ctrl.Result{}, err
 	}
 
+	enactmentProgressing, err := r.isEnactmentProgressing(instance)
+	if err != nil {
+		log.Error(err, "Error getting enactment status")
+		return ctrl.Result{}, err
+	}
+
 	err = r.initializeEnactment(*instance)
```
qinqon, line 161: Instead of re-read it before this, we can just return here the previousCondition as a pointer so we do something like 

```suggestion
	previousConditions, err = r.initializeEnactment(*instance)
	...
	if isEnactmentProgressing(previousConditions){...
```

qinqon (changes requested):
I think we can return the previous conditions from initializeEnactment, we can do some e2e tests here changing the maxUnavailableField at NNCP and creating manually and enactment with progressing and checking that NNCP finish allright, we can use a  nodeSelector for just one node to check if it works.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -186,13 +188,13 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 	if err != nil {
 		errmsg := fmt.Errorf("error reconciling NodeNetworkConfigurationPolicy at desired state apply: %s, %v", nmstateOutput, err)
 
-		enactmentConditions.NotifyFailedToConfigure(errmsg)
+		defer enactmentConditions.NotifyFailedToConfigure(errmsg)
```
qinqon, line 191: This is dangerous here since we have the other defer `defer policyconditions.Update(r.Client, request.NamespacedName)` we don't know which one will be run first and nncp conditions update depends on enactments since it's calculate from there, I suggest to just move this just before return and add a comment.

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -186,13 +188,13 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
...
 		log.Error(errmsg, fmt.Sprintf("Rolling back network configuration, manual intervention needed: %s", nmstateOutput))
 		return ctrl.Result{}, nil
 	}
 	log.Info("nmstate", "output", nmstateOutput)
 
-	enactmentConditions.NotifySuccess()
+	defer enactmentConditions.NotifySuccess()
```
qinqon, line 197: ditto

qinqon (changes requested):
Let's not update enactment conditins with defer since we don't know the order with the nncp conditions update defer.

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: I'm okay with dropping the `defer` here, as most of the time in which killing handler casues inconsistency is spent in Progressing state.
I would just like to point out that `deferred` calls are stacked and upon return executed in last-in-first-out order https://tour.golang.org/flowcontrol/13

rhrazdil:
>we can do some e2e tests here changing the maxUnavailableField at NNCP and creating manually and enactment with progressing and checking that NNCP finish allright, we can use a nodeSelector for just one node to check if it works.

@qinqon Just to make sure I get you correctly:
1. Create enactment with Progressing condition
2. Create policy with status NodeRunningUpdate already incremented
3. Expect policy to become available

controllers/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -62,4 +71,115 @@ var _ = Describe("NodeNetworkConfigurationPolicy controller predicates", func()
...
+			Expect(shouldIncrement).To(Equal(c.shouldIncrement))
+
+			if shouldIncrement {
+				err := reconciler.incrementUnavailableNodeCount(&nncp)
+				if c.shouldConflict {
+					Expect(err).Should(WithTransform(apierrors.IsConflict, BeTrue()), "should conflict")
+				} else {
+					Expect(err).ToNot(HaveOccurred())
+				}
+			}
```
qinqon, line 131: Can we exercise this code using the Reconcile function instead of the helpers ? If after a refactor one of the helpers is not being called at Reconcile this test will be not converting the issue.

qinqon (changes requested):
If there is not too much overhead let's do the unit test exercising Reconcile.

qinqon:
We will do a fix for the unit test on a follow up 
/lgtm
/approve

qinqon:
@rhrazdil please re-push it so dco job kicks in.

PR #775, opened 2021-06-14
[release-0.43] Don't confict with nodeRunningUpdate same node (#773)

qinqon:
When "parallel" option is false if the NNCP status field
nodeRunningUpdate is not empty it means that a node is progressing so
Reconcile return a conflict, in the case of handler restarting before
"unclaiming" the NNCP will be reconcile again and a conflict will be
return. This change fix that by not returning conflict if the handler
restarting is the one that has claim the NNCP.

>
> /kind bug
> /kind enhancement

/hold
Depends on https://github.com/kubevirt/project-infra/pull/1322

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -20,9 +20,10 @@ package controllers
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


qinqon:
/hold 
Depends on https://github.com/kubevirt/project-infra/pull/1322

PR #776, opened 2021-06-15
[release-0.47] Avoid incrementing NRU when NNCE

qinqon:
At the beginning, check if NNCE exists for a policy and is in
progressing state.

If it is, then it means that handler was killed while applying the
desired state - do not increment unavailableNodeCount in such case.

There is still a window where killing handler would cause
inconsistency - between status is set to Failed or Success and the
unavailableNodeCount is decremented.

>
> /kind bug
> /kind enhancement

Depends on https://github.com/kubevirt/project-infra/pull/1324

**Release note**:

```release-note
NONE
```


qinqon:
/hold
CI missing

qinqon:
/override pull-kubernetes-nmstate-e2e-operator-k8s-0.47

We are not using the installer at OCP 4.8

PR #777, opened 2021-06-16
Control plane, Assign Priority class to pods

oshoval:
Add `system-node-critical` to `nmstate-handler`,
since this pod should run on each node,

Add `system-cluster-critical` to `nmstate-webhook`
and to `nmstate-cert-manager` pods,
since those pods aren't bound to a specific node,
yet those are important control plane pods.

This will make the control plane less sensitive to preemption
than user workloads.

```release-note
None
```


oshoval:
@qinqon need ok to test please and review if possible

oshoval:
I don't think the failure is due to this PR

/retest

PR #778, opened 2021-06-18
Remove go install step

qinqon:

>
> /kind bug
> /kind enhancement

With fully go mod support go install is not needed since go run will use the version from go.mod + tools.go and also cache it also use the ./vendor directory.

**Release note**:

```release-note
NONE
```


rhrazdil (approved):
/lgtm
/approve

PR #779, opened 2021-06-18
Use go cmd from go get

qinqon:

>
> /kind bug
> /kind enhancement

The install-go.sh install the golang version from go.mod under ./build
directory to build the project with a specific version. This is no
longer ncessary since go allows to download a specific version of golang
with "go get golang.org/dl/go1.10.7" and then "go download go1.10.7",
the only issue is it requieres a go to be installed.

Depends on https://github.com/kubevirt/project-infra/pull/1347

**Release note**:

```release-note
NONE
```


qinqon:
we need a prow image that has go 

```
cd /tmp/ && go get golang.org/dl/go1.16
/bin/bash: go: command not found
```

phoracek:
> we need a prow image that has go
> 
> ```
> cd /tmp/ && go get golang.org/dl/go1.16
> /bin/bash: go: command not found
> ```

So we need a Go to install a specific Go?

Makefile
```diff
@@ -60,17 +60,15 @@ BIN_DIR = $(CURDIR)/build/_output/bin/
...
 export GOFLAGS=-mod=vendor
-export GOROOT=$(BIN_DIR)/go/
-export GOBIN=$(GOROOT)/bin/
-export PATH := $(GOROOT)/bin:$(PATH)
 
 export KUBECONFIG ?= $(shell ./cluster/kubeconfig.sh)
 export SSH ?= ./cluster/ssh.sh
 export KUBECTL ?= ./cluster/kubectl.sh
 
+GO_VERSION = $(shell grep "^go " go.mod |awk '{print $$2}')
```
rhrazdil, line 68: typo: $$2 -> $2

rhrazdil (changes requested):

rhrazdil (commented):
What is the benefit compared to using what `install-go.sh` does?

qinqon:
> What is the benefit compared to using what `install-go.sh` does?

Pros
- So we don't have the ./build directory with a ton of binaries
- Binaries will be share between knmstate clones 
- We don't have to maintain the install-go.sh script.

Cons
- A golang version need to be installed

qinqon:
> > we need a prow image that has go
> > ```
> > cd /tmp/ && go get golang.org/dl/go1.16
> > /bin/bash: go: command not found
> > ```
> 
> So we need a Go to install a specific Go?

Yep, go would be requirement for knmstate.

Makefile
qinqon: At Makefiles it has to be `$$` to escape `$`

PR #781, opened 2021-07-05
manifests, Update requests sections

oshoval:
For each of the 3 nmstate pods,
supply the minimum cpu / memory requests.
Values are set according empiric tests.

See https://bugzilla.redhat.com/show_bug.cgi?id=1935218

```release-note
None
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -38,6 +38,10 @@ spec:
           imagePullPolicy: {{ .HandlerPullPolicy }}
           command:
           - manager
+          resources:
+            requests:
+              cpu: "30m"
+              memory: "20Mi"
           env:
             - name: WATCH_NAMESPACE
               value: ""
@@ -118,6 +122,10 @@ spec:
           imagePullPolicy: {{ .HandlerPullPolicy }}
           command:
           - manager
+          resources:
+            requests:
+              cpu: "30m"
+              memory: "30Mi"
           env:
             - name: WATCH_NAMESPACE
               value: ""
@@ -204,6 +212,10 @@ spec:
           imagePullPolicy: {{ .HandlerPullPolicy }}
           command:
             - manager
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "100Mi"
           env:
             - name: WATCH_NAMESPACE
               value: ""


oshoval:
@qinqon i need please ok-to-test

deploy/handler/operator.yaml
```diff
@@ -118,6 +122,10 @@ spec:
           imagePullPolicy: {{ .HandlerPullPolicy }}
           command:
           - manager
+          resources:
+            requests:
+              cpu: "5m"
```
phoracek, line 127: This does not look right. I have a feeling that with 5m, the application would be too slow to be useful. Have you tried running with 5m set as the limit, to see if it is able to run that way?

phoracek (changes requested):

deploy/handler/operator.yaml
oshoval: Its the request, it gives at least 5m and it can breath and extend, 
before that it was guaranteed 0m
We can't give it 5m limit even for test because it does breath more than that, but we dont set the highest

The actual numbers i measured (without setting limits that will throttle it in case of CPU limits from what i understand)

```
plain start |   3m / 18Mi
steady state |  4m / 22Mi
```

we can see that even on the start it used 3m, and reached 4m on the stead state
I can try to set limit as 5 if you desire, but its not reflect the true case, just very corner case when it will get exact 5m and wont breath

phoracek: I know it is just a request. But it is not more useful than 0m if it does not guarantee the application being useful, and that would defy the purpose of this effort.

Setting the request value as limit is a simple way to confirm that the application will be able to serve even under the tight resource situation.

oshoval: cpu limit 5m or 10m

It works but:
the secret takes more time to be created (but its created after 2 minutes since pod creation)
nmstate-webhook will be on `ContainerCreating` for around 2.5 - 5 minutes more than usual because it doesnt have the secret that it needs IIUC.
Eventually it will find the secret and will be `Running`

with 30m
secret created after 80 seconds
webhook turned to Fully (1/1) Running after 2m40s 

I will increase it then to 30m
(also the KMP cert-manager change which is already merged)

unless you think otherwise

thanks

phoracek: Yeah, let's do 30m on both. It is not that high, considering that both knmstate and KMP are heavy projects (compared to CNIs)

oshoval:
addressed comments
thanks

deploy/handler/operator.yaml
oshoval: KMP counterpart https://github.com/k8snetworkplumbingwg/kubemacpool/pull/320

deploy/handler/operator.yaml
```diff
@@ -204,6 +212,10 @@ spec:
           imagePullPolicy: {{ .HandlerPullPolicy }}
           command:
             - manager
+          resources:
+            requests:
+              cpu: "50m"
```
oshoval, line 217: maybe worth to be checked as well ?
atm k8s nmstate is pretty cookie monster

```
plain start |  33m / 67Mi
create / delete 100 bridges |   821m / 108Mi
Steady state after it |   38m / 86Mi
```

phoracek: For the runtime during configuration, I would keep this lower than 0.8 CPU for sure. We know it is very hungry, but I don't think we can justify reserving that much resources for spikes during the configuration.

It would be interesting to see if knmstate that is limited for 50m is still able to perform the whole configuration (although slower) or not

deploy/handler/operator.yaml
oshoval: Note: without limit it takes around 50 seconds total for the nmstate-webhook to become Running

deploy/handler/operator.yaml
oshoval: 50m is not good enough (Processing for unreasonable time / forever), looking to find a better number

oshoval: how many nodes we prefer to be on the cluster?

note:
1 node, latest CNAO, 1 bridge on non primary interface - takes 110 seconds to be created
without setting any limits

phoracek: Since the effort to define limits is mostly driven by single node usecases, we have it simple here.

oshoval: 50m works but 3 times slower, and it might also not be safe, updated this with some other changes, tested them all together on CNAO

oshoval:
addressed comments

oshoval:
The failures are not due to this PR 
appears also on
https://github.com/kubevirt/cluster-network-addons-operator/pull/923

oshoval:
I dont think failures here are due to the PR
one of them at least (the future one) i saw in the past

deploy/handler/operator.yaml
```diff
@@ -204,6 +212,10 @@ spec:
           imagePullPolicy: {{ .HandlerPullPolicy }}
           command:
             - manager
+          resources:
+            requests:
+              cpu: "200m"
```
phoracek, line 217: This means that we will consume 20 percent of a CPU at all times. I wonder if we could be more conservative here while still offering good service

oshoval: from what i understand and checked (kubectl top pod) unless i am wrong
we don't consume it all the time, the cpu request also determines the share of the cpu and then the pods are "nice" to each other, but please correct me if i am wrong

I can check and reduce it to 150m if it will work good enough,
50m took 3 times the regular 2 minutes that it takes without limits
it takes 2 minutes minimum which is pretty long anyhow

150 would be fine ? or to try 100m / 120m even if it hit performance ?

phoracek: Keep me honest here, but would not that mean that idle knmstate is occupying space that could've been used by workload?

I think that even 50m should do as long as the configuration then succeeds... The point is that it would remain operational and won't be insanely slow. 3 times the time does not sound that terrible, especially if it allows for better use of the cluster. Setting these values too high may increase requirements of the whole HCO by too much.

Let's start with something between 50 and 150 if possible.

oshoval: by space i guess you meant CPU and not RAM, since this is the discussion
from what i saw, i am not sure it takes it on idle,
anyhow i will ask that offline to verify, and meanwhile will try to check 50 - 150m

oshoval: reduced to 100m (operates as good as without limits)
unless you really want below it

phoracek: Thanks sounds like a good start

phoracek (commented):
/lgtm

oshoval:
Thanks Petr

@qinqon can you take a look please?

oshoval:
/retest

(happens also on https://github.com/nmstate/kubernetes-nmstate/pull/784)

qinqon:
Future lane failure is related to some breaking changes at nmstate-git they are removing the json schema dependency 

/lgtm
/approve

PR #793, opened 2021-08-04
Use nmstate API for setting port vlan trunks in favor of using vlan-filtering script

### LINKED ISSUE #171: Support vlan-filtering configuration
Starting with version 0.3, nmstate allows users to configure [vlan-filtering options](https://github.com/nmstate/nmstate/pull/892), e.g.:

```yaml
interfaces:
  - name: linux-br0
    type: linux-bridge
    state: up
    bridge:
      port:
        - name: eth1
          vlan:
            mode: trunk
            trunk-tags:
            - id: 101
            - id-range:
                min: 500
                max: 599
            tag: 100
            enable-native: true
```

kubernetes-nmstate however handles the vlan-filtering configuration with a [workaround script](https://github.com/nmstate/kubernetes-nmstate/blob/master/build/bin/vlan-filtering) which makes sure to enable vlan-filtering by default and treat all bridge ports as trunks.

In order to introduce support for this feature while maintaining backwards compatibility, we should:

* Drop the custom script.
* In case no VLAN configuration is passed by the user, we should fill in defaults matching the behaviour of the script.

### Original description

Until we don't use NetworkManager to set vlan filtering (with or without nmstate) the linux bridge vlan_filtering flag is only set after desiredState is processed, so after reboot the flag is not going to be there until kubernetes-nmstate is not up and NodeNetworkState processed.

rhrazdil:

>
> /kind bug
/kind enhancement

This PR removes usage of vlan-filtering script which is hardcoded to set vlan trunk on every port of every bridge that is
added via NNCP desiredState.
Since nmstate now provides API for this configuration, there is no need for using the script.

The current behaviour is not changed for backwards compatibility.
When a NNCP adding a linux-bridge with state `up` is created, default vlan configuration is added to all bridge ports that do not define custom `vlan` configuration.

This at the same time allows users to define custom vlan configuration, we our script no longer interferes.

Closes #171 

```release-note
NONE 
```


phoracek (commented):
Would you please introduce a new doc example with this new option? It is not a mandatory - feel free to keep this one focus on replacement of internals.

pkg/helper/bridges_test.go
```diff
@@ -89,51 +140,54 @@ var _ = Describe("Network desired state bridge parser", func() {
...
-			Expect(obtainedBridgesAndPorts).To(HaveKeyWithValue("br1", BeEmpty()))
+			Expect(updatedDesiredState).To(MatchYAML(desiredState))
 		})
 	})
 	Context("when there are bridges up", func() {
 		BeforeEach(func() {
 			desiredState = someBridgesUp
 		})
-		It("should return the map with the bridges and ports", func() {
+		It("should add default vlan filtering to linux-bridge ports", func() {
```
AlonaKaplan, line 179: I would add another test case for -
`When("some ports have vlan configuration while other don't"`...

test/e2e/handler/simple_bridge_and_bond_test.go
```diff
@@ -118,9 +118,9 @@ func bondUpWithEth1Eth2AndVlan(bondName string) nmstate.State {
 
 var _ = Describe("NodeNetworkState", func() {
 	Context("when desiredState is configured", func() {
-		Context("with a linux bridge up with no ports", func() {
+		Context("with a linux bridge up with some ports", func() {
```
AlonaKaplan, line 121: You wrote on the commit message -
```
linux-bridge CNI configures vlan_filtering flag for veth pairs, so there shouldn't be a need for it here.
```
Can you please explain what is vlan filtering on veth pair and what problem it solves?
I mean, what if the CNI didn't configure the vlan filtering on the veth, would it cause any problem?

AlonaKaplan:
General question, why not using `NodeNetworkConfigurationPolicy mutating webhook` to mutate the policy with the default vlan filtering values instead of doing the same operation separately for each node?

pkg/helper/client.go
```diff
@@ -126,6 +126,12 @@ func ApplyDesiredState(client client.Client, desiredState shared.State) (string,
 	// working fine after apply
 	probes := probe.Select(client)
 
+	desiredState, err := applyDefaultVlanFiltering(desiredState)
```
AlonaKaplan, line 129: If not doing it in the webhook as suggested [here](https://github.com/nmstate/kubernetes-nmstate/pull/793#issuecomment-896573208), I would at least update the enactment as well. So it will be more obvious that the bridge default value are desired and there is no diff between the desired and the current state.

pkg/helper/bridges.go
```diff
@@ -4,33 +4,58 @@ import (
...
+	ifaces := gjson.ParseBytes(result).Get("interfaces").Array()
+	for ifaceIndex, iface := range ifaces {
+		if !isLinuxBridgeUp(iface) {
+			continue
 		}
+		for portIndex, port := range iface.Get("bridge.port").Array() {
+			if hasVlanConfiguration(port) {
+				continue
+			}
+			result, err = sjson.SetBytes(result, fmt.Sprintf("interfaces.%d.bridge.port.%d.vlan", ifaceIndex, portIndex), defaultVlanFiltering)
```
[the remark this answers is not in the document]
rhrazdil: the pvid is set to 1 by default when vlan is configured in nmstate, I have compared all the bridge properties with this patch and the current state to make sure that there is indeed no difference from current behaviour

test/e2e/handler/simple_bridge_and_bond_test.go
rhrazdil: Now when I think about this sentence, I see is as not relevant really, because this change doesn't affect it in any way.
But since you're asking, the way I understand it is that there are northbound and southbound bridge ports, northbound are the ports configured by nmstate, southbound are the veth pairs created by CNI that connect virt-launcher pod with the bridge.
These veth pairs need to have the vlan filtering configured in compliance to what we configure by nmstate (or the vlan_filtering script respectively) to be able to communicate

pkg/helper/client.go
rhrazdil: Done.
I have added a webhook that adds the default vlan configuration to the NNCP, so both NNCP and NNCEs have the desiredState with default values
I'm still adding the defaults in the Reconcile to make sure old NNCPs have the defaults after nmstate update.

We could also update the NNCP during the Reconcile, which I'm not doing in the current state of this PR, because I'm not convinced it's worth the API load this would trigger - IIUC, each node would reconciled each NNCP twice in cases when the NNCP would be updated right?

rhrazdil: On second thought, given that the webhook should add the defaults in majority of cases, I suppose we can check the default in Reconcile and update the policy case it's needed

rhrazdil: @AlonaKaplan 
After discussion with Petr, we decided it's not needed nor desired to include internal defaults into NNCP. The reason for this being that there are many various defaults set by the system as well.

The current change includes the defaults vlan setting into the NNCE to avoid the discrepancy between desiredState and NNS (since the 'default' we set for VLANs are not really system defaults).

pkg/helper/bridges_test.go
rhrazdil: Added

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -155,6 +155,11 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 		return ctrl.Result{}, err
 	}
 
+	err = r.setPolicyDesiredStateDefaults(instance)
```
AlonaKaplan, line 158: Why are you updating the policy instance and not the enactment directly?
For example. you had to change the `incrementUnavailableNodeCount` behaviour because of the policy instance change, updating the enactment directly would avoid it.

pkg/webhook/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_suite_test.go
```diff
@@ -17,7 +18,8 @@ import (
 
 func TestUnit(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "NNCP Webhook Test Suite")
+	junitReporter := reporters.NewJUnitReporter("junit.webhook-webhook_suite_test.xml")
+	RunSpecsWithDefaultAndCustomReporters(t, "NNCP Webhook Test Suite", []Reporter{junitReporter})
```
AlonaKaplan, line 22: Why is this change needed?

test/e2e/handler/utils.go
```diff
@@ -123,6 +124,12 @@ func updateDesiredStateAtNodeAndWait(node string, desiredState nmstate.State) {
 	waitForAvailableTestPolicy()
 }
 
+func desiredStateWithDefaults(desiredState nmstate.State) nmstate.State {
```
AlonaKaplan, line 127: Please create a manual hard coded version of the desired state.
The e2e tests shouldn't use the same code the dev is using for verification. Bug in the dev code won't be discovered this way.

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: I have considered two options here:
1. update the enactment + apply the defaults to the policy right before setting the desiredState in `pkg/helper/client.go ApplyDesiredState`
2. just update the NNCP, initialize the NNCE based on the updated version and use it to also set the desired state

I considered the 2. option to be more straightforward, because it doesn't mean repeating the same work - parsing the desiredState and applying defaults to it.
It did require to change the `incrementUnavailableNodeCount` so the the function doesn't modify the parameter passed to it, which I think is better anyway.

pkg/webhook/nodenetworkconfigurationpolicy/nodenetworkconfigurationpolicy_suite_test.go
rhrazdil: I noticed that the webhook tests didn't produce xml output, I'll remove this  change from this PR as it's not related anymore

controllers/nodenetworkconfigurationpolicy_controller.go
AlonaKaplan: What about option 3 - use the desired state from the enactment on `ApplyDesiredState`. I believe it is more correct even not related to this PR. Since the enactment represent the desired state of the current node.ApplyDesiredState
And this way you won't have to carry about updating the policy instance.

rhrazdil: Yes, that makes sense as well. Updated accordingly

controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -160,6 +160,12 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 		log.Error(err, "Error initializing enactment")
 	}
 
+	enactmentInstance, err := r.applyEnactmentDefaults(instance)
```
AlonaKaplan, line 163: Not related to this PR, but what is the reason for not returning an error when `initializeEnactment` fails.

AlonaKaplan, line 163: I would move this code to be inside `initializeEnactment` so it will be created with the default from the beginning and there won't be a need for 2 extra api calls (retrieving and updating the enactment).

rhrazdil: I'm not sure if moving this code into `initEnactment` is worth saving the calls, as it would make the function a little too complex IMO.
Mainly because `inititalizeEnactment` returns conditions of the 
WDYT about a compromise - saving 1 API call - by setting the desiredState with defaults in the `initializeEnactment` function and instead of this code only fetch the updated enactment?
That saves one API (write) call.

The main issue I have with moving this whole piece to `initializeEnactment` is that the function returns previous enactment  Status, and I don't like the idea of mixing the previous enactmentStatus with the current one.

rhrazdil: Actually, I think it is related now.
Previously, I suppose, we were okay with the enactment not created, as it didn't have any consequence for applying the policy desiredState.
But the current code passes the desiredState in the enactment, so if it happened to be nil, it would blow.

rhrazdil: addressed in separate commit for easier review https://github.com/nmstate/kubernetes-nmstate/pull/793/commits/c1ec4a4fb04ba0f8c95825112b40c053a032b5c8

rhrazdil: [Updated](https://github.com/nmstate/kubernetes-nmstate/pull/793/commits/c1ec4a4fb04ba0f8c95825112b40c053a032b5c8)

docs/examples/linux-bridge-vlan.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: nmstate.io/v1beta1
```
phoracek, line 1: Could we add it to the list in the README?

docs/examples/linux-bridge-vlan.yaml
```diff
@@ -0,0 +1,29 @@
...
+            - name: eth1
+            vlan:
+              mode: trunk
+              trunk-tags:
+                - id: 101
+                - id-range:
+                    max: 500
+                    min: 200
+              tag: 100
+              enable-native: true
```
phoracek, line 29: Please end it with a newline

docs/examples/linux-bridge-vlan.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: nmstate.io/v1beta1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: linux-bridge-vlan
+spec:
+  desiredState:
+    interfaces:
+      - name: br1
+        description: Linux bridge with eth1 as a port with vlan configuration
```
phoracek, line 9: custom vlan configuration?

phoracek (changes requested):
Minor nits.

Could we also test that we are able to disable vlan-filtering (maybe with vlan: {})?

docs/examples/linux-bridge-vlan.yaml
rhrazdil: Done

docs/examples/linux-bridge-vlan.yaml
rhrazdil: Done

docs/examples/linux-bridge-vlan.yaml
rhrazdil: Done

rhrazdil:
>Could we also test that we are able to disable vlan-filtering (maybe with vlan: {})?


Good point, added this case to e2e (https://github.com/nmstate/kubernetes-nmstate/pull/793/commits/bc345316695f1189e4c6fd9e6c731c19735dcb02#diff-f9955f6099eeab92950ef0ee96d13b0afd37827dbef68b7c9b92c80e9d6aabf6R139) and unit tests as well

phoracek (approved):
This is great stuff, kudos!

/lgtm
/approve

PR #797, opened 2021-08-05
Set max unavailable to 10%

phoracek:

>

/kind enhancement

The default value of maxUnavailable is 1. That serializes upgrades and
may drag for very long time. This change turns it into 10% to speed up
the upgrade process while keeping services reasonably available.

https://bugzilla.redhat.com/show_bug.cgi?id=1990065

**Release note**:

```release-note
Set maximum unavailable replicas of handler to 10%
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -182,6 +182,10 @@ spec:
   selector:
     matchLabels:
       name: {{template "handlerPrefix" .}}nmstate-handler
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 10%
   template:
     metadata:
       labels:


deploy/handler/operator.yaml
```diff
@@ -182,6 +182,10 @@ spec:
   selector:
     matchLabels:
       name: {{template "handlerPrefix" .}}nmstate-handler
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 10%
```
rhrazdil, line 188: kubernetes-nmstate default is 50%, any reason to not keep the same?

rhrazdil: This is after update right?

phoracek: I was following the recommended default recommended in https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#upgrade-and-reconfiguration without giving it a further thought.

This is for the upgrade only AFAIK.

I don't fully follow what is the 50% used for in e2e, do you recall?

rhrazdil: > I was following the recommended default recommended in https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#upgrade-and-reconfiguration without giving it a further thought.

Aaah, of course, I got it. At first I thought that this is somehow setting the NNCP defauld of maxUnavailable, but this is a different maxUnavailable :).

> I don't fully follow what is the 50% used for in e2e, do you recall?
Not sure what you mean, but as I said, the 50% I pointed to is related to NNCP rollout, not handler upgrade :)

/lgtm
/approve

Feel free to unhold if all is clear
/hold

phoracek: Thanks o/

PR #799, opened 2021-08-06
Run handlers on workers only in CI

rhrazdil:

This PR adds nodeSelector to NMState CR, which makes nmstate-handler pods schedule 
only on worker nodes. This aims to mimick clusters that split workloads and control plane.

With master nodes not having nmstate-handler, there is also no NNS, meaning that `allNodes`
variable used in e2e tests looses it's point - removing in favor of keeping only one list of nodes,
which are the worker nodes.

This made some helper functions obsolete, so finally, last commit deletes unused code.

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/sync.sh
+++ b/cluster/sync.sh
@@ -7,8 +7,14 @@ source ./cluster/sync-operator.sh
 
 kubectl=./cluster/kubectl.sh
 
+nmstate_cr_manifest=deploy/crds/nmstate.io_v1beta1_nmstate_cr.yaml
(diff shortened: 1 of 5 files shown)


phoracek (commented):
/approve
/lgtm
/hold

Very nice! Feel free to cancel the hold whenever

rhrazdil:
This should be good to go, thanks!
/unhold

PR #802, opened 2021-08-10
Custom infra tolerations and selectors

### LINKED ISSUE #790: Custom nodeSelector and tolerations for webhook and cert-manager
**What happened**:
I tried to run nmstate into a kubernetes cluster deployed with Rancher.
Since the operator is using default kubernetes master labels and taints the cert-manager and webhook are not starting on a Rancher k8s cluster (they used different labels and taints).

It has to be a way to configure nodeSelector and tolerations for cert-manager and webhook deployments target nodes.

Deploy a rancher cluster. Install nmstate operator (by altering the operator nodeSelector and tolerations manually in the operator manifest). Create a nmstate custom object. Watch webhook and cert-manager deployment pods remaining in Pending.

Rancher uses the following labels for master nodes:

`node-role.kubernetes.io/controlplane: "true"`

and the following tolerations are usually needed (assuming etcd runs alongside kubernetes api and controllers):

```
    - effect: NoSchedule
      key: node-role.kubernetes.io/controlplane
      operator: Exists
    - effect: NoExecute
      key: node-role.kubernetes.io/etcd
      operator: Exists
```
**Environment**:

The issue is not related to network manager itself or to k8s and/or OS version so the environment is not relevant.

brutus333:
Currently, the nmstate operator uses default master nodes labels and taints (from a vanilla k8s deployment) for webhook and cert-manager deployments.
This change allows the customization of both labels and tolerations.

fixes #790

>
> /kind bug

/kind enhancement

Some k8s engine (e.g. Rancher) does not use the default labels and taints - so in this cases nmstate operator can't be deployed.

**Release note**:

```release-note
NONE
```


api/v1beta1/nmstate_types.go
```diff
@@ -34,6 +34,14 @@ type NMStateSpec struct {
 	// If Tolerations is specified, the handler daemonset will be also scheduled on nodes with corresponding taints
 	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+	// WebhookNodeSelector is an optional selector that will be added to webhook & certmanager Deployment manifests
+	// If WebhookNodeSelector is specified, the webhook and certmanager will run only on nodes that have each of the indicated
+	// key-value pairs as labels applied to the node.
+	// +optional
+	WebhookNodeSelector map[string]string `json:"webhookNodeSelector,omitempty"`
```
phoracek, line 41: I think we should not make this specific to implementation details - us using a webhook. Could we instead have it so there is a selector/toleration for "control plane" or "infra" ... not sure about the naming.

brutus333: I thought about using infra initially but control plane looks like a more inspired choice. Should I replaced Webhook with ControlPlane for nodeselectors & tolerations?

phoracek: I think that would be nice, although I'm little conflicted as everything we have is kind of a control plane, mmm.

Anyhow, I would like to hear @rhrazdil input on this as he worked on similar topic in the past.

rhrazdil (commented):
@phoracek  Would it make sense to introduce PlacementConfiguration API similar to what was done in CNAO?
https://github.com/kubevirt/cluster-network-addons-operator/blob/c6a4fba63ee5ebf8612efa4061ea984b6acd00a3/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go#L38

phoracek:
I think it would. But it has one major issue - we would break the current "flat" placement API. I suppose we have to pick between:

Backwards compatible, but asymmetrical API.

```yaml
tolerations:
  ...
nodeSelectors:
  ...
infraTolerations:
  ...
infraNodeSelectors:
  ...
```

Symmetrical API, but I'm not sure if we can really call it "workloads".

```yaml
  placementConfiguration:
    infra:
      nodeSelectors:
        ..
      tolerations:
        ..
    workloads:
      nodeSelectors:
        ..
      tolerations:
        ..
```

I'm slightly leaning towards the flat one, but I suppose it is a matter of taste (and it can bloat to become way uglier). I will not fight for that though. We have to make sure that we handle:

* Proper naming
* Backwards compatibility, possibly implemented internally

brutus333:
@phoracek : I tend to be in favor of symmetrical API instead of flat (data seems better structured) but I don't have a strong opinion on this.
Please let me know what you decide that have to be implemented.

rhrazdil:
@phoracek Correct, the Symmetrical API looks better IMO, but we would need to keep backwards compatibility with current `nodeSelector` field, plus I agree that workloads doesn't fit in the context of knmstate. Infra and Handler would make more sense.

Nice thing about the structured API is that it is (or could be, if the fields are well named) self-explanatory, whereas having just 
```yaml
tolerations:
  ...
nodeSelectors:
  ...
infraTolerations:
  ...
infraNodeSelectors:
```
could force users look into documentation to see what is the meaning.

I'm not strongly opinionated either, going with the flat model seems like the easier path with respect to the already present `nodeSelector` field in the API.

I would consider the structured Symmetric model if we expected NMState CR to have many attributes or more components to apply nodeSelector/tolerations to.
I think we don't expect either of the two.

phoracek:
@brutus333 we discussed this offline with Radim and we'd like to pursue following: Convert the current API into a structured one. The naming and structure should follow this format ...

```yaml
  placementConfiguration:
    infra:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: Exists
    workloads:
      nodeSelector:
        node-role.kubernetes.io/worker: ""
```

While introducing this new structure, we must assure backwards compatibility with the old selector API.

Does this sound fine to you?

brutus333:
> @brutus333 we discussed this offline with Radim and we'd like to pursue following: Convert the current API into a structured one. The naming and structure should follow this format ...
> 
> ```yaml
>   placementConfiguration:
>     infra:
>       affinity:
>         nodeAffinity:
>           requiredDuringSchedulingIgnoredDuringExecution:
>             nodeSelectorTerms:
>             - matchExpressions:
>               - key: node-role.kubernetes.io/master
>                 operator: Exists
>     workloads:
>       nodeSelector:
>         node-role.kubernetes.io/worker: ""
> ```
> 
> While introducing this new structure, we must assure backwards compatibility with the old selector API.
> 
> Does this sound fine to you?

Yes, I'll take it next week, now I' in holiday.

phoracek:
Enjoy your holidays!

phoracek:
I see no reason not to include the new attribute in both versions. Older versions of the operator would just ignore it, new would handle it. I would vote for that.

We have to maintain backwards compatibility, anything else would be a major annoyance to users. Also, the proposed change is quite small and does not merit a major version bump.

Would  think make sense:
* Add the new attribute as proposed in https://github.com/nmstate/kubernetes-nmstate/pull/802#issuecomment-901932168 to all versions of the API spec. 
* Keep the old placement API (NodeSelector and Toletaions), this is only for backwards compatibility. We should change the operator, so it converts this into the new API and writes it back on the API server.

brutus333:
> I see no reason not to include the new attribute in both versions. Older versions of the operator would just ignore it, new would handle it. I would vote for that.
> 
Acknowledge that.
> We have to maintain backwards compatibility, anything else would be a major annoyance to users. Also, the proposed change is quite small and does not merit a major version bump.
> 
What do you mean with major version bump? Are you referring to nmstate project version or to CRD version? I was planning to use v1 for new nmstate CRD version.
> Would think make sense:
> 
> * Add the new attribute as proposed in [Custom infra tolerations and selectors #802 (comment)](https://github.com/nmstate/kubernetes-nmstate/pull/802#issuecomment-901932168) to all versions of the API spec.
> * Keep the old placement API (NodeSelector and Toletaions), this is only for backwards compatibility. We should change the operator, so it converts this into the new API and writes it back on the API server.

I'd keep the flat model for v1beta1 and v1alpha1 versions but I'd add the infraSelector and infraTolerations similar to the code existing now in this PR. Adding additional fields does not invalidate the old APIs but I prefer to not add the structured fields to old API since it can bring confusion (and we might introduce duplicated configuration - for example NodeSelector and Tolerations will be part of both flat and structured data).

A different topic refers to the way now the CRDs are generated: kubebuilder project documents the use of kustomize to patch the CRD with conversion key: https://book.kubebuilder.io/multiversion-tutorial/deployment.html#deployment-and-testing.

But in nmstate operator case it seems that CRDs are generated via Makefile with: `go run sigs.k8s.io/controller-tools/cmd/controller-gen "crd:trivialVersions=true" paths="./..." output:crd:artifacts:config=deploy/crds` and the resulting CRDs are used so no kustomize is called. Is my understanding correct? If so, please give me some hints on how to change the bundling to include conversion key.

controllers/nmstate_controller.go
```diff
@@ -173,12 +173,24 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1beta1.NMState) error
...
+	amd64AndCRInfraNodeSelector := instance.Spec.InfraNodeSelector
+	if amd64AndCRInfraNodeSelector == nil {
+		amd64AndCRInfraNodeSelector = amd64ArchOnMasterNodeSelector
+	} else {
+		amd64AndCRInfraNodeSelector["beta.kubernetes.io/arch"] = "amd64"
+	}
+
+	webhookTolerations := instance.Spec.InfraTolerations
+	if webhookTolerations == nil {
+		webhookTolerations = []corev1.Toleration{masterExistsNoScheduleToleration}
```
qinqon, line 185:
```suggestion
	infraTolerations := instance.Spec.InfraTolerations
	if infraTolerations == nil {
		infraTolerations = []corev1.Toleration{masterExistsNoScheduleToleration}
```

api/v1beta1/nmstate_types.go
```diff
@@ -34,6 +34,14 @@ type NMStateSpec struct {
...
 	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+	// InfraNodeSelector is an optional selector that will be added to webhook & certmanager Deployment manifests
+	// If InfraNodeSelector is specified, the webhook and certmanager will run only on nodes that have each of the indicated
+	// key-value pairs as labels applied to the node.
+	// +optional
+	InfraNodeSelector map[string]string `json:"webhookNodeSelector,omitempty"`
+	// InfraTolerations is an optional list of tolerations to be added to webhook & certmanager Deployment manifests
+	// If InfraTolerations is specified, the handler daemonset will be able to be scheduled on nodes with corresponding taints
+	InfraTolerations []corev1.Toleration `json:"webhookTolerations,omitempty"`
```
qinqon, line 44: This has to be optional too

deploy/crds/nmstate.io_nmstates.yaml
```diff
@@ -89,6 +89,58 @@ spec:
...
+              webhookNodeSelector:
+                additionalProperties:
+                  type: string
+                description: InfraNodeSelector is an optional selector that will be
+                  added to webhook & certmanager Deployment manifests If InfraNodeSelector
+                  is specified, the webhook and certmanager will run only on nodes
+                  that have each of the indicated key-value pairs as labels applied
+                  to the node.
+                type: object
+              webhookTolerations:
```
qinqon, line 101: This CRD has to be re-generated is still pointing to "webhook"

qinqon:
@brutus333 can you squash the commits since they have already mixed contents let's just have one commit.

qinqon (changes requested):
Can you rebase the PR ?

controllers/nmstate_controller.go
brutus333: I changed the code in latest commit.

api/v1beta1/nmstate_types.go
brutus333: Ack. Changed in last commit.

deploy/crds/nmstate.io_nmstates.yaml
brutus333: I regenerated it and push the changes.

qinqon:
/hold

Implementation looks sane but I want to verify integration with cluster-network-addons-operator.

brutus333:
> /hold
> 
> Implementation looks sane but I want to verify integration with cluster-network-addons-operator.

@qinqon : what about schema change? Should we pursue it in current PR or should we put it in a different PR?

qinqon:
So, cluster-network-addons-opertor is affected by it https://github.com/kubevirt/cluster-network-addons-operator/blob/04936fc91abcd86bd9cf9893c2d1386a73981eed/hack/components/bump-nmstate.sh#L37-L43.

qinqon:
> > /hold
> > Implementation looks sane but I want to verify integration with cluster-network-addons-operator.
> 
> @qinqon : what about schema change? Should we pursue it in current PR or should we put it in a different PR?

At schema level we are just adding new features to NMState CR we just have to document it.

Related to CNAO what we have change is the name of the placeholders this is not the schema, CNAO is going to install the manifest it has on it's image so is a 1-1 relation so we don't have to do a migration or anything.

Long store short let's add some docs related to the new knobs at NMState and we will also need to adapt CNAO @rhrazdil are you ok with that ?

rhrazdil:
@qinqon I'm okay with that approach, thanks!

brutus333:
> Long store short let's add some docs related to the new knobs at NMState and we will also need to adapt CNAO @rhrazdil are you ok with that ?

@qinqon: where do you propose to add the documentation? I can't find a place where nmstate CRD is documented (apart from comments in the source code).

qinqon:
> > > /hold
> > > Implementation looks sane but I want to verify integration with cluster-network-addons-operator.
> > 
> > 
> > @qinqon : what about schema change? Should we pursue it in current PR or should we put it in a different PR?
> 
> At schema level we are just adding new features to NMState CR we just have to document it.
> 
> Related to CNAO what we have change is the name of the placeholders this is not the schema, CNAO is going to install the manifest it has on it's image so is a 1-1 relation so we don't have to do a migration or anything.
> 
> Long store short let's add some docs related to the new knobs at NMState and we will also need to adapt CNAO @rhrazdil are you ok with that ?



> > Long store short let's add some docs related to the new knobs at NMState and we will also need to adapt CNAO @rhrazdil are you ok with that ?
> 
> @qinqon: where do you propose to add the documentation? I can't find a place where nmstate CRD is documented (apart from comments in the source code).

I am going to create an issue so we document the NMState CR

/lgtm
/approve

PR #808, opened 2021-08-19
Set the node as the owner reference of the enactmant

AlonaKaplan:

>
/kind bug
> /kind enhancement

If the node is removed the enactment should be garbage collected.
Currently we have a leak.

Setting both the node and the policy as the owner references of the
enactment means that the enactmant will be removed only when both the node
and the policy are removed. Therefore, We have to choose only one of the
two to be the owner ref.
The removal of the other one should be handled by the controller.
We don't have a centrelized controller, node specific controllers are
running on each node.
It doesn't make sense to handle a node removal by a controller running on
the removed node.
Therefore, the node was chosen to be the owner ref while the policy removal
will be handled by the policy controller.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/v1beta1/nodenetworkconfigurationenactment_types.go
+++ b/api/v1beta1/nodenetworkconfigurationenactment_types.go
@@ -33,12 +33,12 @@ type NodeNetworkConfigurationEnactment struct {
 	Status shared.NodeNetworkConfigurationEnactmentStatus `json:"status,omitempty"`
 }
 
-func NewEnactment(nodeName string, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
+func NewEnactment(node *corev1.Node, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
 	enactment := NodeNetworkConfigurationEnactment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: shared.EnactmentKey(nodeName, policy.Name).Name,
+			Name: shared.EnactmentKey(node.Name, policy.Name).Name,
 			OwnerReferences: []metav1.OwnerReference{
-				{Name: policy.Name, Kind: policy.TypeMeta.Kind, APIVersion: policy.TypeMeta.APIVersion, UID: policy.UID},
+				{Name: node.Name, Kind: "Node", APIVersion: "v1", UID: node.UID},
 			},
 			// Associate policy with the enactment using labels
 			Labels: names.IncludeRelationshipLabels(map[string]string{shared.EnactmentPolicyLabel: policy.Name}),
--- a/api/v1beta1/nodenetworkconfigurationenactment_types_test.go
+++ b/api/v1beta1/nodenetworkconfigurationenactment_types_test.go
@@ -1,0 +1,41 @@
+package v1beta1
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("NodeNetworkEnactment", func() {
+	var (
+		nncp = NodeNetworkConfigurationPolicy{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "nmstate.io/v1beta1",
+				Kind:       "NodeNetworkConfigurationPolicy",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "policy1",
+				UID:  "12345",
+			},
+		}
+		node = corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "node1",
+				UID:  "54321",
+			},
+		}
+	)
+
+	Context("NewEnactment", func() {
+		It("should have the node as the owner reference of the created enactment", func() {
+			nnce := NewEnactment(&node, nncp)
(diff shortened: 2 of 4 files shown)


rhrazdil (approved):
Thanks!
/approve
/lgtm

api/v1beta1/nodenetworkconfigurationenactment_types.go
```diff
@@ -33,12 +33,13 @@ type NodeNetworkConfigurationEnactment struct {
...
 
-func NewEnactment(nodeName string, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
+func NewEnactment(node *corev1.Node, policy NodeNetworkConfigurationPolicy) NodeNetworkConfigurationEnactment {
 	enactment := NodeNetworkConfigurationEnactment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: shared.EnactmentKey(nodeName, policy.Name).Name,
+			Name: shared.EnactmentKey(node.Name, policy.Name).Name,
 			OwnerReferences: []metav1.OwnerReference{
-				{Name: policy.Name, Kind: policy.TypeMeta.Kind, APIVersion: policy.TypeMeta.APIVersion, UID: policy.UID},
+				{Name: policy.Name, Kind: policy.Kind, APIVersion: policy.APIVersion, UID: policy.UID},
```
phoracek, line 41: Does not having two reference mean that it will be removed only when both of these are gone? While what we want to do is to remove it as soon as one of them is gone. Please keep me honest here.

phoracek (commented):
/hold

api/v1beta1/nodenetworkconfigurationenactment_types.go
rhrazdil: It should based on a little google research of mine. 
[E2E](https://github.com/nmstate/kubernetes-nmstate/blob/main/test/e2e/handler/nncp_cleanup_test.go) tests should keep us covered as we're testing NNCEs are cleaned up on NNCP deletion

phoracek: Oh, that's awesome then, nice cleanup. Sorry about the raised hold.

rhrazdil: No need to apologize, looks like your concerns are correct!
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/808/pull-kubernetes-nmstate-e2e-handler-k8s/1429712164502900736

phoracek: Bless the tests \o/

rhrazdil:
/lgtm cancel

Looks like NNCEs are indeed not removed with when only one or ownerReferrenced resources is removed.

api/v1beta1/nodenetworkconfigurationenactment_types.go
AlonaKaplan: yep, you're right!

rhrazdil:
Thanks! 
One issue I see with this change is that we may leak NNCEs in scenarios when nodes are rebooted.
Consider:
```
1. create a policy
cat <<EOF | kubectl creat -f -
apiVersion: nmstate.io/v1beta1
kind: NodeNetworkConfigurationPolicy
metadata:
  name: linux-bridge
spec:
  desiredState:
    interfaces:
    - name: br11
      description: Linux bridge with eth1 as a port
      type: linux-bridge
      state: up
      bridge:
        options:
          stp:
            enabled: false
        port:
        - name: eth1
EOF

2. Reboot some node
3. Wait for the node to come up
4. Delete NNCP
```

The result is that the NNCE on the node that was rebooted is still present, because session key of the handler
is not valid anymore. It takes some while for the handler to get restarted after reboot.

IMO it's not a blocker for the PR, but it's something we should be aware of at the very least.


/lgtm

Adding hold to give Petr a chance since he's started review
/hold

phoracek:
I won't get a chance to re-review it today. Please feel free to continue without me.

rhrazdil:
After offline discussion with @AlonaKaplan, we agreed to address the issue with node restart in a follow up PR by checking for orphaned NNCEs at handler startup.

Unholding this
/unhold

AlonaKaplan:
> Thanks!
> One issue I see with this change is that we may leak NNCEs in scenarios when nodes are rebooted.
> Consider:
> 
> ```
> 1. create a policy
> cat <<EOF | kubectl creat -f -
> apiVersion: nmstate.io/v1beta1
> kind: NodeNetworkConfigurationPolicy
> metadata:
>   name: linux-bridge
> spec:
>   desiredState:
>     interfaces:
>     - name: br11
>       description: Linux bridge with eth1 as a port
>       type: linux-bridge
>       state: up
>       bridge:
>         options:
>           stp:
>             enabled: false
>         port:
>         - name: eth1
> EOF
> 
> 2. Reboot some node
> 3. Wait for the node to come up
> 4. Delete NNCP
> ```
> 
> The result is that the NNCE on the node that was rebooted is still present, because session key of the handler
> is not valid anymore. It takes some while for the handler to get restarted after reboot.
> 
> IMO it's not a blocker for the PR, but it's something we should be aware of at the very least.

Good point! I think we can add a finalizar to the policy to solve it. The problem is that we have the same bug if the policy node selector is updated during the node down time and finalizer won't solve it.
So, another option that will solve the two bugs is to cleanup the enactments on sartup (remove orphan enactments and enactments that don't match the policy node selector).

If that's ok, I will do it on a separate PR.

PR #809, opened 2021-08-20
Filter CNI veth interfaces by type and DeviceState (un/managed)

rhrazdil:

>
> /kind bug
/kind enhancement

Commit1:    
Filter interfaces by type and DeviceState
Instead of filtering interfaces in NNS by name regex,
use type attribute and deviceState obtained from NetworkManager.
Drop veth interfaces that are not being managed by NetworkManager.

Commit2:
e2e: Add test verifying interface_filter isn't failing

Since NetworkManager client could fail to connect to dbus
(potentially), we should make sure that we catch such
events.

When the NetworkManager client fails to connect, the NNS
is not filtered and error message is logged by handler.

This commit adds test verifying that handler log doesn't
contain such message.

**Release note**:

```release-note
NNS interfaces are filtered based on type and whether managed by networkManager, instead of using regex
```


pkg/state/filter.go
```diff
@@ -1,47 +1,59 @@
 package state
 
 import (
-	"os"
+	networkmanager "github.com/phoracek/networkmanager-go/src"
```
phoracek, line 4: Awww

test/e2e/handler/nns_filter_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+var _ = Describe("[nns] NNS Interface filter", func() {
+  BeforeEach(func() {
+    // Make sure NNSes are present
+    for _, node := range nodes {
+      key := types.NamespacedName{Name: node}
+      _ = nodeNetworkState(key)
+    }
+  })
+  It("should not log errors related to NNS interface filtering", func() {
```
phoracek, line 19: Is this file formatted with gofmt?

rhrazdil: Nope, I forgot. Fixed

test/e2e/handler/utils.go
```diff
@@ -521,6 +521,11 @@ func vlan(node string, iface string) string {
 	return gjson.ParseBytes(currentStateJSON(node)).Get(vlanFilter).String()
 }
 
+func kubectl(command ...string) (string, error) {
+	out, err := cmd.Kubectl(command...)
+	return out, err
+}
+
```
qinqon, line 528: Why do we need this if we have cmd.Kubectl ?

qinqon (changes requested):
All good just one super small nit.

test/e2e/handler/utils.go
rhrazdil: No reason, removed this, thanks!

PR #810, opened 2021-08-23
handler,webhook:use variable to set replicas

alonSadan:

>
> /kind bug
/kind enhancement

In order to set the number of webhook replicas dynamically, the operator
yaml should accept a variable to set number of replicas. An example use-case
for that would be to set the number of webhook replicas to 1 when running
on a Single-Node-Openshift cluster, and 2 otherwise.

Signed-off-by: alonsadan <asadan@redhat.com>

This PR leaves the the number of replicas as 2, and only defines the variable with a fixed value. In a follow-up PR this could be modified to be set dynamically.
This PR si originated in @phoracek  request [here](https://github.com/kubevirt/cluster-network-addons-operator/pull/973#discussion_r693719016)
**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nmstate_controller.go
+++ b/controllers/nmstate_controller.go
@@ -173,6 +173,8 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1beta1.NMState) error
 		handlerTolerations = []corev1.Toleration{operatorExistsToleration}
 	}
 
+	const WEBHOOK_REPLICAS = int32(2)
+
 	data.Data["HandlerNamespace"] = os.Getenv("HANDLER_NAMESPACE")
 	data.Data["HandlerImage"] = os.Getenv("RELATED_IMAGE_HANDLER_IMAGE")
 	data.Data["HandlerPullPolicy"] = os.Getenv("HANDLER_IMAGE_PULL_POLICY")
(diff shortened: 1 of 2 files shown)


controllers/nmstate_controller.go
```diff
@@ -189,6 +189,7 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1beta1.NMState) error
 	data.Data["CAOverlapInterval"] = ""
 	data.Data["CertRotateInterval"] = ""
 	data.Data["CertOverlapInterval"] = ""
+	data.Data["WebhookReplicas"] = int32(2)
```
rhrazdil, line 192: Could you please move this line next to other Webhook variables? Either before `WebhookNodeSelector` or after `WebhookAffinity`.
Also, it would be nice if you could define a WEBHOOK_REPLICAS constant

rhrazdil (changes requested):

controllers/nmstate_controller.go
alonSadan: [Done.](https://github.com/nmstate/kubernetes-nmstate/compare/bbbb0a1989a87e288fc800d2f875f00d072f3c45..be645d50e4d2f2b3c16050ee4f82a9e5f33e73b2)

alonSadan:
Thanks you @rhrazdil . [Addressed your comments.](https://github.com/nmstate/kubernetes-nmstate/compare/bbbb0a1989a87e288fc800d2f875f00d072f3c45..be645d50e4d2f2b3c16050ee4f82a9e5f33e73b2)

RamLavi (approved):
/lgtm

rhrazdil:
Thanks!
/lgtm

/unhold

alonSadan:
Thanks @phoracek @rhrazdil

phoracek:
We will need to release knmstate in order to propagate this to CNAO. Two patches were added since the last tag: this and Alona's squash of controllers. Our release lane is currently broken due to missing credentials, but I think I can push it manually. @rhrazdil do you think we are in a good spot to tag v0.53.0?

phoracek:
Maybe we could try to get the VLAN PR in to make it more interesting :eyes:

PR #813, opened 2021-08-26
Introduce nodenetworkconfigurationenactment_controller

AlonaKaplan:

>
/kind bug
> /kind enhancement

The purpose of this controller is to clean up enactments their policy was already deleted but for some reason the policy controller missed the policy removal event (for example the node was down when the policy was removed).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nodenetworkconfigurationenactment_controller.go
+++ b/controllers/nodenetworkconfigurationenactment_controller.go
@@ -1,0 +1,112 @@
+/*
+Copyright The Kubernetes NMState Authors.
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/nmstate/kubernetes-nmstate/api/shared"
+	nmstatev1beta1 "github.com/nmstate/kubernetes-nmstate/api/v1beta1"
+	"github.com/nmstate/kubernetes-nmstate/pkg/enactment"
+)
+
+// NodeReconciler reconciles a Node object
+type NodeNetworkConfigurationEnactmentReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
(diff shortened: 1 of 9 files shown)


controllers/nodenetworkconfigurationenactment_controller.go
```diff
@@ -0,0 +1,112 @@
...
+		log.Error(err, "Error retrieving policy")
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{RequeueAfter: enactment.EnactmentRefreshWithJitter()}, nil
+}
+
+func (r *NodeNetworkConfigurationEnactmentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// By default all this functors return true so controller watch all events,
+	// but we only want to watch create/delete for current node.
```
rhrazdil, line 87: `onCreationForThisEnactment` returns `true` only for CreateEvent, I think we are only interested in create event, right?

controllers/nodenetworkconfigurationenactment_controller_test.go
```diff
@@ -0,0 +1,112 @@
...
+		BeforeEach(func() {
+			request.Name = enactment.Name
+		})
+		It("should re-enqueue", func() {
+			result, err := reconciler.Reconcile(context.Background(), request)
+			Expect(err).ToNot(HaveOccurred())
+			expectRequeueAfterIsSetWithEnactmentRefresh(result)
+		})
+	})
+	Context("and policy doen't exist", func() {
```
rhrazdil, line 78: nit: doesn't

pkg/enactment/refresh.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+const (
+	EnactmentRefresh          = 5 * time.Hour
+	EnactmentRefreshMaxFactor = 0.1
+)
+
+// EnactmentRefreshWithJitter add some jitter to to the refresh rate so it does
+// not hit apiserver at the same time.
+func EnactmentRefreshWithJitter() time.Duration {
```
rhrazdil, line 16: Please consider naming this function just `EnactmentRefreshWithJitter`, that way this func would be called as `enactment.RefreshRateWithJitter()`, which reads better

pkg/enactment/refresh.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	EnactmentRefresh          = 5 * time.Hour
+	EnactmentRefreshMaxFactor = 0.1
+)
+
+// EnactmentRefreshWithJitter add some jitter to to the refresh rate so it does
```
rhrazdil, line 14: nit: ...adds ~~some~~ jitter to ~~to~~....

rhrazdil (changes requested):
Nice, I have only a few nits and one suggestion regarding func name.

One concern I thought about is upgrade compatibility with old NNCEs, which are missing the new label.
Since this PR is a solution for corner case, and not primary mechanism for NNCE cleanup, so I believe we're good.

pkg/enactment/refresh.go
AlonaKaplan: done

AlonaKaplan:
> Nice, I have only a few nits and one suggestion regarding func name.
> 
> One concern I thought about is upgrade compatibility with old NNCEs, which are missing the new label.
> Since this PR is a solution for corner case, and not primary mechanism for NNCE cleanup, so I believe we're good.

You're right, the old NNCEs will miss the label. But the old NNCEs have the policy as their owner ref, so we should be covered.

controllers/nodenetworkconfigurationenactment_controller.go
rhrazdil: Thanks, I meant to point out the the comment says 
>// but we only want to watch create/delete for current node.

Which is inconsistent with the code

rhrazdil:
> You're right, the old NNCEs will miss the label. But the old NNCEs have the policy as their owner ref, so we should be covered.

Ahh of course! We're all good then

rhrazdil (approved):
Thanks!
/approve
/lgtm

PR #815, opened 2021-08-31
Try updating nncp with non-cached client as backup

rhrazdil:
After applying a policy, the cached client may return
outdated resource if network configuration breaks
connectivity, which may take up to a minute to refresh again.

A previous commit addressed this by adding a custom backoff
timeout that would wait for the cached client to return
updated resource. But that solution is sub-optimal since
the wait time may take a lot of time, causing the handler to
be effectively stuck.

To improve this behaviour, this change updates NNCP in two
steps
1. Try using cached client
2. Try using non-cached client if all attempts with cached
   client fail on conflict

This way, handler isn't blocking reconcile for long period of time

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -362,9 +362,20 @@ func (r *NodeNetworkConfigurationPolicyReconciler) incrementUnavailableNodeCount
 
 func (r *NodeNetworkConfigurationPolicyReconciler) decrementUnavailableNodeCount(policy *nmstatev1beta1.NodeNetworkConfigurationPolicy) {
 	policyKey := types.NamespacedName{Name: policy.GetName(), Namespace: policy.GetNamespace()}
+	err := tryDecrementingUnavailableNodeCount(r.Client, r.Client, policyKey)
+	if err != nil {
+		r.Log.Error(err, "error decrementing unavailableNodeCount with cached client, trying again with non-cached client.")
+		err = tryDecrementingUnavailableNodeCount(r.Client, r.APIClient, policyKey)
+		if err != nil {
+			r.Log.Error(err, "error decrementing unavailableNodeCount with non-cached client")
+		}
+	}
+}
+
+func tryDecrementingUnavailableNodeCount(statusWriterClient client.StatusClient, readerClient client.Reader, policyKey types.NamespacedName) error {
 	instance := &nmstatev1beta1.NodeNetworkConfigurationPolicy{}
-	err := retry.RetryOnConflict(policyconditions.StatusUpdateRetry, func() error {
-		err := r.Client.Get(context.TODO(), policyKey, instance)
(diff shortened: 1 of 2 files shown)


controllers/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -362,21 +362,30 @@ func (r *NodeNetworkConfigurationPolicyReconciler) incrementUnavailableNodeCount
...
+	if err != nil {
+		r.Log.Error(err, "error decrementing unavailableNodeCount with cached client, trying again with non-cached client.")
+		err = tryDecrementingUnavailableNodeCount(r.Client, r.APIClient, policyKey)
+		if err != nil {
+			r.Log.Error(err, "error decrementing unavailableNodeCount with non-cached client")
+		}
+	}
+}
+
+func tryDecrementingUnavailableNodeCount(cli client.Client, readerCli client.Reader, policyKey types.NamespacedName) error {
```
qinqon, line 375: Looks like controller runtime has the notion of client.Reader and client.Writer, can we replace "cli client.Client" with "writerCli client.Writer" ?

pkg/policyconditions/conditions.go
```diff
@@ -98,12 +89,25 @@ func SetPolicyFailedToConfigure(conditions *nmstate.ConditionList, message strin
...
+		logger.Error(err, "failed to update policy status using cached client. Retrying with non-cached.")
+		err = update(cli, apiReader, apiReader, policyKey)
+		if err != nil {
+			logger.Error(err, "failed to update policy status using non-cached client.")
+		}
+	}
+	return err
+}
+
+func update(apiWriter client.Client, apiReader client.Reader, policyReader client.Reader, policyKey types.NamespacedName) error {
```
qinqon, line 103: Same here

qinqon (changes requested):
Let's try to use client.Writer.

controllers/nodenetworkconfigurationpolicy_controller.go
rhrazdil: Oddly, when I use `client.Writer` or `client.StatusWriter`, I'm getting an error saying that client.Writer/StatusWriter doesn't implement Status()

rhrazdil: Got it, `client.StatusWriter` is the correct type.

rhrazdil: I'm wondering if it's better to call the StatusWriter `cachedStatusWriter` to stress the point that one of the clients is cached and the other is not, but we don't have this notion used elsewhere so it may be just confusing...

qinqon: The fact that is cached is a caller detail not a callee.

rhrazdil: Updated

qinqon (approved):
/lgtm
/approve

PR #818, opened 2021-09-06
e2e: test linux-bridge policy succeeds when updated

rhrazdil:
Reported bug [0] revealed that nmstate verifies whether
currentState and last applied desiredState has changed
(and if it has, raises an error).

This causes errors on NNCP updates with kubernetes-nmstate
v0.52.0 and earlier.

The issue is caused by kubernetes-nmstate itself due to
vlan filtering configuration applied to linux-bridge ports
using a script, instead of nmstatectl.

This commit adds e2e test verifying that policy
adding linux-bridge succeeds when updated.

https://bugzilla.redhat.com/show_bug.cgi?id=2000052

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/simple_bridge_and_bond_test.go
+++ b/test/e2e/handler/simple_bridge_and_bond_test.go
@@ -184,6 +184,9 @@ var _ = Describe("NodeNetworkState", func() {
 					hasVlans(node, secondSecondaryNic, 2, 4094).Should(Succeed())
 				}
 			})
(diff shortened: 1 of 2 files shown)


rhrazdil:
After offline discussion with @qinqon, we have decided that we don't need this test anymore - 
the vlan-filtering script is gone, so we're not doing anything hacky.
nmstate should be able to re-apply the same configuration.
/close

PR #819, opened 2021-09-07
probes, tcp: Fallback to TCP SYNC handshake

qinqon:

>
> /kind bug
/kind enhancement

At some environments the default gw have ICMP blocked, this PR add a
fallback at default gw ping probe when ICMP ping is not working that sends a
TCP SYN packet (just the header) to the default gw at port 80 port.
It wil wait for SYN/ACK (port open) or RST (port not open) in case the
host is up.up.

**Release note**:

```release-note
fallback to TCP SYNC probe is ICMP probe is failing.
```


pkg/probe/probes.go
```diff
@@ -72,6 +73,23 @@ func ping(target string, timeout time.Duration) (string, error) {
 	})
 }
 
+func pingWithTCP(target string, timeout time.Duration) error {
+	tcpingProbe := tcping.NewProbe(tcping.GetInterface(""), 3000*time.Millisecond, true)
+
+	// Port 80 does not need to be open, this just check the tcp handshake with
```
phoracek, line 79: Really? What if firewall straight up rejects our request, then we don't even get to anything listening on TCP, do we?

qinqon: Well maybe reaching the firwall is good enough :-)

qinqon:
/hold

We need some e2e tests disabling ICMP

pkg/probe/probes.go
rhrazdil: The target may always drop the request with no response right? I think no solution is 100%. Using TCP port 80 should have good chance, as it's usually allowed for HTTP traffic...

qinqon:
Deactivating ICMP at dnsmasq container breaks kubevirtci
```
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
Update error: admission webhook "nodenetworkconfigurationpolicies-progress-validate.nmstate.io" denied the request: failed to admit NodeNetworkConfigurationPolicy test-policy: message: policy test-policy is still in progress. 
U
```

EdDev:
> At some environments the default gw have ICMP blocked, this PR add a
> fallback at default gw ping probe when ICMP ping is not working that sends a
> TCP SYN packet (just the header) to the default gw at port 80 port.
> It wil wait for SYN/ACK (port open) or RST (port not open) in case the
> host is up.up.

I am not sure why this is an alternative to ICMP.
If one disables ICMP, I would expect anything else to be dropped as well, otherwise the device is open for attacks (e.g. DoS).

If instead you would have suggested to add another probe which the user can choose from, then it would have made more sense to me.
It is also possible to consider the target IP for this check, it does not have to be the D.G.

qinqon:
> > At some environments the default gw have ICMP blocked, this PR add a
> > fallback at default gw ping probe when ICMP ping is not working that sends a
> > TCP SYN packet (just the header) to the default gw at port 80 port.
> > It wil wait for SYN/ACK (port open) or RST (port not open) in case the
> > host is up.up.
> 
> I am not sure why this is an alternative to ICMP.
> If one disables ICMP, I would expect anything else to be dropped as well, otherwise the device is open for attacks (e.g. DoS).
> 
> If instead you would have suggested to add another probe which the user can choose from, then it would have made more sense to me.
> It is also possible to consider the target IP for this check, it does not have to be the D.G.

I like it better users selecting the IP to use ICMP with a fallback to default gw if not specified, @rhrazdil @phoracek are you ok with that ?

rhrazdil:
> I like it better users selecting the IP to use ICMP with a fallback to default gw if not specified, @rhrazdil @phoracek are you ok with that ?

I think it makes sense for the user to have the option of selecting the target IP for the probe

qinqon:
> > I like it better users selecting the IP to use ICMP with a fallback to default gw if not specified, @rhrazdil @phoracek are you ok with that ?
> 
> 
> 
> I think it makes sense for the user to have the option of selecting the target IP for the probe

Also make per NNCP right ? Not being able to pinging default gw could be a result of a change at network configuration.

rhrazdil:
> Also make per NNCP right ? Not being able to pinging default gw could be a result of a change at network configuration.

I'm wondering, could we get to a situation where the configuration changes and a policy is re-reconciled on the node, causing the probe the worked for the first time to not work for the second time?
I guess it may happen in cases where the desired state is unrelated to reachability of the provided probe target IP?

Having ability for a global target IP could make sense in some environments. Should we provide an option to use domain as well?
In cases where IPs are not static

qinqon:
We are going to do a different approach, allowing the user to provide what to ping/resolv on this scenarios

PR #826, opened 2021-09-09
kubevirtci, Simplify kubevirtci.sh script

oshoval:
The hash is no longer needed.
only that tag is needed in order to determine the kubevirtci
version that should be used.

```release-note
None
```


RamLavi (approved):
/lgtm

oshoval:
Thanks Ram,
@qinqon can you take a look please?

PR #829, opened 2021-09-10
Mount ovs socket

qinqon:

>
> /kind bug
> /kind enhancement

The nmstate OVS functionality depends on the OVS socket to be mounted at
the handler container but this is only possible if OVS is installed at
nodes, this change add optional mount to the handler manifests
controlled by the environment variable ENABLE_OVS on the operator, it
also sets the env variable at the openshift operator dockerfile to have
proper OVS functionality there.

**Release note**:

```release-note
mount OVS socket if needed with ENABLE_OVS env var.
```

### DIFF
--- a/controllers/nmstate_controller_test.go
+++ b/controllers/nmstate_controller_test.go
@@ -305,6 +305,65 @@ var _ = Describe("NMState controller reconcile", func() {
 			Expect(anyTolerationsPresent(infraTolerations, ds.Spec.Template.Spec.Tolerations)).To(BeFalse())
 		})
 	})
+	Context("when OVS is NOT enabled", func() {
+		var (
+			request ctrl.Request
+		)
+		BeforeEach(func() {
+			s := scheme.Scheme
+			s.AddKnownTypes(nmstatev1beta1.GroupVersion,
+				&nmstatev1beta1.NMState{},
+			)
+			objs := []runtime.Object{&nmstate}
+			// Create a fake client to mock API calls.
+			cl = fake.NewFakeClientWithScheme(s, objs...)
+			reconciler.Client = cl
+			request.Name = existingNMStateName
+			result, err := reconciler.Reconcile(context.Background(), request)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result).To(Equal(ctrl.Result{}))
+		})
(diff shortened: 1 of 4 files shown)


rhrazdil:
/lgtm
/approve

Feel free to unhold when ready
/hold

build/Dockerfile.operator
```diff
@@ -9,4 +9,6 @@ COPY deploy/handler/service_account.yaml /bindata/kubernetes-nmstate/rbac/
 COPY deploy/handler/role.yaml /bindata/kubernetes-nmstate/rbac/
 COPY deploy/handler/role_binding.yaml /bindata/kubernetes-nmstate/rbac/
 
+ENV ENABLE_OVS=""
```
rhrazdil, line 12: Is this correct? AFAIK we currently don't have OVS installed in knmstate CIWhy is this here? AFAIK we currently don't have OVS installed in knmstate CI

qinqon: This is the Dockerfile for openshift so it does not affect the CI

qinqon: Fixed I was all wrong

controllers/nmstate_controller_test.go
```diff
@@ -305,6 +305,65 @@ var _ = Describe("NMState controller reconcile", func() {
 			Expect(anyTolerationsPresent(infraTolerations, ds.Spec.Template.Spec.Tolerations)).To(BeFalse())
 		})
 	})
+	Context("when OVS is NOT enabled", func() {
+		var (
+			request ctrl.Request
+		)
+		BeforeEach(func() {
+			s := scheme.Scheme
```
rhrazdil, line 313: nit: We duplicate this block of code a lot, couldn't we move the BeforeEach one level up and share this setup for both specs?
Not a must for this PR, but it wouldn't be a bad thing IMO

rhrazdil (commented):
Looks good. One comment regarding the test setups, but that's not a must for this PR.
/lgtm

controllers/nmstate_controller_test.go
qinqon: It's very declarative, I prefer to do this as follow up since the issue is fixing is important for OVNKubernetes+kubevirt

PR #838, opened 2021-09-29
Return nil when reached maxUnavailable count

rhrazdil:
kubernetes requeues request with rate limiting when an error
is returned, regardless of what is in the returned Result.
See [k8s controller](https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/internal/controller/controller.go#L297)

As a consequence, in big clusters, nodes that wait for policy
need to wait for increasingly long time between each reconcile
loop.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
Fix issue with handlers waiting in Pending state for longer than neccessary
```

### DIFF
--- a/controllers/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/nodenetworkconfigurationpolicy_controller.go
@@ -185,7 +185,8 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 		if err != nil {
 			if apierrors.IsConflict(err) {
 				enactmentConditions.NotifyPending()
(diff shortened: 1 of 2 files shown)


rhrazdil:
pull-kubernetes-nmstate-e2e-handler-k8s-future failed on a known flake, 
previous failures were caused by quay.io [outage](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/838/pull-kubernetes-nmstate-e2e-handler-k8s/1443486943055712256)

PR #840, opened 2021-09-30
[release-0.47] Return nil when reached maxUnavailable count

rhrazdil:
This is a cherry-pick of #838

kubernetes requeues request with rate limiting when an error
is returned, regardless of what is in the returned Result.

As a consequence, in big clusters, nodes that wait for policy
need to wait for increasingly long time between each reconcile
loop.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
Fix issue with handlers waiting in Pending state for longer than neccessary
```


rhrazdil:
The job ran out of time before finishing
/retest

PR #843, opened 2021-10-01
Use ISO8601 timestamp format for zap log messages

creydr:

/kind enhancement

Currently the timestamps in the logs are in epoch format. This makes it really hard to get the time of the event. This PR uses the ISO8601 format for encoding timestamps:

```
{"level":"info","ts":"2021-10-01T11:42:17.421Z","logger":"controller-runtime.manager.controller.nodenetworkstate","msg":"Starting Controller","reconciler group":"nmstate.io","reconciler kind":"NodeNetworkState"}
```
instead of 
```
{"level":"info","ts":"1633079530.9509594","logger":"controller-runtime.manager.controller.nodenetworkstate","msg":"Starting Controller","reconciler group":"nmstate.io","reconciler kind":"NodeNetworkState"}
```

In addition it adds the following flags (which are from the operator-sdk prior v1.0.0):

* `zap-devel`: Enable zap development mode (changes defaults to console encoder, debug log level, disables sampling and stacktrace from 'warning' level)")
* `zap-encoder`: Zap log encoding (`json` or `console`)
* `zap-level`: Zap log level (one of `debug`, `info`, `error` or any integer value > 0)
* `zap-sample`: Enable zap log sampling. Sampling will be disabled for integer log levels > 1
* `zap-time-encoding`: Sets the zap time format (`epoch`, `millis`, `nano`, or `iso8601`)
* `zap-stacktrace-level`: Set the minimum log level that triggers stacktrace generation

The `--v` parameter is marked as deprecated in favor of `zap-devel`.

The code for defining the flags, parsing them, etc. originated from the [operator/operator-sdk](https://github.com/operator-framework/operator-sdk/blob/v0.19.4/pkg/log/zap) repository. Unfortunately this was removed since v1.0.0.

**Release note**:

```release-note
Deprecates the --v flag in favor of the --zap-devel flag
Uses ISO8601 format for timestamp encoding of log messages from zap logger
action required: if your backend systems rely on timestamps in the epoch format, remove the --zap-time-encoding=iso8601 flag, as epoch is the default.
```


phoracek:
Thanks for the contribution @creydr

@rhrazdil would you please help review this PR?

rhrazdil:
Hello @creydr , thanks for your PR.
I like the idea of making the timestamp more human-readable, however, I'm not sure if this PR in it's current state
is the best approach.
I don't think it's desirable to bring in zap package - we would then effectively became maintainers of the code.
And it's a lot of code.

Why not define a configuration with the said `ISO8601` format (we can discuss potential alternatives later) and passing it as 
an option to `zap.New`?

Something like this might work, of course ideally wrapped in a function and not directly in `main.go`, but you get the point:
```
	ecfg := zapcore.EncoderConfig{
		// Keys can be anything except the empty string.
		TimeKey:        "ts",
		LevelKey:       "level",
		NameKey:        "name",
		CallerKey:      "caller",
		FunctionKey:    zapcore.OmitKey,
		MessageKey:     "msg",
		StacktraceKey:  "error",
		LineEnding:     zapcore.DefaultLineEnding,
		EncodeLevel:    zapcore.LowercaseLevelEncoder,
		EncodeTime:     zapcore.ISO8601TimeEncoder,
		EncodeDuration: zapcore.StringDurationEncoder,
		EncodeCaller:   zapcore.ShortCallerEncoder,
	}
	encoder := zapcore.NewJSONEncoder(ecfg)
	ctrl.SetLogger(zap.New(zap.UseDevMode(logType != "production"), zap.Encoder(encoder)))
```

rhrazdil:
>>Thanks! Just one small thing, wouldn't you mind squashing the two commits into one commit enriched by a brief description of the change?

>If this is the usual way you do it here I can squash them. Anyhow I kept them separate as they are 2 different changes actually (whereas one was kind the "base" for the other). So let me know if this makes sense for you, or if I should squash them.
I enriched the commit messages with a brief description.

Thanks for the descriptions, this change makes sense as it is now. No need to squash.

/lgtm
/approve

PR #844, opened 2021-10-04
e2e: wait for re-reconcile after node reboot

### LINKED ISSUE #192: Default route at bridge e2e test is not working
We have mark it as pending, but we need to debug it, fix it and un-pending it.

rhrazdil:
Instead of waiting for update of LastSuccessfulUpdate of NNS,
wait for enactment to become progressing and then policy to
be Available again.

The reason we can't just wait for the timestamp update after
node reboot is that when a node boots up, the policy
reconcile may happen before NNS update, which means that the
NNS still contains the state with the policy configuration.
In this case, there is no change in the node state, which
means NNS is not updated (because there's no difference of the node state).

With that said, we still need to make sure we wait for the policy
is reconciliation after reboot, before verifying the configuration
is present. This change uses the enactment and policy condition
transitions instead of the timestamp.

See this [comment](https://github.com/nmstate/kubernetes-nmstate/issues/192#issuecomment-932324882) for detailed explanation

Fixes #192 

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/default_bridged_network_test.go
+++ b/test/e2e/handler/default_bridged_network_test.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
+	enactmentconditions "github.com/nmstate/kubernetes-nmstate/pkg/enactmentstatus/conditions"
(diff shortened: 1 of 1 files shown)


qinqon:
Super awesome catch !!!

/lgtm
/approve

PR #845, opened 2021-10-04
Fix MISSING values in handler logs

tompsota:
The commit fixes bug in func Byf, where all arguments were passed as the first argument of the message, while all other arguments were left empty.

Example:

Now, all arguments are passed to the first argument only, causing other arguments to display `%!{}(MISSING)` values.
`STEP: Check [node02 eth0 2 4094] has %!s(MISSING) with vlan filtering vids %!d(MISSING)-%!d(MISSING)`

This is fixed in this PR, so arguments are correctly placed to its positions.
`STEP: Check node02 has eth0 with vlan filtering vids 2-4094`

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


rhrazdil:
Thanks!
/lgtm
/approve

PR #865, opened 2021-11-03
Fix NodeNetworkState e2e tests to use primary NIC name from environment variable

creydr:

>

/kind bug

> /kind enhancement

Fixes the e2e tests of NodeNetworkState to use the primary NIC configured via env vars instead of hard coded interface name. Hard coding the interface name can cause issues, if the primary NIC isn't named `eth0` (as done in http://github.com/openshift/kubernetes-nmstate):

```
• Failure [5.086 seconds]
NodeNetworkState
kubernetes-nmstate/test/e2e/handler/error_messages_formatting_test.go:51
  with invalid field
  kubernetes-nmstate/test/e2e/handler/error_messages_formatting_test.go:67
    should keep desired parts of the message [It]
    kubernetes-nmstate/test/e2e/handler/error_messages_formatting_test.go:87

    Expected
        <string>: error reconciling NodeNetworkConfigurationPolicy at desired state apply: ,
        failed to execute nmstatectl set --no-commit --timeout 480: 'exit status 1'
        libnmstate.error.NmstateLibnmError
          Activate profile uuid:99eefc3b-362f-4209-b9a6-ee141ba01e5e iface:eth0 type
            ethernet failed
              error=nm-manager-error-quark
                No suitable device found for this connection (device enp1s0 not available because profile is not compatible with device (mismatching interface name)). (3)
        
    to contain substring
        <string>: libnmstate.error.NmstateVerificationError

```

**Release note**:
```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/error_messages_formatting_test.go
+++ b/test/e2e/handler/error_messages_formatting_test.go
@@ -1,6 +1,8 @@
 package handler
 
 import (
+	"fmt"
+
 	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
 	enactmentconditions "github.com/nmstate/kubernetes-nmstate/pkg/enactmentstatus/conditions"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


rhrazdil (commented):
Thanks!
/approve
/lgtm

PR #866, opened 2021-11-03
Add handler E2E tests for DNS configuration

yboaron:

/kind enhancement

What this PR does / why we need it:
This PR adds E2E DNS tests for handler

Signed-off-by: Yossi Boaron yboaron@redhat.com

### DIFF
--- a/test/e2e/handler/dns_test.go
+++ b/test/e2e/handler/dns_test.go
@@ -1,0 +1,106 @@
+package handler
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
+)
+
+func dnsConfig(searchDomain1, searchDomain2, server1, server2, ipFamily, ipFamily1, dnsTestNic string) nmstate.State {
+	return nmstate.NewState(fmt.Sprintf(`dns-resolver:
+  config:
+   search:
+   - %s
+   - %s
+   server:
+   - %s
+   - %s
+interfaces:
+- name: %s
+  type: ethernet
+  state: up
+  %s:
+    auto-dns: false
+    enabled: true
+    dhcp: true
+  %s:
+    auto-dns: false
+
+`, searchDomain1, searchDomain2, server1, server2, dnsTestNic, ipFamily, ipFamily1))
+}
+
+func dnsAbsent(dnsTestNic string) nmstate.State {
+	return nmstate.NewState(fmt.Sprintf(`dns-resolver:
+  config:
+    state: absent
+interfaces:
+- name: %s
+  type: ethernet
+  state: up
+  ipv4:
+    auto-dns: true
+  ipv6:
+    auto-dns: true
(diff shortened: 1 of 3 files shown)


qinqon:
/hold

Let's first fix dns at NNS with https://github.com/nmstate/kubernetes-nmstate/pull/867

qinqon:
@yboaron can you use NNS to check now that we show the dns-resolve there ?

qinqon:
@yboaron looks like the e2e tests are failing, can you check it out before doing any review ?

yboaron:
> @yboaron looks like the e2e tests are failing, can you check it out before doing any review ?

Hmm, I'm not sure why tests fail, on OCP (virtual baremetal deployment) both tests are green

yboaron:
Seems that second (with V6 U/S servers) dns always fails while the test with V4 U/S servers is green, I assume that 'clearing' [1] DNS configuration doesn't work.

[1]
dns-resolver:
  config:
    state: absent
interfaces:
- name: %s
  type: ethernet
  state: up
  %s:
    enabled: false

qinqon:
/hold cancel
Now the approach looks good we have to fix the CI

qinqon:
/hold
We are going to focus the DNS test, something is fishy at CI.

test/e2e/handler/dns_test.go
```diff
@@ -0,0 +1,102 @@
...
+					dnsResolverSearchForNodeEventually(node).Should(ContainElement(searchDomain1))
+					dnsResolverSearchForNodeEventually(node).Should(ContainElement(searchDomain2))
+				}
+			})
+		})
+		Context("with V6 upstream servers", func() {
+			BeforeEach(func() {
+				// read primary DNS server from one of the nodes
+				serverList := dnsResolverForNode(nodes[0], "dns-resolver.running.server")
+				updateDesiredStateAndWait(dnsConfig(searchDomain1, searchDomain2, serverList[0], upStream1V6, "ipv6", dnsTestNic))
```
qinqon, line 82: serverList[0] can be an IPv4 but for IPv4 you have auto-dns so it will fail since you cannot configure dns with auto-dns,

You need to pass all the list and disable auto-dns for both IPv4 and IPv6.

qinqon (changes requested):

test/e2e/handler/dns_test.go
```diff
@@ -0,0 +1,106 @@
...
+`, dnsTestNic))
+}
+
+var _ = Describe("Dns configuration", func() {
+	Context("when desiredState is configured", func() {
+		var (
+			searchDomain1 = "fufu.ostest.test.metalkube.org"
+			searchDomain2 = "sometest.com"
+			upStream1V4   = "8.8.9.9"
+			upStream1V6   = "2001:db8::1:2"
```
qinqon, line 56:
```suggestion
			server1V4   = "8.8.9.9"
			server1V6   = "2001:db8::1:2"
```

qinqon (changes requested):
nit for naming.

qinqon (approved):
/lgtm
/approve

PR #867, opened 2021-11-04
Show DNS Resolver in NNS

rhrazdil:

>
> /kind bug
/kind enhancement

DNS Resolver is not included in NodeNetworkState, adding types to marshal/unmarshal
DNS data and include it in the NNS as well.

**Release note**:

```release-note
Show dns-resolver in NodeNetworkState
```

### DIFF
--- a/pkg/state/filter_test.go
+++ b/pkg/state/filter_test.go
@@ -341,4 +341,32 @@ routes:
 		})
 	})
 
+	Context("With DNS Resolver populated", func() {
+		BeforeEach(func() {
+			state = nmstate.NewState(`interfaces:
+  - name: eth1
+    state: up
+    type: ethernet
+dns-resolver:
+  config:
+    search:
+    - example.com
+    - example.org
+    server:
+    - 2001:4860:4860::8888
+    - 8.8.8.8
+  running:
+    search:
+    - example.running.com
+    - example.running.org
+    server:
+    - 8.8.4.4`)
+		})
+
+		It("Should keep the DNS Resolver intact", func() {
+			returnedState, err := filterOut(state, ifaceStates)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(returnedState).To(MatchYAML(state))
+		})
+	})
 })
--- a/pkg/state/type.go
+++ b/pkg/state/type.go
@@ -3,12 +3,14 @@ package state
 import (
 	"encoding/json"
 	"fmt"
+
 	"sigs.k8s.io/yaml"
 )
 
 type rootState struct {
-	Interfaces []interfaceState `json:"interfaces" yaml:"interfaces"`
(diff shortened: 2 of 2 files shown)


pkg/state/filter_test.go
```diff
@@ -341,4 +341,32 @@ routes:
 		})
 	})
 
+	Context("With DNS Resolver populated", func() {
+		BeforeEach(func() {
+			state = nmstate.NewState(`interfaces:
+  - name: eth1
```
creydr, line 347: could we use the `firstSecondaryNic` variable here instead of using `eth1` directly? Because the e2e tests in OCP use another NIC name for the first secondary (from `FIRST_SECONDARY_NIC` env var)

rhrazdil: Hey, this is just a unit test and the names don't really matter

creydr: Ah sorry. My fault

qinqon:
So this was introduced at https://github.com/nmstate/kubernetes-nmstate/pull/709, maybe we don't need that anymore.

rhrazdil:
> So this was introduced at #709, maybe we don't need that anymore.

#709 introduces structures for routes and interfaces, but not the dns-resolver.

qinqon:
> > So this was introduced at #709, maybe we don't need that anymore.
> 
> #709 introduces structures for routes and interfaces, but not the dns-resolver.

Looks like nmstate is going to move to yaml 1.2 at the rust implementation, so when we integrate with it we can revert the PR.

qinqon:
/retest

Something is fishy at CI maybe it's time to use podman.

```
creating container buildx_buildkit_kubernetes-nmstate-builder0
#1 22.62 error: failed to list workers: Unavailable: connection error: desc = "transport: error while dialing: dial unix /run/buildkit/buildkitd.sock: connect: no such file or directory"
#1 creating container buildx_buildkit_kubernetes-nmstate-builder0 22.2s done
#1 
```

yboaron:
Thanks for adding this fix.

Checked it in my env - looks OK.

It is worth mentioning that (at least based on what I get in my env) , it seems that dnsresolver equals to  /var/run/NetworkManager/resolv.conf and not /etc/resolv.conf

qinqon:
> Thanks for adding this fix.
> 
> Checked it in my env - looks OK.
> 
> It is worth mentioning that (at least based on what I get in my env) , it seems that nns.dnsresolver equals to /var/run/NetworkManager/resolv.conf and not /etc/resolv.conf

That depends on the NetworkManager configuration, you can configure if you want it to manage /etc/resolv.conf or not, for example at OCP we have NetworkManager not manager /etc/resolv.conf but at the same time we have the prepend dispatcher script.

From an kubernetes-nmstate perspectives safest is to just check NNS.

/lgtm
/approve

PR #873, opened 2021-11-09
Fix handler e2e test interface cleanups

creydr:

>

/kind bug

> /kind enhancement

_tl;dr_: Updates some e2e tests to cleanup only the NICs which got changed during the test.

For the E2E tests of the handler, the primary and both secondary interfaces are only reset on **worker** nodes before the suite starts: https://github.com/nmstate/kubernetes-nmstate/blob/621d3fc9b3ec50a6765b00af379272a9e5c7d495/test/e2e/handler/main_test.go#L77 and https://github.com/nmstate/kubernetes-nmstate/blob/621d3fc9b3ec50a6765b00af379272a9e5c7d495/test/e2e/handler/utils.go#L107-L110
 
The `AfterEach()` function for the whole suite, checks that the interfaces matches the state before the tests are run (compares with the state collected in the `BeforeEach()`): 
https://github.com/nmstate/kubernetes-nmstate/blob/621d3fc9b3ec50a6765b00af379272a9e5c7d495/test/e2e/handler/main_test.go#L113-L121

Some e2e tests call the `resetPrimaryAndSecondaryNICs()` on all nodes (including control-plane nodes). For example see line 41 in:
https://github.com/nmstate/kubernetes-nmstate/blob/621d3fc9b3ec50a6765b00af379272a9e5c7d495/test/e2e/handler/node_selector_test.go#L36-L47

This can cause issues, if the secondary interfaces on the control-plane nodes have been in state `UP` before the suite started, which will cause the suite to fail. Check the following scenario:

1. control-plane node has firstSecondaryNic state set to up
2. E2E test suite starts
3. Before each test, in the BeforeEach(), the current state of control-planes NICs are saved for later comparison
4. The test "cleans up" all interfaces by calling `setDesiredStateWithPolicyWithoutNodeSelector(TestPolicy, resetPrimaryAndSecondaryNICs())`, which sets control-plane node has firstSecondaryNic state set to down (as done for all nodes)
5. The AfterEach defined for the suite compares the state before the test (collected in 3.) with the current state and fails, as the firstSecondaryNic state is now set to down (done in 4.)

This is only an issue, in case the environment against which the E2E tests run, has the firstSecondaryNic and secondSecondaryNic state set to up initially (which e.g. is the case for OCP). Anyhow the tests should clean it up / reset the environment correctly.

This PR addresses this and fixes the tests to clean up properly.

**Special Notes for Reviewers:**
1. In cases where tests should run on each node (including control-plane), e.g. https://github.com/nmstate/kubernetes-nmstate/blob/621d3fc9b3ec50a6765b00af379272a9e5c7d495/test/e2e/handler/nns_update_timestamp_test.go#L68-L91
    I adjusted the test to use a policy which does not affect the nodes secondary NICs (in this case this was possible, by using `linuxBrUpNoPorts()` instead of `linuxBrUp()`).
2. Alternatively to updating the tests, a `resetDesiredStateForControlPlaneNodes()` could have been implemented and called in the `BeforeSuite()` so even the secondaryNics of the control-plane nodes would be "reset", but imo each test should clean up correctly.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/main_test.go
+++ b/test/e2e/handler/main_test.go
@@ -68,6 +68,7 @@ var _ = BeforeSuite(func() {
 	nodeList := corev1.NodeList{}
 	filterWorkers := client.MatchingLabels{"node-role.kubernetes.io/worker": ""}
 	err = testenv.Client.List(context.TODO(), &nodeList, filterWorkers)
+	Expect(err).ToNot(HaveOccurred())
 	for _, node := range nodeList.Items {
 		if containsNode(allNodes, node.Name) {
 			nodes = append(nodes, node.Name)
--- a/test/e2e/handler/nns_update_timestamp_test.go
+++ b/test/e2e/handler/nns_update_timestamp_test.go
@@ -67,14 +67,12 @@ var _ = Describe("[nns] NNS LastSuccessfulUpdateTime", func() {
 	})
 	Context("when network configuration is changed by a NNCP", func() {
 		BeforeEach(func() {
-			// We want to test all the NNS so we apply policies to control-plane and workers
-			setDesiredStateWithPolicyWithoutNodeSelector(TestPolicy, linuxBrUp(bridge1))
+			// We want to test all the NNS so we apply policies to control-plane and workers (use linuxBrUpNoPorts to not affect the nodes secondary interfaces state)
+			setDesiredStateWithPolicyWithoutNodeSelector(TestPolicy, linuxBrUpNoPorts(bridge1))
 			waitForAvailableTestPolicy()
 		})
 		AfterEach(func() {
 			setDesiredStateWithPolicyWithoutNodeSelector(TestPolicy, linuxBrAbsent(bridge1))
-			waitForAvailableTestPolicy()
-			setDesiredStateWithPolicyWithoutNodeSelector(TestPolicy, resetPrimaryAndSecondaryNICs())
 			waitForAvailableTestPolicy()
 			deletePolicy(TestPolicy)
 		})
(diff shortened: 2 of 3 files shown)


rhrazdil (commented):
Hey @creydr, nice PR.
I have one thoughg. What if we saved the initial state of nodes and then applied
desired state with some initial state. That would be for example setting first and second secondary ifaces down.
And at the end, we would restore the initial state to the original.

This way, tests woud still be checking that the state is correct at the end, and most importantly
tests don't need to make assumptions about state of ifaces.
After merging this, we may still break you in the future when we forget about this and merge new test that doesn't follow this.

What do you think?
I'm okay going with this approach, just wondering, it may not be too difficult to make this more robust.

rhrazdil (approved):
We agreed to proceed with the current solution and implement an improvemet as a followup.

The improvement could work as follows:
```
BeforeSuite
   save current state of all nodes
   apply initial 'standardized' desired state, for example primary up, secondaries down 
  BeforeEach
    Store the state
    Do some configuration
  AfterEach
    remove the configuration and check we have the original state (the one store at BeforeEach)
AfterSuite
  restore the saved state of all nodes to the very beginning
```

/approve
/lgtm

creydr:
> Hey @creydr, nice PR. I have one thought. What if we saved the initial state of nodes and then applied desired state with some initial state. That would be for example setting first and second secondary ifaces down. And at the end, we would restore the initial state to the original.
> 
> This way, tests woud still be checking that the state is correct in `AfterEach`, but most importantly tests wouldn't need to make assumptions about state of ifaces. After merging this, we may still break you in the future when we forget about this and merge new test that doesn't follow this.
> 
> What do you think? I'm okay going with this approach, just wondering, it may not be too difficult to make this more robust.

Hi @rhrazdil,
thanks a lot for your feedback. As we discussed in Slack, I would do the following as a follow up on this:

* `BeforeSuite`
  * save current state of all nodes
  * apply initial 'standardized' desired state, for example primary up, secondaries down
* `BeforeEach`
  * Store the state
  * Do some configuration
* `AfterEach`
  * remove the configuration and check we have the original state (the one store at `BeforeEach`)
* `AfterSuite`
  * restore the saved state of all nodes (collected in `BeforeSuite`)

--> we would reset the first- & secondaryNICs of the master nodes at the beginning of the suite as well, but restore the inital configuration of the node at the end of the suite. So tests could reset the NICs of the master nodes as well.

PR #879, opened 2021-11-15
Add docs on using NMState CR knobs

### LINKED ISSUE #822: Add docs on using NMState CR knobs.
**What happened**:
There are some knobs like affinity and node toleration that need to be documented at NMState CR.

**Environment**:

- Others:

tompsota:
The NodeSelector, Tolerations, InfraNodeSelector and InfraTolerations are now documented in the User guide.

resolves #822 

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/user-guide/102-configuration.md
+++ b/docs/user-guide/102-configuration.md
@@ -476,6 +476,22 @@ node05.linux-bridge-maxunavailable   Progressing
 node06.linux-bridge-maxunavailable   Pending
 ```
 
+# Component Placement
+
+In NMState, you can constrain assignment of kubernetes-nmstate components to individual nodes. There are the following options:
+
+- **NodeSelector**
+    - Optional. If specified, handler pods will run only on nodes matching the selector.
+- **Tolerations**
+    - Optional. With tolerations, handler pods will be scheduled onto appropriate nodes.
+- **InfraNodeSelector**
+    - Optional. If specified, webhook and certmanager pods will run only on nodes matching the selector.
+- **InfraTolerations**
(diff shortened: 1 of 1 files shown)


docs/user-guide/102-configuration.md
```diff
@@ -476,6 +476,20 @@ node05.linux-bridge-maxunavailable   Progressing
...
 ```
 
+# Component Placement
+
+In NMState, you can constrain controllers to be assigned to specific nodes using component placement. There are several options:
+
+- **NodeSelector**
+    - Optional selector. If specified, handler will run only on nodes that have matching key-value pairs as labels applied to the node.
+- **Tolerations**
+    - Optional selector. Tolerations work together with taints to ensure that handler is not scheduled onto inappropriate nodes. One or more taints are applied to a node; this marks that the node should not accept any handler that do not tolerate the taint.
```
rhrazdil, line 486: Optional ~~selector~~

I would not go into that level of detail, I believe having the link to k8s documentatiton below is already enough.

`Tolerations applied to handler pods.` Should be enough I think

docs/user-guide/102-configuration.md
```diff
@@ -476,6 +476,20 @@ node05.linux-bridge-maxunavailable   Progressing
...
+- **InfraNodeSelector, InfraTolerations**
+    - Optional selectors. They have similar purpose as NodeSelector and Tolerations, respectively. However, they restrict webhook and certmanager instead of handler.
```
rhrazdil, line 488: Optional ~~selectors~~

It may be better to list both fields separately, with relevant description

rhrazdil (approved):
Thanks!
/lgtm
/approve

PR #883, opened 2021-11-16
handler, container: Use centos stream

qinqon:

>
> /kind bug
/kind enhancement

The NMState team is realsing newer versions of nmstate using centos
stream instead of copr, this change use centos stream to consume those
so we don't need to use copr.

**Release note**:

```release-note
Use quay.io centos stream8 container for handler.
```


yboaron:
Checked it with dns e2e test  PR - looks ok.
/lgtm

rhrazdil (approved):
/approve

PR #891, opened 2021-11-23
Fix issue with missing vlans after upgrade

rhrazdil:
When upgrading from an older knmstate version that uses
vlan-filtering script to a version relying on nmstate
API, nmstate fails on verification when reconciling
bridge with ports configured by the older knmstate
version.

This change applies the vlans using nmcli, so that
there is no discrepancy between kernel and NM configuration.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
Fix issue with missing vlans on linux-bridge ports after upgrade from older versions
```

### DIFF
--- a/pkg/helper/client.go
+++ b/pkg/helper/client.go
@@ -129,6 +129,11 @@ func ApplyDesiredState(client client.Client, desiredState shared.State) (string,
 		return "Ignoring empty desired state", nil
 	}
 
+	out, err := EnableVlanFiltering()
+	if err != nil {
+		return out, fmt.Errorf("failed to enable vlan filtering via nmcli: %s", err.Error())
+	}
+
 	// Before apply we get the probes that are working fine, they should be
 	// working fine after apply
 	probes := probe.Select(client)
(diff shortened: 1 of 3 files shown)


pkg/helper/client.go
```diff
@@ -129,6 +130,19 @@ func ApplyDesiredState(client client.Client, desiredState shared.State) (string,
...
+		return "failed to get currentState", err
+	}
+	upBridgesWithPorts, err := GetUpLinuxBridgesWithPorts(shared.NewState(currentState))
+	if err != nil {
+		return "failed to list bridges with ports", err
+	}
+	out, err := enableVlanFiltering(upBridgesWithPorts)
+	if err != nil {
+		return fmt.Sprintf("failed to enable vlan-filtering via nmcli: %s", out), err
+	}
```
qinqon, line 144: Can we move all this code to bridge.go ? It's going to be easier to remove the code when we need it.

qinqon (changes requested):
Let's put all this custom code to bridge.go it's going to be easier to remove in the future.

pkg/helper/client.go
rhrazdil: Sure, updated

PR #893, opened 2021-11-23
[Release 0.47] Fix issue with missing vlans after upgrade

rhrazdil:
Backport of https://github.com/nmstate/kubernetes-nmstate/pull/891

When upgrading from an older knmstate version that uses
vlan-filtering script to a version relying on nmstate
API, nmstate fails on verification when reconciling
bridge with ports configured by the older knmstate
version.

This change applies the vlans using nmcli, so that
there is no discrepancy between kernel and NM configuration.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
Fix issue with missing vlans on linux-bridge ports after upgrade from older versions
```

### DIFF
--- a/automation/check-patch.e2e-k8s.sh
+++ b/automation/check-patch.e2e-k8s.sh
@@ -33,7 +33,7 @@ main() {
 
     export E2E_TEST_SUITE_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml"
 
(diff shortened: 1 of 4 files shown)


qinqon:
/override pull-kubernetes-nmstate-e2e-handler-k8s-future-0.47

PR #910, opened 2021-11-26
Use Eventually when waiting for policy to be set Degraded

rhrazdil:
This test checks that after first enactment fails configuration, the
policy is set to Degraded.
The problem is, Enactments update their status before the
policy status, so we need to wait for a bit.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


qinqon:
/lgtm
/approve

Can we Chery pick to latest release branch?

Issue #864, opened 2021-11-02
[RFE] Extract nmstate API schema into a dedicated repository


phoracek:
**What happened**:

Due to the API being a part of this repository, anybody who wants to vendor it into their project can suffer a dependency hell. As there are new nmstate API based projects being created, written in Go, this becomes a problem.

**What you expected to happen**:

The API schema should be extracted to a dedicated repository.

filanov:
It doesn't need to be a dedicated repo, maybe just a different sub-module https://github.com/go-modules-by-example/submodules

filanov:
example from hive https://github.com/openshift/hive/tree/master/apis

phoracek:
Thanks @filanov that helps o/

filanov:
Hey @phoracek opened https://github.com/nmstate/kubernetes-nmstate/pull/878

filanov:
This issue can be closed :)

phoracek:
Thanks @filanov, great stuff!

https://github.com/nmstate/kubernetes-nmstate/pull/911

PR #911, opened 2021-11-28
Sub models

filanov:
Is this a BUG FIX or a FEATURE ?:

/kind enhancement

What this PR does / why we need it:

Anyone who wants to vedor APIs into their project can suffer a dependecy hell.
The solution is to use submodels in the prokect to separate the API form the rest of the code and the imports.

Special notes for your reviewer:

Release note:

```release-note
Create a golang submodule for `./api` so it can be imported easily.
```


filanov:
I'm not sure i like all the hacks that i had to do in the makefile but locally it worked for me

qinqon:
> I'm not sure i like all the hacks that i had to do in the makefile but locally it worked for me

Maybe it's more explicit to put a Makefile under `apis", the Makefile is already quite complex.

filanov:
> > I'm not sure i like all the hacks that i had to do in the makefile but locally it worked for me
> 
> Maybe it's more explicit to put a Makefile under `apis", the Makefile is already quite complex.

The problem that i had was around the unit tests, the solution that i found was to run `go list` that will bring all the packages related to the go.mod in the context, so i had to give a list of all the sub modules get the list of packages and only then run the tests. 
I will add explanation to the commit message to make sure that it's clear why i did those things. 
Can you please enable the tests so i at least i will  know that they are passing? 
I will hold the PR until i finish editing the commit messages. 
/hold

filanov:
> 

should it get ok to test?

Makefile
```diff
@@ -161,8 +163,12 @@ push-operator: operator-manager
...
 push: push-handler push-operator
 
-test/unit: $(GO)
-	NODE_NAME=node01 $(GINKGO) $(unit_test_args) $(WHAT)
+.PHONY: $(addprefix test-unit-submodules-,$(WHAT))
+$(addprefix test-unit-submodules-,$(WHAT)):
+	# hande unit test for submodule
+	(cd $(subst test-unit-submodules-,,$@) ; NODE_NAME=node01 $(GINKGO) $(unit_test_args) `$(GO) list -f {{.Dir}} ./...`)
+
+test/unit: $(GO) $(addprefix test-unit-submodules-,$(WHAT))
```
qinqon, line 171: Relaxing a little the WHAT filter we can do 

```Makefile
test/unit/api: $(GO)                                                            
    cd api && $(GINKGO) $(unit_test_args) ./...                                 
                                                                                
test/unit: test/unit/api                                                        
    NODE_NAME=node01 $(GINKGO) $(unit_test_args) $(WHAT) 
```

Makefile
```diff
@@ -201,8 +207,9 @@ release: $(GO)
 	hack/release.sh
 
 vendor: $(GO)
-	$(GO) mod tidy
-	$(GO) mod vendor
+	for f in `find . -name go.mod -not -path '*/vendor/*' -not -path '*/build/*'`; do \
+		cd $$(dirname $$f); $(GO) mod tidy; $(GO) mod vendor; cd $(PWD); \
+	done
```
qinqon, line 212: We are not going to have more sub modules, let's just create a target for api.

```suggestion
vendor: vendor-api                                                              
    $(GO) mod tidy                                                              
    $(GO) mod vendor                                                            
                                                                                
vendor-api: $(GO)                                                               
    cd api && $(GO) mod tidy && $(GO) mod tidy                
```

filanov:
> > I'm not sure i like all the hacks that i had to do in the makefile but locally it worked for me
> 
> Maybe it's more explicit to put a Makefile under `apis", the Makefile is already quite complex.

@qinqon regarding the makefile, there are multiple targets that affect everything and some of them are calling scripts as well, so splitting may be not trivial and even add more complexity.
Maybe it should be done in a separate PR?

qinqon (changes requested):
We are going to have only one sub module, let's harcode a little the Makefile, the proposed solution add too much complexity for an already complex Makefile.

qinqon:
Also the `make lint` has to be fixed doing the following at ./hack/lint.sh

```bash
#!/bin/bash -e                                                                  
golangci_lint_version=v1.42.1                                                   
if [ ! -f $(go env GOPATH)/bin/golangci-lint ]; then                            
    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin $golangci_lint_version
fi                                                                              
golangci-lint run                                                               
(                                                                               
   cd api                                                                       
   golangci-lint run --config ../.golangci.yml                                  
)               
```

Makefile
filanov: The reason i did it this way was because of `pkg/names`, i had to make it a submodule as well because api importing from it, so if it's not a submodule it will create a cyclic dependency. 
Running `NODE_NAME=node01 $(GINKGO) $(unit_test_args) $(WHAT)` on pkg will fail because it will not be able to run the unit tests of `pkg/names`

Makefile
filanov: There are two submodles that i had to create, `api` and  `pkg/names`. 
`pkg/namese` are used by api so unless it's a submodule it will create a cyclic dependency. 
I can create multiple targets, just was thinking about something more generic.

filanov: fixed

Makefile
filanov: I forgot to add pkg/names to `WHAT`, added that as well. unit tests passed locally.

qinqon: Let's move `pkg/names` to `api/names`, having a submodule only for `pkg/names` does not make any sense then we can do the Makefile change:

```Makefile
test/unit/api: $(GO)                                                            
    cd api && $(GINKGO) $(unit_test_args) ./...                                 
                                                                                
test/unit: test/unit/api                                                        
    NODE_NAME=node01 $(GINKGO) $(unit_test_args) $(WHAT) 
```

filanov: I was thinking that it's not related to the `api` package but sure no problem

filanov: fixed

qinqon (approved):

qinqon:
/hold
lint is failing @filanov can you fix it ?

filanov:
> /hold lint is failing @filanov can you fix it ?
> 
> I think it's trying to lint the vendor under ./api and it times out.

Lint failed on timeout, locally it works fine
```
  Running [/home/runner/golangci-lint-1.42.1-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
  
  Error: golangci-lint exit with code 4
  Ran golangci-lint in 81625ms
```

qinqon:
> @qinqon is there an option to re-run it

For github actions usually you can click on re-run, but it may need permissions, let me retry.

phoracek:
Can we close https://github.com/nmstate/kubernetes-nmstate/issues/864 now?

filanov:
> Can we close #864 now?

Yep

PR #912, opened 2021-11-29
Remove leader election

qinqon:

>
> /kind bug
/kind enhancement

The need for leader elections is no longer needed since kubernetes-nmstdate use `Recreate` strategy to upgrade so only one instace is active and also cert-manager is no longer living on the webhook so we we don't need leader election there either.

**Release note**:

```release-note
Remove leader election mechanism from webhook and cert-manager
```


rhrazdil (approved):
/lgtm
/approve

qinqon:
/retest
Affected by https://github.com/nmstate/kubernetes-nmstate/pull/914

PR #917, opened 2021-11-29
Don't wait for node boot when waiting for nncp transition

rhrazdil:
The function waiting for node reboot is checking if ssh connects
every 5 seconds, which seems to be enough for a simple policy to
start reconciling and even finish, meaning that subsequent
check, that waits for a NNCP transition timestamp to be updated
fails (as no more transitions happen).

This change doesn't wait for the node to boot and relies only on
the transition timestamp update.

Also, fixed restartNodeWithoutWaiting which didn't return an error

>
> /kind bug
> /kind enhancement

Fixes flake that occured in https://github.com/nmstate/kubernetes-nmstate/pull/914

**Release note**:

```release-note
NONE
```


qinqon:
@rhrazdil looks like it breaks stuff.

/retest

PR #919, opened 2021-12-01
make: Remove install go infra

qinqon:
Since we have the go.mod "go" directive this is safe enough to ensure
that the golang compiler used has the needed capabilities, this change
remove the infra to use a vendored golang toolchain.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


rhrazdil (approved):
/lgtm
/approve

Issue #927, opened 2021-12-06
DNS tests are blocking CI


rhrazdil:
**What happened**:

DNS tests are failing consistently

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- `NodeNetworkState` on affected nodes (use `kubectl get nodenetworkstate <node_name> -o yaml`):
- Problematic `NodeNetworkConfigurationPolicy`:
- kubernetes-nmstate image (use `kubectl get pods --all-namespaces -l app=kubernetes-nmstate -o jsonpath='{.items[0].spec.containers[0].image}'`):
- NetworkManager version (use `nmcli --version`)
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Others:

PR #929, opened 2021-12-06
Fix issues in DNS tests

rhrazdil:
Addresses https://github.com/nmstate/kubernetes-nmstate/issues/927

Following issues are causing the tests to fail:
1. removing of IP address from an interface was done by using
`absent` withing ipv4 subtree. This is likely a bug in nmstate
as this subtree doesn't specify such property.
2. in static_addr_and_route_test, one IP address was assigned
to all worker nodes. This change designates only one worker
node to have the NNCPs configured. This reduction
doesn't limit our coverage as it shouldn't matter if this configuration
is applied to one node or more.
3. some policies missed ipv4/6.enabled: true, which may not be causing
an issue, but this change specifies the enabled property explicitly.

>
> /kind bug
> /kind enhancement

Fixes https://github.com/nmstate/kubernetes-nmstate/issues/927

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/dns_test.go
+++ b/test/e2e/handler/dns_test.go
@@ -27,6 +27,8 @@ interfaces:
     enabled: true
     dhcp: true
   %s:
+    enabled: true
+    dhcp: true
     auto-dns: false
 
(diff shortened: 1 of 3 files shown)


qinqon (approved):
/lgtm
/approve

PR #934, opened 2021-12-07
Retry when incrementing maxUnavailable

rhrazdil:
/kind bug

Status update in incrementUnavailableNodeCount may fail due to a conflict.
It doesn't look good especially when maxUnavailable is set to 100%, as
seeing enactment in Pending state is not expected.

Addresses BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2029767

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/handler/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/handler/nodenetworkconfigurationpolicy_controller.go
@@ -40,7 +40,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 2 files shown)


controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -185,7 +184,7 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(ctx context.Context
 	if r.shouldIncrementUnavailableNodeCount(instance, previousConditions) {
 		err = r.incrementUnavailableNodeCount(instance)
 		if err != nil {
-			if apierrors.IsConflict(err) {
+			if apierrors.IsConflict(err) || errors.Is(err, node.MaxUnavailableLimitReachedError{}) {
```
qinqon, line 187: we don't need `IsConflict` anymore right ?

qinqon (changes requested):

controllers/handler/nodenetworkconfigurationpolicy_controller.go
rhrazdil: In case we happen to get a conflict in all attempts, we can still get a conflict error, can't we?

qinqon: Right, the conflict error is returned as stated here
https://github.com/kubernetes/client-go/blob/8bffd27f0ac161b10909e7593264cb87a616c410/util/retry/util.go#L62-L64

qinqon (approved):
/lgtm
/approve

PR #935, opened 2021-12-07
webhook: Deny NNCP update if capture is modified

qinqon:

>
/kind bug
> /kind enhancement

Updateing the capture field at NNCP can break scenarios where nodes are
rebooted and NetworkManager configuration is not presistent. This change
prevent that but just disallow update capture field, if it need to be
modified the NNCP has to be re-created.

This is a first step before implementing a better solution.

**Release note**:

```release-note
Deny capture field update at webhook level.
```

### DIFF
--- a/test/e2e/handler/webhook_test.go
+++ b/test/e2e/handler/webhook_test.go
@@ -86,4 +86,20 @@ var _ = Describe("Validation Admission Webhook", func() {
 			Expect(err).To(MatchError("admission webhook \"nodenetworkconfigurationpolicies-create-validate.nmstate.io\" denied the request: failed to admit NodeNetworkConfigurationPolicy this-is-longer-than-sixty-three-characters-hostnames-bar-bar.com: message: invalid policy name: \"this-is-longer-than-sixty-three-characters-hostnames-bar-bar.com\": must be no more than 63 characters. "))
 		})
 	})
+	Context("When a policy capture field is updated", func() {
+		BeforeEach(func() {
+			By("Create a policy without capture field")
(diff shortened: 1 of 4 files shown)


test/e2e/handler/webhook_test.go
```diff
@@ -86,4 +86,20 @@ var _ = Describe("Validation Admission Webhook", func() {
 			Expect(err).To(MatchError("admission webhook \"nodenetworkconfigurationpolicies-create-validate.nmstate.io\" denied the request: failed to admit NodeNetworkConfigurationPolicy this-is-longer-than-sixty-three-characters-hostnames-bar-bar.com: message: invalid policy name: \"this-is-longer-than-sixty-three-characters-hostnames-bar-bar.com\": must be no more than 63 characters. "))
 		})
 	})
+	Context("When a policy is created with capture field", func() {
+		BeforeEach(func() {
+			By("Create a policy")
```
rhrazdil, line 91: Should we add an expect that would check the capture field is indeed set?

qinqon: It doe not create the capture field, we try to create it later on doing a NNCP modification, that's how I check that the webhook is working without adding capture field to TestPolicy, so it play nice with the rest of the tests.

rhrazdil: Ahh, I see. I find the Context a bit misleading then.
Could it be `When a policy capture field is updated`.

rhrazdil:
/lgtm
/approve
Thanks!

PR #936, opened 2021-12-07
Provide a way to skip the image builds

creydr:

>
> /kind bug
/kind enhancement

Sometimes it is not needed (or even not possible) to build the images via `make push` (or `make handler` / `make operator`). E.g. some CI environments use their own builders and don't support calls to podman or docker. This PR addresses it and adds an option to skip the image builds via the `SKIP_IMAGE_BUILD` var. That way the builds via podman or docker can be skipped.

I also fixed the podman build (f3f8f5181a91d0ec4678ce49f8cec6c676f178a1), as it seemed not to respect the `SKIP_PUSH` var as done with docker: https://github.com/nmstate/kubernetes-nmstate/blob/e1eb3ecdb2fe8d570ba4d96f268c43328dde7017/hack/build-push-container.docker.sh#L23-L27

**Release note**:
```release-note
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,8 @@ INDEX_VERSION ?= 1.0.0
 # Default index image tag
 INDEX_IMG ?= $(IMAGE_REGISTRY)/$(IMAGE_REPO)/kubernetes-nmstate-operator-index:$(INDEX_VERSION)
 
+SKIP_IMAGE_BUILD ?= false
+
 all: check handler
 
 check: vet whitespace-check gofmt-check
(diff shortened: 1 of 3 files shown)


qinqon:
You can do the same as this PR just calling make handler-manager or make operator-manager, they will build the binaries.

creydr:
> You can do the same as this PR just calling make handler-manager or make operator-manager, they will build the binaries.

That's true, but this doesn't work, when we want to run the e2e tests and have to setup the cluster before with the `cluster-sync-operator` target, which internally builds the whole operator (what is sometimes not needed in case the operator was build as part of the CI pipeline before):

https://github.com/nmstate/kubernetes-nmstate/blob/e1eb3ecdb2fe8d570ba4d96f268c43328dde7017/cluster/sync-operator.sh#L9-L13

https://github.com/nmstate/kubernetes-nmstate/blob/e1eb3ecdb2fe8d570ba4d96f268c43328dde7017/cluster/sync-common.sh#L56-L78

qinqon:
/lgtm
/approve

Fair enough

PR #943, opened 2021-12-13
nmpolicy e2e: test moving static ip to a bridge using capture

AlonaKaplan:

>
> /kind bug
/kind enhancement

This PR intoducing a test adding a bridge and moving the static networking information (ip and routes) of eth1 to the bridge.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/static_addr_and_route_test.go
+++ b/test/e2e/handler/static_addr_and_route_test.go
@@ -44,6 +44,23 @@ routes:
 `, firstSecondaryNic, firstSecondaryNic))
 }
 
+func ipV4AddrAndRouteAndBridgeAbsent(firstSecondaryNic, bridgeName string) nmstate.State {
+	return nmstate.NewState(fmt.Sprintf(`interfaces:
+  - name: %s
+    type: ethernet
+    state: up
+    ipv4:
+      enabled: false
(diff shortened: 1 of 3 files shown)


qinqon:
@AlonaKaplan can you rebase ? lint is fixed.

test/e2e/handler/static_addr_and_route_test.go
```diff
@@ -151,6 +169,35 @@ var _ = Describe("Static addresses and routes", func() {
 			})
 		})
 
+		Context("with bridge taking over the static networking of the slave", func() {
```
rhrazdil, line 172: You could maybe split the current Context into two contexts, the outer one would set up the static IP and route in BeforeEach and the inner one would  handle the policy with the capture.

Not a must though.

AlonaKaplan: Since the `AfterEach` is removing it all in one `updateDesiredStateAndWait(ipV4AddrAndRouteAndBridgeAbsent(firstSecondaryNic, bridgeName))` I prefer to do the settings in the same `BeforeEach`.

rhrazdil: Fair enough

PR #944, opened 2021-12-13
e2e: Increse time for checking policy is degraded

rhrazdil:
In case conflicts occur, it may take over one second
to update policy status. Increasing to 2 seconds.
We can't increase the timeout too much as then the
test wouldn't test what we need.

It would be better to use a transition timestamp,
but currently, transition timestamp is updated also
when Reason changes, so we lose the time when the
policy actually turned Degraded/Available.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


qinqon:
@rhrazdil can you rebase ? I think lint is fixed

PR #947, opened 2021-12-17
Fix goheader linter

tompsota:
Resolved warnings from goheader linter

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/header.tpl
+++ b/hack/header.tpl
@@ -1,6 +1,5 @@
 Copyright The Kubernetes NMState Authors.
 
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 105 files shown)


hack/header.tpl
```diff
@@ -1,6 +1,5 @@
 Copyright The Kubernetes NMState Authors.
```
phoracek, line 2: You have removed a line from the template, but in all the rendered headers there are two empty lines. Is it on purpose?

phoracek (commented):
/hold

It does not look consistent to me. Feel free to cancel the hold once answered

phoracek:
@tompsota the merge was faster, but would you please see my comment above?

hack/header.tpl
tompsota: Yes, it's on purpose. For some reason the goheader linter fails if there are 2 newlines in the template (most likely a bug). I tried to point this out in the commit message of 6c5d6b91a2d830085c87fde1f9f7e1fc78b407e2 , but maybe I was too brief.

PR #956, opened 2022-01-06
deps: Pin nmstate to 1.0-x

qinqon:
The latest nmstate 1.2.0 has a bug that breaks k8s connectivity [1].
This change pin to a working version

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2037411

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
Pin nmstate to 1.1.0
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -6,7 +6,7 @@ RUN env
 COPY build/_output/bin/handler.manager.linux-$TARGETARCH /usr/local/bin/manager
 
 RUN dnf install -b -y dnf-plugins-core && \
-    dnf install -b -y -x "*alpha*" -x "*beta*" nmstate && \
+    dnf install -b -y -x "*alpha*" -x "*beta*" nmstate-1.0.2 && \
     dnf install -b -y iproute iputils && \
     dnf remove -y dnf-plugins-core && \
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN env
 COPY build/_output/bin/handler.manager.linux-$TARGETARCH /usr/local/bin/manager
 
 RUN dnf install -b -y dnf-plugins-core && \
-    dnf install -b -y -x "*alpha*" -x "*beta*" nmstate && \
+    dnf install -b -y -x "*alpha*" -x "*beta*" nmstate-1.0.2 && \
```
phoracek, line 9: The PR says 1.1 but here I see 1.0. Which one is wrong?

phoracek (changes requested):

build/Dockerfile
qinqon: We have try 1.1.0 and it's failing so we pin at 1.0 but it's failing too.

rhrazdil:
/hold

we shouldn't need this as soon as https://github.com/nmstate/kubernetes-nmstate/pull/958 is merged

PR #957, opened 2022-01-06
Fix periodic re-reconciles of NNCPs with name equal to node

rhrazdil:
Policies that are named the same as worker nodes are getting
periodically re-reconciled.
It seems that the `For(&corev1.Node{})` creates a watch of it's own
that overrides the builder.WithPredicates(onLabelsUpdatedForThisNode)
defined in the subsequent `Watches(...)` statement.

This commit creates the controller manually without using
controller-runtime builder.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
Fix issue with NNCP named the same as nodes geting periodically re-reconciled
```

### DIFF
--- a/controllers/handler/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/handler/nodenetworkconfigurationpolicy_controller.go
@@ -28,8 +28,8 @@ import (
 	"k8s.io/client-go/util/retry"
 
 	ctrl "sigs.k8s.io/controller-runtime"
-	builder "sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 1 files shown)


controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -258,13 +258,19 @@ func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Man
...
-		Complete(r)
+	c, err := controller.New("node_labels", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return errors.Wrap(err, "failed to create node_labels controller to NNCP Reconciler watching node label changes")
+	}
+	err = c.Watch(
+		&source.Kind{Type: &corev1.Node{}},
+		handler.EnqueueRequestsFromMapFunc(allPolicies),
+		onLabelsUpdatedForThisNode,
+	)
```
qinqon, line 269: We don't need `Complete(r)` ?

rhrazdil: We're creating the controller with `controller.New` on line, so we don't need Complete, as complete is used to instantiate the controller from the controller builder

PR #961, opened 2022-01-17
Use nmcli to obtain NetworkManager version

tompsota:
'NetworkManager' command might not be available at nodes, which may lead to failing tests. The 'nmcli' command should not cause this issue.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/handler/node_controller.go
+++ b/controllers/handler/node_controller.go
@@ -19,6 +19,7 @@ package controllers
 
 import (
 	"context"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -117,10 +118,13 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 }
 
 func (r *NodeReconciler) getDependencyVersions() *nmstate.DependencyVersions {
-	handlerNetworkManagerVersion, err := nmstate.ExecuteCommand("NetworkManager", "--version")
+	handlerNetworkManagerVersion, err := nmstate.ExecuteCommand("nmcli", "--version")
 	if err != nil {
 		r.Log.Info("error retrieving handler NetworkManager version: %s", err.Error())
 	}
+	// remove leading characters up to last space
+	split := strings.Split(handlerNetworkManagerVersion, " ")
+	handlerNetworkManagerVersion = split[len(split)-1]
 
 	handlerNmstateVersion, err := nmstate.ExecuteCommand("nmstatectl", "--version")
 	if err != nil {


controllers/handler/node_controller.go
```diff
@@ -117,10 +118,11 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 }
 
 func (r *NodeReconciler) getDependencyVersions() *nmstate.DependencyVersions {
-	handlerNetworkManagerVersion, err := nmstate.ExecuteCommand("NetworkManager", "--version")
+	handlerNetworkManagerVersion, err := nmstate.ExecuteCommand("nmcli", "--version")
 	if err != nil {
 		r.Log.Info("error retrieving handler NetworkManager version: %s", err.Error())
 	}
+	handlerNetworkManagerVersion = strings.TrimPrefix(handlerNetworkManagerVersion, "nmcli tool, version ")
```
rhrazdil, line 125: It may be better to split the output by a space and pick the last element, so we don't depend on the string to stay the same:
 
```go
s := strings.Split(handlerNetworkManagerVersion, " ")
// also check s contains at least 1 element after splitting
handlerNetworkManagerVersion = handlerNetworkManagerVersion[len(s)-1]

```

rhrazdil (changes requested):

controllers/handler/node_controller.go
```diff
@@ -117,10 +118,13 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 }
 
 func (r *NodeReconciler) getDependencyVersions() *nmstate.DependencyVersions {
-	handlerNetworkManagerVersion, err := nmstate.ExecuteCommand("NetworkManager", "--version")
+	handlerNetworkManagerVersion, err := nmstate.ExecuteCommand("nmcli", "--version")
 	if err != nil {
 		r.Log.Info("error retrieving handler NetworkManager version: %s", err.Error())
 	}
+	// remove leading characters up to last space
+	split := strings.Split(handlerNetworkManagerVersion, " ")
```
rhrazdil, line 126: Could you please check for length of the split before indexing it? To make sure that split has at least 1 element, and avoid indexing with negative number (which would panic)

tompsota: If string does not contain separator, Split returns a slice of length 1 whose only element is original string. So this check is not necessary, imo - it will always evaluate to true. Example: [https://go.dev/play/p/wcabqGIWJAr](https://go.dev/play/p/wcabqGIWJAr)

PR #969, opened 2022-01-21
build: Use multistage to build the manager

qinqon:
In the future the project is going to be integrated with the rust
version from nmstate, this integration is going to be done using CGO and
for that we need to install the nmstate rust libraries and headers. This
change transition to multistage so the nmstate-devel library can be
installed at the build stage.

>
> /kind bug
> /kind enhancement

The CI builds amd64 and arm64 version of the containers, with this PR instead of using go crosscompiling it uses buildx qemu image so the arm64 build is slower.

**Release note**:

```release-note
NONE
```


creydr (commented):
Any reasons for keeping the `hack/build-manager.sh` when we build the binaries it in the container now?

qinqon:
> Any reasons for keeping the `hack/build-manager.sh` when we build the binaries it in the Dockerfiles now?

Nope, removing...

PR #972, opened 2022-01-24
operator: Use priority class cluster critical

qinqon:
/kind bug

The operator can be replaced by a pod running at different node. This
change represent that by setting priority class at operator to
`system-cluster-critical`.

**Release note**:

```release-note
Use system-cluster-critical as priority class for the operator
```

### DIFF
--- a/deploy/operator/operator.yaml
+++ b/deploy/operator/operator.yaml
@@ -24,6 +24,7 @@ spec:
         - key: node-role.kubernetes.io/master
           operator: Exists
           effect: NoSchedule
+      priorityClassName: system-cluster-critical
       containers:
         - name: nmstate-operator
           args:


deploy/operator/operator.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         - key: node-role.kubernetes.io/master
           operator: Exists
           effect: NoSchedule
+      priorityClassName: system-cluster-critical
```
rhrazdil, line 27: Should be system-node-critical, right?

rhrazdil (approved):

deploy/operator/operator.yaml
phoracek: Mmmm, I don't remember in detail, but I think that system-critical means "kill all other on this node before me". That is unnecessarily brutal as for the operator it does not matter on which node in the cluster it runs. So cluster-critical would be enough?

phoracek:
/hold

A concern about the level

deploy/operator/operator.yaml
rhrazdil: I'm not certain about the priority level itself. 
`system-cluster-critical` means that operator should not be preempted by user workloads, but can be OOM killed.
`system-node-critical` means that operator should not even get OOM killed.

Currently, virt-handler pods use `system-node-critical`, IIUC, I don't think it makes sense for an operator to have lower priority than the deployed operands, so if we go with `system-cluster-critical`, we should set it on the handlers too

oshoval:
please update the release note

deploy/operator/operator.yaml
qinqon: Looks like virt-operator is using `priorityClassName: kubevirt-cluster-critical`  https://github.com/kubevirt/kubevirt/blob/099be903803556bfe3a85075a7b55f0a711d9ca7/manifests/generated/operator-csv.yaml.in#L879

The class is 
```yaml
apiVersion: scheduling.k8s.io/v1
description: This priority class should be used for KubeVirt core components only.
kind: PriorityClass
metadata:
  name: kubevirt-cluster-critical
value: 1000000000
```
https://github.com/kubevirt/kubevirt/blob/099be903803556bfe3a85075a7b55f0a711d9ca7/manifests/generated/kubevirt-priority-class.yaml

rhrazdil: Totally, now when I realized that the operator may be scheduled elsewhere in the cluster when kiled and still provide the same service, wherease the handlers need to run on each node, it makes perfect sense

qinqon:
/retest pull-kubernetes-nmstate-e2e-operator-k8s

Looks like ARM build for the operator is failing.

PR #981, opened 2022-02-03
Don't involve notReady nodes in NNCP status calculation

rhrazdil:
In big clusters, it's possible that some nodes are not ready, either
due to maintenance or networking issue. This change excludes NotReady
nodes from NNCP Status calculation, so that policy is considered
as Available, when it successfully applies to all Ready nodes.

When NotReady node turns Ready, it re-reconciles and updates the policy
accordingly.

If there already is Available enactment on a node, that becomes NotReady afterwards,
the calculation is not affected, because we only consider enactments for
current `policyGeneration`.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NNCP doesn't turn Degraded when NotReady nodes don't apply enactments
```

### DIFF
--- a/pkg/policyconditions/conditions.go
+++ b/pkg/policyconditions/conditions.go
@@ -146,6 +146,8 @@ func update(apiWriter client.Client, apiReader client.Reader, policyReader clien
 			return errors.Wrap(err, "getting nodes running kubernets-nmstate pods failed")
 		}
 		numberOfNmstateMatchingNodes := len(nmstateMatchingNodes)
+		numberOfReadyNmstateMatchingNodes := len(node.FilterReady(nmstateMatchingNodes))
+		numberOfNotReadyNmstateMatchingNodes := numberOfNmstateMatchingNodes - numberOfReadyNmstateMatchingNodes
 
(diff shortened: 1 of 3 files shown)


pkg/policyconditions/conditions.go
```diff
@@ -155,29 +157,55 @@ func update(apiWriter client.Client, apiReader client.Reader, policyReader clien
...
+				"Policy is progressing %d/%d nodes finished",
+				numberOfFinishedEnactments,
+				numberOfReadyNmstateMatchingNodes,
+			)
+			if numberOfNotReadyNmstateMatchingNodes != 0 {
+				message += fmt.Sprintf(
+					", %d nodes ignored due to NotReady state",
+					numberOfNotReadyNmstateMatchingNodes,
+				)
+			}
```
qinqon, line 192: Can we create a little closure with this before the if statement to reuse things ?

qinqon (changes requested):
nit for code reuse.

pkg/policyconditions/conditions.go
rhrazdil: Thanks, done

PR #982, opened 2022-02-03
Add progressing condition to NNCP

rhrazdil:
Add a new condition to NNCP: Progressing.
Progressing condition is set to true
when at least one enactment is still in progress.

To avoid clearing the Progressing condition by Pending enactments,
NNCP conditions are reset only when it's not progress.

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
NNCP has a new Progressing conditon set to true while enactments are in progress
```

### DIFF
--- a/vendor/github.com/nmstate/kubernetes-nmstate/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/vendor/github.com/nmstate/kubernetes-nmstate/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -42,13 +42,15 @@ type NodeNetworkConfigurationPolicyStatus struct {
 }
 
 const (
-	NodeNetworkConfigurationPolicyConditionAvailable ConditionType = "Available"
-	NodeNetworkConfigurationPolicyConditionDegraded  ConditionType = "Degraded"
+	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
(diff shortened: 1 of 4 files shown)


vendor/github.com/nmstate/kubernetes-nmstate/api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -42,13 +42,15 @@ type NodeNetworkConfigurationPolicyStatus struct {
 }
 
 const (
-	NodeNetworkConfigurationPolicyConditionAvailable ConditionType = "Available"
-	NodeNetworkConfigurationPolicyConditionDegraded  ConditionType = "Degraded"
+	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
```
rhrazdil, line 45: @qinqon  seems odd to change vendor. Is it okay or should we in case of updating the API first update the API, merge the update and when the update is vendored, use it in code?

qinqon: It's not possible to do on the same PR at a pair of commits ?

rhrazdil: Ah I see, thanks. I've splitted it to separate commits

PR #995, opened 2022-02-22
Fix golangci linters

tompsota:
The warnings from `gocyclo` and `funlen` linters are now resolved.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/handler/main.go
+++ b/cmd/handler/main.go
@@ -34,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	// +kubebuilder:scaffold:imports
 
@@ -56,6 +57,8 @@ import (
 	"github.com/nmstate/kubernetes-nmstate/pkg/webhook"
 )
 
+const generalExitStatus int = 1
+
 type ProfilerConfig struct {
 	EnableProfiler bool   `envconfig:"ENABLE_PROFILER"`
 	ProfilerPort   string `envconfig:"PROFILER_PORT" default:"6060"`
@@ -76,8 +79,8 @@ func init() {
 }
 
 func main() {
-	if mainHandler() == 1 {
-		os.Exit(1)
+	if mainHandler() == generalExitStatus {
+		os.Exit(generalExitStatus)
 	}
 }
 
@@ -99,7 +102,6 @@ func mainHandler() int {
 	}
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opt)))
-
 	// Lock only for handler, we can run old and new version of
 	// webhook without problems, policy status will be updated
 	// by multiple instances.
@@ -107,7 +109,7 @@ func mainHandler() int {
 		handlerLock, err := lockHandler()
 		if err != nil {
 			setupLog.Error(err, "Failed to run lockHandler")
-			return 1
+			return generalExitStatus
 		}
 		defer handlerLock.Unlock()
 		setupLog.Info("Successfully took nmstate exclusive lock")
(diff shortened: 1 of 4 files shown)


controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -157,14 +149,39 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(_ context.Context,
...
+) error {
+	err := r.Client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Policy is not found, removing previous enactment if any")
+			err = r.deleteEnactmentForPolicy(request.NamespacedName.Name)
+			return err
+		}
+		log.Error(err, "Error retrieving policy")
+		// Error reading the object - requeue the request.
```
phoracek, line 168: This comment is not relevant in the scope of the helper function - the re-queue happens on the caller side

cmd/handler/main.go
```diff
@@ -100,6 +92,15 @@ func mainHandler() int {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opt)))
 
+	if mainHandler() == 1 {
```
rhrazdil, line 95: Could we use a constant?

controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -122,17 +122,9 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(_ context.Context,
 	_ = context.Background()
 	log := r.Log.WithValues("nodenetworkconfigurationpolicy", request.NamespacedName)
 
-	// Fetch the NodeNetworkConfigurationPolicy instance
 	instance := &nmstatev1.NodeNetworkConfigurationPolicy{}
-	err := r.Client.Get(context.TODO(), request.NamespacedName, instance)
+	err := r.fetchNodeNetworkConfigurationPolicyInstance(instance, &request, &log)
```
rhrazdil, line 126: I'm not sure about the motivation for the function.
Also, in go, it's common practice to not use `get/set/fetch/lookup` prefixes in function names.

controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -157,14 +149,39 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(_ context.Context,
...
+func (r *NodeNetworkConfigurationPolicyReconciler) fetchNodeNetworkConfigurationPolicyInstance(
+	instance *nmstatev1.NodeNetworkConfigurationPolicy,
+	request *ctrl.Request,
+	log *logr.Logger,
+) error {
+	err := r.Client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Policy is not found, removing previous enactment if any")
+			err = r.deleteEnactmentForPolicy(request.NamespacedName.Name)
```
rhrazdil, line 164: I don't like that this function, that according to its name merely "fetches" a NNCP also deletes something else.

controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -157,14 +149,39 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(_ context.Context,
...
+			return err
+		}
+		log.Error(err, "Error retrieving policy")
+		// Error reading the object - requeue the request.
+		return err
+	}
+	return nil
+}
 
+func (r *NodeNetworkConfigurationPolicyReconciler) reconcileEnactmentConditions(
```
rhrazdil, line 174: IMO, it doesn't make sense to call this reconcile*
as in k8s , reconcile is a method of a reconciler, that is called by a controller.

pkg/policyconditions/conditions.go
```diff
@@ -178,75 +177,9 @@ func update(apiWriter client.Client, apiReader client.Reader, policyReader clien
...
-		numberOfFinishedEnactments := enactmentsCountByCondition.Available() +
-			enactmentsCountByCondition.Failed() +
-			enactmentsCountByCondition.Aborted()
-
-		logger.Info(
-			fmt.Sprintf("numberOfNmstateMatchingNodes: %d, enactments count: %s",
-				numberOfNmstateMatchingNodes,
-				enactmentsCountByCondition),
-		)
+		setPolicyStatus(policy, &nmstateMatchingNodes, &enactments, &logger)
```
rhrazdil, line 180: I think that abstracting the status logic to a separate function could be a benefit, but can we limit the function to do only that?
And collect/calculate the node statuses outside of this function

rhrazdil (commented):
The PR is pretty big, it would if it was spli into at least 2 commits, each delating with a different linter or a specific linter fix.

tompsota:
> The PR is pretty big, it would if it was spli into at least 2 commits, each delating with a different linter or a specific linter fix.

Both linters were complaining about approximately the same issues, so we could only split it by files and not by linters, if that's ok

pkg/policyconditions/conditions.go
```diff
@@ -258,6 +201,88 @@ func update(apiWriter client.Client, apiReader client.Reader, policyReader clien
 	})
 }
 
+func setPolicyStatus(
+	policy *nmstatev1.NodeNetworkConfigurationPolicy,
+	nodeStatus *nodeStatusInfo,
+	logger *logr.Logger,
+) {
+	logger.Info(
```
rhrazdil, line 209: I think we could simply log this information before this function is called. Then we don't need the logger argument.

rhrazdil (changes requested):
We need to work a bit on the second commit, I have only minor nits in the third commit.

cmd/handler/main.go
```diff
@@ -254,12 +160,145 @@ func mainHandler() int {
...
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create NodeNetworkConfigurationEnactment controller", "controller", "NMState")
+		return err
+	}
+
+	return nil
+}
+
+func createHealthyFile() error {
+	// Handler runs with host networking so opening ports is problematic
```
rhrazdil, line 228: This comment at this scope doesn't make much sense, it should be before the function call.

rhrazdil (changes requested):
One last nit, looking good otherwise!

rhrazdil (approved):
/lgtm
/approve

PR #997, opened 2022-02-24
cluster: Unpin NetworkManager 1.34

qinqon:

>
/kind bug
> /kind enhancement

The bug [1] that forced kubernetes-nmstate to pin to NM 1.34 has being
fixed, this change use the NM provided at kubevirtci.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2037411

**Release note**:

```release-note
Unping NM from 1.34
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -21,10 +21,10 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
     done
 done
 
-echo 'Installing Open vSwitch and NetworkManager 1.34 on nodes'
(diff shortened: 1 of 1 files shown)


cluster/up.sh
```diff
@@ -24,7 +24,7 @@ done
 echo 'Installing Open vSwitch and NetworkManager 1.34 on nodes'
```
rhrazdil, line 24: Can you also remove the version from the ehco please?

rhrazdil (changes requested):

PR #998, opened 2022-02-24
capture: Calculated captured state from nmstatectl

qinqon:

>
/kind bug
> /kind enhancement

Calculating the captured state with the NNS is problematic since there
are some latencies between nmstatectl set and NNS being updated, so it
could happend that next NNCP is calculated with inaccurate network
configuration. This change read the currentState from nmstatectl to get
accurate data.

Fix bz https://bugzilla.redhat.com/show_bug.cgi?id=2057613

**Release note**:

```release-note
Calculate captured state from nmstatectl show outpout
```


rhrazdil (commented):
The fix looks good.
@qinqon Interesting that we didn't hit this issue in u/s CI, do you think it's worth adding a test scenario that reproduced this in a followup?

/lgtm
/approve

qinqon:
> The fix looks good. @qinqon Interesting that we didn't hit this issue in u/s CI, do you think it's worth adding a test scenario that reproduced this in a followup?

Since it's a race it depends how much time happends between one NNCP and the next, also what do we do on the next NNCP, if both NNCP use capture and one capture is using the previous as input then we can end with the issue at hand.

rhrazdil:
>Since it's a race it depends how much time happends between one NNCP and the next, also what do we do on the next NNCP, if both NNCP use capture and one capture is using the previous as input then we can end with the issue at hand.

Thanks, understood.

qinqon:
/retest

It suffers from DNS resolve ping, let's retry 

```
: failed checking DNS connectivity: [failed looking up NS root-server.net using name sever 192.168.66.2: lookup root-server.net on [fd00::1]:53: server misbehaving failed looking up NS root-server.net using name sever fd00::1: lookup root-server.net on [fd00::1]:53: dial udp [fd00::1]:53: connect: network is unreachable]: failed running probes after rollback: failed runnig probe 'dns' with after network reconfiguration -> currentState: ---
```

rhrazdil:
This time, the go build failed again

qinqon:
> This time, the go build failed again /test pull-kubernetes-nmstate-e2e-handler-k8s

ARM build sometimes fail...

PR #1006, opened 2022-03-11
Ensure NNCP controllers Reconcile is not run in parallel

creydr:

>
/kind bug
> /kind enhancement

We declare two NNCP controllers (one for changes on NNCP and one for changes on nodes (node labels)):
https://github.com/nmstate/kubernetes-nmstate/blob/1fff555f0a3cbd96769dc6ebb9a2fa427dab92b0/controllers/handler/nodenetworkconfigurationpolicy_controller.go#L251-L269

So there is a change that Reconcile runs in parallel. This can cause issues if the same policy is getting reconciled in parallel.

This PR addresses it and ensures only one Reconcile is run at a time.

**Special notes for reviewers:**
There is also an [MaxConcurrentReconciles](https://github.com/kubernetes-sigs/controller-runtime/blob/9ee63fc65a9720568f74df9901b16883621636b4/pkg/controller/controller.go#L39) option, but this can only be declared on one controller and not across them.
Also I checked if we can merge the declaration of the two controllers, but didn't find a way for it, so I ended up with using a mutex.

**Release note**:
```release-note
none
```

### DIFF
--- a/controllers/handler/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/handler/nodenetworkconfigurationpolicy_controller.go
@@ -247,21 +247,25 @@ func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Man
 			return allPoliciesAsRequest
 		})
 
-	// Reconcile NNCP if they are created or updated
-	err := ctrl.NewControllerManagedBy(mgr).
-		For(&nmstatev1.NodeNetworkConfigurationPolicy{}).
-		WithEventFilter(onCreateOrUpdateWithDifferentGenerationOrDelete).
-		Complete(r)
+	// Reconcile NNCP if they are created/updated/deleted or
+	// Node is updated (for example labels are changed), node creation event
+	// is not needed since all NNCPs are going to be Reconcile at node startup.
+	c, err := controller.New("NodeNetworkConfigurationPolicy", mgr, controller.Options{Reconciler: r})
 	if err != nil {
-		return errors.Wrap(err, "failed to add controller to NNCP Reconciler listening NNCP events")
+		return errors.Wrap(err, "failed to create NodeNetworkConfigurationPolicy controller")
 	}
 
-	// Reconcile all NNCPs if Node is updated (for example labels are changed), node creation event
-	// is not needed since all NNCPs are going to be Reconcile at node startup.
-	c, err := controller.New("node_labels", mgr, controller.Options{Reconciler: r})
+	// Add watch for NNCP
+	err = c.Watch(
+		&source.Kind{Type: &nmstatev1.NodeNetworkConfigurationPolicy{}},
+		&handler.EnqueueRequestForObject{},
+		onCreateOrUpdateWithDifferentGenerationOrDelete,
+	)
 	if err != nil {
-		return errors.Wrap(err, "failed to create node_labels controller to NNCP Reconciler watching node label changes")
+		return errors.Wrap(err, "failed to add watch for NNCPs")
 	}
+
+	// Add watch to enque all NNCPs on nod label changes
 	err = c.Watch(
 		&source.Kind{Type: &corev1.Node{}},
 		handler.EnqueueRequestsFromMapFunc(allPolicies),


qinqon:
/hold

Want to recheck that there is no way of unify controllers.

rhrazdil:
Nice find.
IIUC, we could add a list of watches using c.Watch: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L122

Couldn't we add a new watch by
```go
err = c.Watch(
		&source.Kind{Type: &nmstatev1.NodeNetworkConfigurationPolicy{}},
		&handler.EnqueueRequestForObject{},
		onCreateOrUpdateWithDifferentGenerationOrDelete,
	)
	if err != nil {
		return errors.Wrap(err, "failed to add watch to enqueue NNCPs reconcile on create or update with different generation or delete")
	}
```
And remove the controller create with `ctrl.NewControllerManagedBy(mgr)`?

creydr:
> Nice find. IIUC, we could add a list of watches using c.Watch: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L122
> 
> Couldn't we add a new watch by
> 
> ```go
> err = c.Watch(
> 	&source.Kind{Type: &nmstatev1.NodeNetworkConfigurationPolicy{}},
> 	&handler.EnqueueRequestForObject{},
> 	onCreateOrUpdateWithDifferentGenerationOrDelete,
> )
> if err != nil {
> 	return errors.Wrap(err, "failed to add watch to enqueue NNCPs reconcile on create or update with different generation or delete")
> }
> ```
> 
> And remove the controller create with `ctrl.NewControllerManagedBy(mgr)`?

Ah didn't know we can concate watches.
So we could maybe even use something like
```
err := ctrl.NewControllerManagedBy(mgr).
	Watches(
		&source.Kind{Type: &nmstatev1.NodeNetworkConfigurationPolicy{}},
		&handler.EnqueueRequestForObject{},
		builder.WithPredicates(onCreateOrUpdateWithDifferentGenerationOrDelete),
	).
	Watches(
		&source.Kind{Type: &corev1.Node{}},
		handler.EnqueueRequestsFromMapFunc(allPolicies),
		builder.WithPredicates(onLabelsUpdatedForThisNode),
	).
	Complete(r)
```

qinqon:
> Nice find. IIUC, we could add a list of watches using c.Watch: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L122
> 
> Couldn't we add a new watch by
> 
> ```go
> err = c.Watch(
> 	&source.Kind{Type: &nmstatev1.NodeNetworkConfigurationPolicy{}},
> 	&handler.EnqueueRequestForObject{},
> 	onCreateOrUpdateWithDifferentGenerationOrDelete,
> )
> if err != nil {
> 	return errors.Wrap(err, "failed to add watch to enqueue NNCPs reconcile on create or update with different generation or delete")
> }
> ```
> 
> And remove the controller create with `ctrl.NewControllerManagedBy(mgr)`?

Yep, the `For` is just syntatic suggar we should use the low level "watch" and call it multiple times, checking at the code they store a slice of watchers https://github.com/kubernetes-sigs/controller-runtime/blob/0a3dd2a36dd971ea2fdcb163b0457e6be0d315bd/pkg/builder/controller.go#L132

@creydr can you try this out ?

creydr:
> @creydr can you try this out ?

sure. Updated the PR and waiting for the tests :)

rhrazdil:
>Ah didn't know we can concate watches.
So we could maybe even use something like
>```
>err := ctrl.NewControllerManagedBy(mgr).
>	Watches(
>		&source.Kind{Type: &nmstatev1.NodeNetworkConfigurationPolicy{}},
>		&handler.EnqueueRequestForObject{},
>		builder.WithPredicates(onCreateOrUpdateWithDifferentGenerationOrDelete),
>	).
>	Watches(
>		&source.Kind{Type: &corev1.Node{}},
>		handler.EnqueueRequestsFromMapFunc(allPolicies),
>		builder.WithPredicates(onLabelsUpdatedForThisNode),
>	).
>	Complete(r)
>```

I think the problem with NewControllerManagedBy is that it mandates usage of `For`

PR #1009, opened 2022-03-14
build: Use go cross compiling instead of qemu

qinqon:
The buildx build command start a QEMU instance if the container arch is
not the one on the host, go toolchain has cross compiling functionality
so qemu is not needed for the compiling stuff. This change force the
compile stage to be amd64 since it's the arch at CI and then runs cross
compiling to compile amd64 or arm64 so it does not compile over qemu.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


build/Dockerfile
```diff
@@ -2,11 +2,15 @@ FROM --platform=linux/amd64 quay.io/centos/centos:stream8 AS build
...
-RUN dnf install -y golang-1.17.*
+RUN mkdir /workdir
 
-COPY . /workdir
+COPY go.mod /workdir
 WORKDIR /workdir
 
+RUN dnf install -y golang-$(sed -En 's/^go +(.*+)$/\1/p' go.mod).*
+
+COPY . .
```
rhrazdil, line 12: What is this magic?

qinqon: We have already change to /workdir so it's not needed to be specified

rhrazdil: Ah of course

rhrazdil (changes requested):
Could you please update also the build/Dockerfile.future?
Or do you wish to merge this one and update future in a followup?

rhrazdil:
/lgtm
Thank you!

PR #1012, opened 2022-03-17
webhook: Distribute webhook at control-plane nodes

qinqon:
/kind bug

At multiple node control-plane webhook was allow to run replicas on the
same node. This change introduces a topologySpreadConstraints directive
at webhook pod to distribute them.

Closes https://bugzilla.redhat.com/show_bug.cgi?id=2056464

**Release note**:

```release-note
Distribute kubernetes-nmstate-webhook pods across control-plane nodes.
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -28,6 +28,13 @@ spec:
       nodeSelector: {{ toYaml .InfraNodeSelector | nindent 8 }}
       tolerations: {{ toYaml .InfraTolerations | nindent 8 }}
       affinity: {{ toYaml .WebhookAffinity | nindent 8 }}
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: kubernetes.io/hostname
+        whenUnsatisfiable: DoNotSchedule
+        labelSelector:
+          matchLabels:
+            component: kubernetes-nmstate-webhook
       priorityClassName: system-cluster-critical
       containers:
         - name: nmstate-webhook
--- a/test/e2e/operator/main_test.go
+++ b/test/e2e/operator/main_test.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
@@ -76,7 +77,6 @@ func TestE2E(t *testing.T) {
 	testenv.TestMain()
 
 	RegisterFailHandler(Fail)
-
 	RunSpecs(t, "Operator E2E Test Suite")
 }
 
@@ -91,8 +91,7 @@ var _ = BeforeSuite(func() {
 
 	By("Getting node list from cluster")
 	nodeList := corev1.NodeList{}
-	err := testenv.Client.List(context.TODO(), &nodeList, &client.ListOptions{})
-	Expect(err).ToNot(HaveOccurred())
+	Expect(testenv.Client.List(context.TODO(), &nodeList, &client.ListOptions{})).To(Succeed())
 	for _, node := range nodeList.Items {
 		nodes = append(nodes, node.Name)
 	}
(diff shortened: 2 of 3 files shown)


deploy/handler/operator.yaml
```diff
@@ -28,6 +28,13 @@ spec:
       nodeSelector: {{ toYaml .InfraNodeSelector | nindent 8 }}
       tolerations: {{ toYaml .InfraTolerations | nindent 8 }}
       affinity: {{ toYaml .WebhookAffinity | nindent 8 }}
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: kubernetes.io/hostname
```
rhrazdil, line 33: Shouldn't the topologyKey be `node-role.kubernetes.io/control-plane`?
Otherwise, I don't understand what is the purpose of adding second master node

test/e2e/operator/main_test.go
```diff
@@ -154,6 +154,21 @@ func eventuallyOperandIsReady(testData operatorTestData) {
...
+
+func shouldBeDistributedPodsWith(listOptions ...client.ListOption) {
+	podList := &corev1.PodList{}
+	Expect(testenv.Client.List(context.TODO(), podList, listOptions...)).To(Succeed())
+	if len(nodes) > 1 {
+		nodesRunningPod := map[string]bool{}
+		for _, pod := range podList.Items {
+			nodesRunningPod[pod.Spec.NodeName] = true
+		}
+		Expect(nodesRunningPod).To(HaveLen(len(podList.Items)), "should run pods at different nodes")
```
rhrazdil, line 170: :+1:

deploy/handler/operator.yaml
rhrazdil: Ah, nope, now I get the `topologySpreadConstraints` behaviour.
Zoning by hostname makes sense, since we don't really know apriori other labels that would distinguish nodes.

So, I don't understand why node2 is labeled as master then.

qinqon: Topology key is the value that mark pods uniquedness, I have mark the node02 as master to be able to do e2e test checking that webhook is deployed at both.

qinqon:
/hold
Let's add a test for single node control-plane

qinqon:
/hold cancel

Now the operator first e2e tests checks with 1 and 2 control plane sizes

test/e2e/operator/nmstate_install_test.go
```diff
@@ -37,14 +38,43 @@ import (
 )
 
 var _ = Describe("NMState operator", func() {
+	DescribeTable("for control-plane size",
+		func(controlPlaneSize int) {
+			if isKubevirtciCluster() {
+				for i := 2; i <= controlPlaneSize; i++ {
```
rhrazdil, line 44: Do you think It would be clearer for a reader if the indexing was from 0 with a local func `nodeNameWIthOffset(num int) string`, that would add `2` ?

I think it would be easier to realize how many times this code loops.
Another option, since we only currently loop once when the PlaneSize is 2, a simple `if` may also be enough, but if you want to keep it general, it's fine

test/e2e/operator/main_test.go
```diff
@@ -77,7 +78,27 @@ func TestE2E(t *testing.T) {
 
 	RegisterFailHandler(Fail)
 
+	/*
+		if isKubevirtciCluster() {
```
rhrazdil, line 82: Should this be removed or uncommented?

rhrazdil (changes requested):

test/e2e/operator/main_test.go
qinqon: Removing

test/e2e/operator/nmstate_install_test.go
qinqon: I will refactor it a little, it's true that it's not very nice for readers

qinqon: It cannot start from 0 since node01 is already part of the control-plane and we don't want to re-label it.

qinqon: I can generalize this completly looking for not control plane and marking them as such

rhrazdil (approved):
Looks nice, thanks
/lgtm
/approve

PR #1018, opened 2022-03-21
Update NNCE LTT only on state change

rhrazdil:
/kind bug

Update NNCe Last Transition Time only when state
of a condition is changed.

**Release note**:

```release-note
update Last Transition Time only when condition state actually changes
```


vendor/github.com/nmstate/kubernetes-nmstate/api/shared/condition_types.go
```diff
@@ -50,10 +50,12 @@ func (conditions *ConditionList) Set(conditionType ConditionType, status corev1.
 
 	// If there is different status, reason or message update it
 	if condition.Status != status || condition.Reason != reason || condition.Message != message {
+		if condition.Status != status {
```
phoracek, line 53: We should not apply patches to vendor/, no?

rhrazdil: We are vendoring the API as a subpackage, so we need to sync changes done in API

phoracek: I see, thanks

qinqon (approved):
/lgtm
/approve

PR #1028, opened 2022-03-28
Nmstate add selfsigned api

rhrazdil:
/kind enhancement

This PR adds API for configuring SelfSigned parameters for cert rotation.

**Release note**:
```release-note
added API to NMState CRD for configuring SelfSigned parameters for cert rotation.
```


qinqon (approved):
/lgtm

/cc @creydr can you take a look ?

controllers/operator/nmstate_controller.go
```diff
@@ -210,6 +210,16 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1.NMState) error {
...
 	}
 
+	selfSignConfiguration := instance.Spec.SelfSignConfiguration
+	if selfSignConfiguration == nil {
+		selfSignConfiguration = &nmstatev1.SelfSignConfiguration{
+			CARotateInterval:    "8760h0m0s",
+			CAOverlapInterval:   "24h0m0s",
+			CertRotateInterval:  "4380h0m0s",
+			CertOverlapInterval: "24h0m0s",
+		}
```
creydr, line 220: just a nit: maybe we could use kubebuilders `// +kubebuilder:default:=xyz` on the type, so this would be evidently in the CRD (https://book.kubebuilder.io/reference/markers/crd-validation.html)

rhrazdil: Hmm, I've moved the defaulting part to the operator code to make it more obvious, what defaults are set, as this is where other default valuse are also set.

The kubebuilder option isn't optimal for this use case, as the whole struct is nullable

PR #1030, opened 2022-04-04
Remove mount options from build files

creydr:

>
> /kind bug
> /kind enhancement

Removes the `mount` option from `RUN` in the Dockerfiles, as in Podman this is only supported since v4 and v4 is not that popular that much yet.

Should be reverted when Podman v4 is out for a little longer (#1031)

**Release note**:
```release-note
none
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -12,7 +12,7 @@ RUN dnf install -y golang-$(sed -En 's/^go +(.*+)$/\1/p' go.mod).*
 
 COPY . .
 
-RUN --mount=type=cache,target=/root/.cache/go-build GOOS=linux GOARCH=${TARGETARCH} go build -o /manager ./cmd/handler
+RUN GOOS=linux GOARCH=${TARGETARCH} go build -o /manager ./cmd/handler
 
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN dnf install -y golang-$(sed -En 's/^go +(.*+)$/\1/p' go.mod).*
 
 COPY . .
 
-RUN --mount=type=cache,target=/root/.cache/go-build GOOS=linux GOARCH=${TARGETARCH} go build -o /manager ./cmd/handler
+RUN GOOS=linux GOARCH=${TARGETARCH} go build -o /manager ./cmd/handler
```
qinqon, line 15: Can we put this version at Dockerfile.podman ? the cache is quite convenient since it will not compile the vendoring everytime you change a line of code.

qinqon (changes requested):
Let's use for now a diferent Dockerfile for podman without the cache option.

creydr:
Created #1031 to track the revert of this.

qinqon (approved):
/lgtm
/approve
/retest

PR #1033, opened 2022-04-04
Update Actions Go version to match the project

phoracek:

 /kind bug

The Go version listed in Actions config is not matching the version used by the project. This patch fixes the mismatch.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,7 +1,5 @@
 name: checks
 on: [push, pull_request]
-env:
-  GO_VERSION: 1.16
 jobs:
   lint:
     name: lint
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -1,7 +1,7 @@
 name: checks
 on: [push, pull_request]
 env:
-  GO_VERSION: 1.16
+  GO_VERSION: 1.17
```
qinqon, line 4: Can you use this PR to integrate with https://github.com/marketplace/actions/go-version-action ?

qinqon (changes requested):

.github/workflows/check.yaml
phoracek: Nice! I updated the config, let's see if it works

PR #1036, opened 2022-04-06
cluster: Use kubectl rollout to deploy

qinqon:
The correct way to check for deployment finished is using kubectl
rollout since it's the samething for deployments and daemonsets.

```release-note
NONE
```

### DIFF
--- a/cluster/sync-operator.sh
+++ b/cluster/sync-operator.sh
@@ -23,14 +23,8 @@ function deploy_operator() {
 
 function wait_ready_operator() {
     # We have to re-check desired number, sometimes takes some time to be filled in
-    if ! eventually isDeploymentOk ${OPERATOR_NAMESPACE} app=kubernetes-nmstate-operator; then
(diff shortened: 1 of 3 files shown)


cluster/sync-operator.sh
```diff
@@ -23,16 +23,10 @@ function deploy_operator() {
 
 function wait_ready_operator() {
     # We have to re-check desired number, sometimes takes some time to be filled in
-    if ! eventually isDeploymentOk ${OPERATOR_NAMESPACE} app=kubernetes-nmstate-operator; then
+    if ! $kubectl rollout status -w -n ${OPERATOR_NAMESPACE} kubernetes-nmstate-operator; then
```
rhrazdil, line 26: shouldn't this be 
```suggestion
    if ! $kubectl rollout status -w -n ${OPERATOR_NAMESPACE} deployment kubernetes-nmstate-operator; then
```

rhrazdil (changes requested):

rhrazdil (approved):
/lgtm
/approve

PR #1045, opened 2022-04-08
build: Use copr nmstate/nmstate-git for future

qinqon:
Using just nmstate from stream9 is not new enough since we are going to
transition soon to stream9 at main, this change uses copr
nmstate/nmstate-git instead wich are RPMs build after a PR is merge at
nmstate base branch.

**Release note**:

```release-note
NONE
```


rhrazdil (approved):
/approve

PR #1052, opened 2022-04-14
rbac: Add missing cluster-reader verbs

oshoval:
When using Openshift, in case cluster-reader
ClusterRole exists,
Users of cluster-readers group should be able to
access k8s nmstate crds, such as nns, nncp and nnce.

Add a ClusterRole that allows it in case
cluster-reader exists.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2022745

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/operator/nmstate_controller.go
+++ b/controllers/operator/nmstate_controller.go
@@ -27,7 +27,9 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -153,6 +155,11 @@ func (r *NMStateReconciler) applyRBAC(instance *nmstatev1.NMState) error {
 	data.Data["HandlerImage"] = os.Getenv("RELATED_IMAGE_HANDLER_IMAGE")
 	data.Data["HandlerPullPolicy"] = os.Getenv("HANDLER_IMAGE_PULL_POLICY")
 	data.Data["HandlerPrefix"] = os.Getenv("HANDLER_PREFIX")
+
+	if err := setClusterReaderExist(r.Client, data); err != nil {
+		return errors.Wrap(err, "failed checking if cluster-reader ClusterRole exists")
+	}
+
 	return r.renderAndApply(instance, data, "rbac", true)
 }
 
@@ -276,5 +283,22 @@ func (r *NMStateReconciler) renderAndApply(
 			return errors.Wrapf(err, "failed to apply object %v", obj)
 		}
 	}
+	return nil
+}
+
+func setClusterReaderExist(c client.Client, data render.RenderData) error {
+	var clusterReader rbac.ClusterRole
+	key := types.NamespacedName{Name: "cluster-reader"}
+	err := c.Get(context.TODO(), key, &clusterReader)
+
+	found := true
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		found = false
+	}
+
+	data.Data["ClusterReaderExists"] = found
 	return nil
 }
--- a/deploy/handler/cluster_role.yaml
+++ b/deploy/handler/cluster_role.yaml
@@ -1,0 +1,19 @@
+{{if .ClusterReaderExists}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nmstate-cluster-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+rules:
+- apiGroups:
+  - nmstate.io
+  resources:
+  - nodenetworkstates
+  - nodenetworkconfigurationpolicies
+  - nodenetworkconfigurationenactments
+  verbs:
+  - get
+  - list
+  - watch
+{{end}}
--- a/test/e2e/operator/nmstate_install_test.go
+++ b/test/e2e/operator/nmstate_install_test.go
@@ -26,6 +26,7 @@ import (
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -130,6 +131,54 @@ var _ = Describe("NMState operator", func() {
 			})
 		})
 	})
+	Context("when cluser-reader exists", func() {
+		const (
+			clusterReaderRoleName = "cluster-reader"
+			testUserNamespace     = "default"
+			serviceAccountName    = "test-user"
+			testUserBindingName   = "test-user-binding"
+		)
+
+		var clusterReaderCreated bool
+
+		BeforeEach(func() {
+			err := createClusterReaderCR(clusterReaderRoleName)
+			Expect(err).To(SatisfyAny(Succeed(), WithTransform(apierrors.IsAlreadyExists, BeTrue())))
+			if err == nil {
+				clusterReaderCreated = true
+			}
+
+			Expect(createTestUserSA(testUserNamespace, serviceAccountName)).To(Succeed(),
+				"should success creating a serviceaccount")
+			Expect(createTestUserCRB(testUserBindingName, testUserNamespace, serviceAccountName, clusterReaderRoleName)).To(Succeed(),
+				"should success creating a clusterrolebinding")
+
+			By("Install NMState for the first time")
+			installNMState(defaultOperator.nmstate)
+			eventuallyOperandIsReady(defaultOperator)
+		})
+		AfterEach(func() {
+			uninstallNMStateAndWaitForDeletion(defaultOperator)
+		})
+		AfterEach(func() {
+			Expect(deleteTestUserCRB(testUserBindingName)).To(Succeed())
+		})
+		AfterEach(func() {
+			Expect(deleteTestUserSA(testUserNamespace, serviceAccountName)).To(Succeed())
+		})
+		AfterEach(func() {
+			if clusterReaderCreated {
+				Expect(deleteClusterReaderCR(clusterReaderRoleName)).To(Succeed())
+			}
+		})
+
+		It("should be able to read NMState resources with cluster-reader user", func() {
+			Eventually(func() error {
+				_, err := cmd.Kubectl("get", "nns", fmt.Sprintf("--as=system:serviceaccount:%s:%s", testUserNamespace, serviceAccountName))
+				return err
+			}, 10*time.Second, time.Second).Should(Succeed())
+		})
+	})
 })
 
 func installOperator(operator operatorTestData) {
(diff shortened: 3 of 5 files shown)


oshoval:
pretty messy, don't think it is due to this PR
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/1052/pull-kubernetes-nmstate-e2e-handler-k8s/1514604352960991232

qinqon:
/hold
This should fixed at kubernetes-nmstate project CNAO main branch no longer install it.

controllers/operator/nmstate_controller.go
```diff
@@ -153,6 +155,12 @@ func (r *NMStateReconciler) applyRBAC(instance *nmstatev1.NMState) error {
 	data.Data["HandlerImage"] = os.Getenv("RELATED_IMAGE_HANDLER_IMAGE")
 	data.Data["HandlerPullPolicy"] = os.Getenv("HANDLER_IMAGE_PULL_POLICY")
 	data.Data["HandlerPrefix"] = os.Getenv("HANDLER_PREFIX")
+
+	err := setClusterReaderExist(r.Client, data)
+	if err != nil {
+		return errors.Wrap(err, "failed checking if cluster-reader ClusterRole exists")
+	}
```
qinqon, line 162:
```suggestion
	if err := setClusterReaderExist(r.Client, data); err != nil {
		return errors.Wrap(err, "failed checking if cluster-reader ClusterRole exists")
	}
```

deploy/handler/cluster_role.yaml
```diff
@@ -0,0 +1,19 @@
+{{if .ClusterReaderExists}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: k8s-nmstate-project
```
qinqon, line 5: The ` is` implicit and `project`  means nothing.
```suggestion
  name: nmstate-cluster-reader
```

qinqon (changes requested):
Can we add a e2e tests u/s simulating openshit behaviour ? would be good test to run also by openshift operator owners.

oshoval:
> Can we add a e2e tests u/s simulating openshit behaviour ? would be good test to run also by openshift operator owners.

Removed the file from U/S it was only for checking it locally,
in order to check the feature we need openshift where we have the cluster-reader role
There we will need to do the following

1. Create a service account
2. bind the SA to the cluster-reader group
3. login as this SA
4. get nns

The PR creates the new ClusterRole, and since it has the special label 
`rbac.authorization.k8s.io/aggregate-to-cluster-reader`
Openshift will create a ClusterRoleBinding for us that binds between the new ClusterRole and the cluster-reader

beside that - addressed comments, thanks

oshoval:
rebased

oshoval:
Thanks Quique, will ask Openshift Operator people to review it

deploy/handler/cluster_role.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: nmstate-cluster-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+rules:
+- apiGroups:
+  - nmstate.io
+  resources:
+  - nodenetworkstates
+  - nodenetworkconfigurationpolicies
+  - nodenetworkconfigurationenactments
```
creydr, line 14: Adding `nmstates` (`nmstates.nmstate.io`) to this role could make sense too so we have all the nmstate.io resources, or WDYT?

controllers/operator/nmstate_controller.go
```diff
@@ -153,6 +155,11 @@ func (r *NMStateReconciler) applyRBAC(instance *nmstatev1.NMState) error {
 	data.Data["HandlerImage"] = os.Getenv("RELATED_IMAGE_HANDLER_IMAGE")
 	data.Data["HandlerPullPolicy"] = os.Getenv("HANDLER_IMAGE_PULL_POLICY")
 	data.Data["HandlerPrefix"] = os.Getenv("HANDLER_PREFIX")
+
+	if err := setClusterReaderExist(r.Client, data); err != nil {
```
creydr, line 159: IMHO it would be cleaner to have this function only return if the cluster-reader-role exists (and an error eventually) and do the `data.Data["ClusterReaderExists"] = clusterReaderRoleExists` here, so we have it aligned with the other template settings. But that might be a personal opinion...

oshoval: Up to you, I started by that, but then it needs to check the error and do the assign, it looked less good imho
this way it is more compact, even that it is a bit less symmetric

deploy/handler/cluster_role.yaml
oshoval: It seems that we can already read it anyhow by the cluster-reader
so do we still want to add it ? (it would be duplicate, unless we remove it from somewhere else that allows it)

oshoval:
The failures doesn't seem related

deploy/handler/cluster_role.yaml
creydr: hmm. ok. ack

oshoval:
Added test that we can read NNS with SA that is bound to cluster-reader

cluster/cluster_reader.yaml
```diff
@@ -0,0 +1,28 @@
+---
```
qinqon, line 1: This has to be under "test" not under cluster.

cluster/up.sh
```diff
@@ -28,3 +28,5 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
     ./cluster/cli.sh ssh ${node} -- sudo systemctl restart openvswitch
     ./cluster/cli.sh ssh ${node} -- sudo systemctl restart NetworkManager
 done
+
+./cluster/kubectl.sh create -f ./cluster/cluster_reader.yaml
```
qinqon, line 32: This is part of the BearEach at tests

test/e2e/handler/nodes_test.go
```diff
@@ -33,6 +34,12 @@ var _ = Describe("Nodes", func() {
 				interfacesNameForNodeEventually(node).Should(ContainElement(primaryNic))
 			}
 		})
+		It("should able to read NMState resources with cluster-reader user", func() {
+			Eventually(func() error {
+				_, err := cmd.Kubectl("get", "nns", "--as=system:serviceaccount:default:test-user")
```
qinqon, line 39: You don't need to call kubectl, you can use the controller-runtime client instancianting it with Impersonate options https://pkg.go.dev/k8s.io/client-go/rest#Config

test/e2e/handler/nodes_test.go
```diff
@@ -33,6 +34,12 @@ var _ = Describe("Nodes", func() {
...
+				return err
+			}, 5*time.Second, 1*time.Second).Should(Succeed())
```
qinqon, line 41:
```suggestion
			}, 5*time.Second, time.Second).Should(Succeed())
```

test/e2e/handler/nodes_test.go
```diff
@@ -33,6 +34,12 @@ var _ = Describe("Nodes", func() {
 				interfacesNameForNodeEventually(node).Should(ContainElement(primaryNic))
 			}
 		})
+		It("should able to read NMState resources with cluster-reader user", func() {
```
qinqon, line 37:
```suggestion
		It("should be able to read NMState resources with cluster-reader role user", func() {
```

deploy/handler/cluster_role.yaml
```diff
@@ -0,0 +1,19 @@
...
+  verbs:
+  - get
+  - list
+  - watch
+{{end}}
```
qinqon, line 19: What do you think about removing this from the "role.yaml" and create another role binding to bind this role to the nmstate-handler user ? this way we don't need the conditional since we must deploy it.

oshoval: We need to allow the cluster-reader to read those, not other user, so it won't help
and imo self contained like this
unless I miss anything ?

cluster/up.sh
oshoval: The cluster-reader creation part of it must be part of the cluster-up before we cluster-sync,
else the reconcile won't create the cluster-role

beside that having a test-user doesn't depend on the `cluster-sync` and might be helpful for other uses as well
as it simulates OCP env that have cluster-reader
and doesn't need to recreated again and again imo

cluster/cluster_reader.yaml
oshoval: due to the reasons mentioned at https://github.com/nmstate/kubernetes-nmstate/pull/1052#discussion_r859554382
i think it is more suitable as part of cluster, as it prefer the cluster (independent of cluster-sync and even should be before that)

test/e2e/handler/nodes_test.go
oshoval: As we spoke, it would require another client configured as impersonate in order to keep the kube:admin client intact,
and doesn't seem to worth it, we can recalculate later if needed

oshoval:
```
+ ./cluster/kubectl.sh rollout status -w -n nmstate ds nmstate-handler
selecting docker as container runtime
Error from server (NotFound): daemonsets.apps "nmstate-handler" not found
+ echo 'Handler haven'\''t turned ready within the given timeout'
Handler haven't turned ready within the given timeout
+ return 1
make: *** [Makefile:175: cluster-sync] Error 1
```
The failures here happen not only on this PR, need to extend the timeout as after few seconds it is fine if you check the cluster manually

oshoval:
addressed comments

oshoval:
/retest 

rollout timeouts - trying to address at https://github.com/nmstate/kubernetes-nmstate/pull/1069

test/e2e/operator/nmstate_install_test.go
```diff
@@ -130,6 +131,40 @@ var _ = Describe("NMState operator", func() {
 			})
 		})
 	})
+	Context("when cluser-reader exists", func() {
+		const (
+			clusterReaderRoleName = "cluster-reader"
+			testUserNamespace     = "default"
```
oshoval, line 137: maybe better on the nmstate NS, so once it is deleted it will clean tests leftovers if any

oshoval:
/test pull-kubernetes-nmstate-e2e-handler-k8s-future 

PR to fix the timeout was merged

oshoval:
wiill return the Eventually, because it take a moment until there is NNS

test/e2e/operator/nmstate_install_test.go
```diff
@@ -130,6 +131,41 @@ var _ = Describe("NMState operator", func() {
...
+			serviceAccountName    = "test-user"
+			testUserBindingName   = "test-user-binding"
+		)
+
+		BeforeEach(func() {
+			Expect(createClusterReaderCR(clusterReaderRoleName)).To(Succeed(),
```
creydr, line 143: This probably fails on testing on Openshift as the cluster role exists already. But you could check in case of an error is it already existed (`errors.IsAlreadyExists(err)`)

test/e2e/operator/nmstate_install_test.go
```diff
@@ -130,6 +131,41 @@ var _ = Describe("NMState operator", func() {
...
+			By("Install NMState for the first time")
+			installNMState(defaultOperator.nmstate)
+			eventuallyOperandIsReady(defaultOperator)
+		})
+		AfterEach(func() {
+			uninstallNMStateAndWaitForDeletion(defaultOperator)
+
+			Expect(deleteTestUserCRB(testUserBindingName)).To(Succeed())
+			Expect(deleteTestUserSA(testUserNamespace, serviceAccountName)).To(Succeed())
+			Expect(deleteClusterReaderCR(clusterReaderRoleName)).To(Succeed())
```
creydr, line 159: only delete, if **you** created it

oshoval: all of them ?
We are safe to do so only for the `deleteClusterReaderCR` right ?
since as you said above i will check if its not already exists, and if it exists i won't create nor delete
The two others we know I created because if they are already there, it would fail on the creation.

In case someone interrupts after creation and before deletion, it is better that we will clean those on the run so the next run will be clean.
Their name is unique (has test in it), so it is unlikely someone else created them.

Another point Is that we can remove split those 3 lines to 3 AfterEach so if one fails, it will still perform the others.

wdyt ?

oshoval: > To be more unique, you could also use the `metadata.GenerateName` and let k8s decide on its "final" name.

I prefer deterministic, unless you think its necessary

oshoval:
Thanks, addressed comments
Also split to separated `AfterEach` so in case one fails, it will still clean the rest

test/e2e/operator/nmstate_install_test.go
```diff
@@ -130,6 +131,54 @@ var _ = Describe("NMState operator", func() {
...
+			Expect(deleteTestUserCRB(testUserBindingName)).To(Succeed())
+		})
+		AfterEach(func() {
+			Expect(deleteTestUserSA(testUserNamespace, serviceAccountName)).To(Succeed())
+		})
+		AfterEach(func() {
+			if clusterReaderCreated {
+				Expect(deleteClusterReaderCR(clusterReaderRoleName)).To(Succeed())
+			}
+		})
```
qinqon, line 173: why Three `AfterEach` ?

qinqon (changes requested):

test/e2e/operator/nmstate_install_test.go
oshoval: so if one fails, it will still try to clean the others

PR #1057, opened 2022-04-20
examples, vlan: specify mandatory vlan.id and vlan.base-iface

rhrazdil:
Without specifying vlan.base-iface and vlan.id, NNCP configuration
fails with:
VLAN tunnel eth1.102 has missing mandatory property

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

**Is this a BUG FIX or a FEATURE ?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind enhancement

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #1058, opened 2022-04-21
e2e,examples: fix e2e examples

rhrazdil:
Currently, examples test only the very last example yaml, which is
'worker selector'. The reason is Ginkgo's multistep flow, which
1. collects all the container nodes (Context, Describe, When) and stores
the runtime blocks (the actual test functions passed to It specs) for
later execution.
2. executes those blocks

This means that in our case, the containers have correct names,
generated during the loop, but the functions all hold reference
to the last example, so all specs work with worker-selector.yaml.

Fixing this by storing to a local variable.

This change also includes fixes to the examples, that are not working
properly. The DNS test in particular requires proper cleanup.
This change adds cleanupState field to the exampleSpec to perform custom
cleanup at AfterEach stage.

We may add a separate example for cleaning the DNS setting, but we still need to clean the DNS configuration in the AfterEach, otherwise we hit issues with the outer-most AfterEach, that attempts to restore interfaces - because we can't remove the interface if DNS configuration is bound to it (it currently requires specific interface, either dynamic with auto-dns: false of static with static gateway), so the cleanupState is still required.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/examples_test.go
+++ b/test/e2e/handler/examples_test.go
@@ -21,13 +21,16 @@ import (
 	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
+
+	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
 )
 
 type exampleSpec struct {
-	name       string
-	fileName   string
-	policyName string
-	ifaceNames []string
+	name         string
+	fileName     string
+	policyName   string
+	ifaceNames   []string
+	cleanupState *nmstate.State
 }
 
 // This suite checks that all examples in our docs can be successfully applied.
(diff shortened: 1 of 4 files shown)


rhrazdil:
@phoracek This explains why the `vlan` example test was passing in CI before
https://github.com/nmstate/kubernetes-nmstate/pull/1057#event-6464791432 was merged

test/e2e/handler/examples_test.go
```diff
@@ -21,20 +21,32 @@ import (
...
 var _ = Describe("[user-guide] Examples", func() {
+	cleanDNSDesiredState := nmstate.NewState(`dns-resolver:
+    config:
+      search: []
+      server: []
+interfaces:
+- name: eth1
+  state: absent
+`)
+
```
qinqon, line 49: Make sense to have an example with this ?

rhrazdil: >We may add a separate example for cleaning the DNS setting, but we still need to clean the DNS configuration in the AfterEach, otherwise we hit issues with the outer-most AfterEach, that attempts to restore interfaces - because we can't remove the interface if DNS configuration is bound to it (it currently requires specific interface, either dynamic with auto-dns: false of static with static gateway), so the cleanupState is still required.

PR #1062, opened 2022-04-22
Set operator replica & minReplica numbers depending on cluster topology

creydr:

>
/kind bug
> /kind enhancement

Currently the replica count and min. replicas of the PodDisruptionBudget for the nmstate controller are hard coded:
https://github.com/nmstate/kubernetes-nmstate/blob/a9bb5da2a198fbe8114b9b64405c4ed693b0dce8/controllers/operator/nmstate_controller.go#L193-L196. 

This can lead to upgrade problems in case the cluster has only one node (e.g. on SNO, Minikube, ...). This PR addresses it and sets these values depending on the cluster topology.

**Release note**:
```release-note
none
```


controllers/operator/nmstate_controller.go
```diff
@@ -237,6 +239,34 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1.NMState) error {
...
+
+		singleNodeClusterReplicaCountDesired = 1
+		singleNodeClusterReplicaMinCount     = 0
+	)
+
+	nodes := corev1.NodeList{}
+	err := r.Client.List(context.TODO(), &nodes, client.MatchingLabels(nodeSelector))
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to get nodes: %w", err)
+	}
```
qinqon, line 261: The webhook is deployed only on the control-plane so this has to count number of control plane nodes excluding workers.

qinqon (changes requested):

controllers/operator/nmstate_controller.go
creydr: So does the `nodeSelector` label match to worker nodes too?
https://github.com/nmstate/kubernetes-nmstate/blob/847d80edc8ee3828c52caad39d4fd05f88200b7b/controllers/operator/nmstate_controller.go#L194-L201

qinqon:
/override pull-kubernetes-nmstate-e2e-handler-k8s-future

Looks like rust nmstatectl is the default now at stream9 and it's not ready yet.

controllers/operator/nmstate_controller.go
```diff
@@ -237,6 +239,34 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1.NMState) error {
...
+// 1. min. number of replicas
+// 2. number of desired replicas
+// 3. error
+func (r *NMStateReconciler) webhookReplicaCount(nodeSelector map[string]string) (int, int, error) { //nolint:gocritic
+	const (
+		multiNodeClusterReplicaCountDesired = 2
+		multiNodeClusterReplicaMinCount     = 1
+
+		singleNodeClusterReplicaCountDesired = 1
+		singleNodeClusterReplicaMinCount     = 0
```
qinqon, line 254: Without webhook installed NNCP cannot be applied since MutatingWebhookConfiguration is deployed.

qinqon (changes requested):

PR #1071, opened 2022-04-28
test, reset: Remove reset NNCP if it fail

qinqon:

>
/kind bug
> /kind enhancement

If the reset nics NNCP at e2e test fails the NNCP is not deleted so the
next test that try to reset it will apply it but it will take no efect.
This change delete the NNCP with defer so is always deleted.

**Release note**:

```release-note
NONE
```


rhrazdil (approved):
/approve

PR #1072, opened 2022-04-29
cluster: Add sleep before checking rollout

qinqon:
Running rollout sometimes fails if the resources has not being already
created. This change add a little sleep to ensure that, doing an
eventually on creation is not worth it.

```release-note
NONE
```

### DIFF
--- a/cluster/sync-operator.sh
+++ b/cluster/sync-operator.sh
@@ -22,7 +22,10 @@ function deploy_operator() {
 }
 
 function wait_ready_operator() {
-    # We have to re-check desired number, sometimes takes some time to be filled in
+    # Wait a little for resources to be created
+    sleep 5
+
+    # Wait for deployment rollout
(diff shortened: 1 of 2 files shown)


rhrazdil (approved):
/approve

rhrazdil:
the tests timed out
/retest

cluster/sync-operator.sh
```diff
@@ -22,7 +22,10 @@ function deploy_operator() {
 }
 
 function wait_ready_operator() {
-    # We have to re-check desired number, sometimes takes some time to be filled in
+    # Wait a little for resources to be created
+    sleep 5
```
oshoval, line 26: Thanks

Maybe instead of sleep we can check when the resources are Ready?

rhrazdil: That's what we do, but when the resource doesn't exist yet, wait for condition fails because the resource is not found.

PR #1075, opened 2022-04-29
test: Wait for terminal NNCP status

qinqon:
If a NNCP is marked as Degraded the test will continue at eventually.
This change finish the eventually look if a terminal state (Degraded or
Available) is rechead and then check the expected status.

**Release note**:

```release-note
NONE
```


rhrazdil (approved):
/approve

PR #1079, opened 2022-05-06
Let Ginkgo recover on fail in runAndWait()

creydr:

>
/kind bug
> /kind enhancement

When there is an error in one of the goroutines in `runAndWait()` (e.g. in `logDeviceStatus` because it can't be ssh'ed to the node), the tests fail with
```
...
  When you, or your assertion library, calls Ginkgo's Fail(),
  Ginkgo panics to prevent subsequent assertions from running.
  Normally Ginkgo rescues this panic so you shouldn't see it.
  However, if you make an assertion in a goroutine, Ginkgo can't capture the
  panic.
  To circumvent this, you should call
  	defer GinkgoRecover()
  at the top of the goroutine that caused this panic.
```
This PR addresses it and adds the ginkgo.GinkgoRecover() to the go routines.

**Release note**:
```release-note
none
```

### DIFF
--- a/test/reporter/reporter.go
+++ b/test/reporter/reporter.go
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/ginkgo/v2/types"
 )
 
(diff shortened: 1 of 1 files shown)


qinqon:
/lgtm
/approve
Thanks!

PR #1083, opened 2022-05-11
api: Remove vendor dir

qinqon:
/kind bug

The api golang lib is not released anywhere so vendor directory is not
necessary. Also `go mod tidy` at top directory was missing so the
vedored api version was old.

**Release note**:

```release-note
Remove api vendor directory and use the latest api version at root project.
```


rhrazdil (approved):
/approve

PR #1086, opened 2022-05-18
infra: Don't use vendor for tools

qinqon:

>
> /kind bug
> /kind enhancement

The vendor directory should concern only the product tools cannot
pollute it.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -97,7 +97,7 @@ check: lint vet whitespace-check gofmt-check
 format: whitespace-format gofmt
 
 vet:
-	go vet ./...
+	GOFLAGS=-mod=mod go vet ./...
```
rhrazdil, line 100: do we need this flag on line 134 in manifests target as well?

Makefile
```diff
@@ -60,9 +60,9 @@ export KUBECTL ?= ./cluster/kubectl.sh
 KUBECTL ?= ./cluster/kubectl.sh
 OPERATOR_SDK ?= $(GOBIN)/operator-sdk
 
-GINKGO = go run github.com/onsi/ginkgo/v2/ginkgo
-CONTROLLER_GEN = go run sigs.k8s.io/controller-tools/cmd/controller-gen
-OPM = go run -tags=json1 github.com/operator-framework/operator-registry/cmd/opm
+GINKGO = GOFLAGS=-mod=mod go run github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
+CONTROLLER_GEN = GOFLAGS=-mod=mod go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
```
rhrazdil, line 64: Could to use the same version of controller-gen as before? Because the newer version removes crd:trivialVersions .
To make this change more clear, I think it would be better to use here the version from go.mod, and only then bump it, which will requre removing the trivialVersions param from makefile

rhrazdil (commented):
Could you please split `go mod tidy/vendor` changes to a separate commit?  The size of the diff makes it hard to even scroll over the changes

api/Makefile
```diff
@@ -1,4 +1,4 @@
-CONTROLLER_GEN = go run sigs.k8s.io/controller-tools/cmd/controller-gen
+CONTROLLER_GEN = GOFLAGS=-mod=mod go run sigs.k8s.io/controller-tools/cmd/controller-gen
```
rhrazdil, line 1:
```suggestion
CONTROLLER_GEN = GOFLAGS=-mod=mod go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.0
```

Makefile
```diff
@@ -60,9 +60,9 @@ export KUBECTL ?= ./cluster/kubectl.sh
 KUBECTL ?= ./cluster/kubectl.sh
 OPERATOR_SDK ?= $(GOBIN)/operator-sdk
 
-GINKGO = go run github.com/onsi/ginkgo/v2/ginkgo
-CONTROLLER_GEN = go run sigs.k8s.io/controller-tools/cmd/controller-gen
-OPM = go run -tags=json1 github.com/operator-framework/operator-registry/cmd/opm
+GINKGO = GOFLAGS=-mod=mod go run github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
+CONTROLLER_GEN = GOFLAGS=-mod=mod go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.0
```
rhrazdil, line 64: Could we have something like 

`GO_MOD_FLAGS = GOFLAGS=-mod=mod `
GINKGO = go run github.com/onsi/ginkgo/v2/ginkgo@v2.1.4

And then use it for example like 
```
vet:
	$(GO_MOD_FLAGS) go vet ./...

test-e2e-upgrade: manifests
	KUBECONFIG=$(KUBECONFIG) OPERATOR_NAMESPACE=$(OPERATOR_NAMESPACE) $(GO_MOD_FLAGS) GINKGO="$(GINKGO)" ./hack/run-e2e-test-upgrade.sh $(e2e_test_args) $(E2E_TEST_SUITE_ARGS)
```

Because currently, for test-e2e-upgrade target, the GOFLAGS=-mod=mod is interpreted as command name

qinqon: Used different approach

rhrazdil (approved):
Thanks!
/approve

Issue #725, opened 2021-04-15
RFE: Configure static IPs on a pool of nodes with a single Policy


nabbas-ca:
**What happened**:
When trying to set a static ip to an extra interface (in my case, non-routable ip, for internal network), i had to create a different policy for every node. Basically, using a node selector for each node, and have a policy per IP to be assigned. This translates to multiple nncp objects (order of x), and so many nnce objects (order of x^2). All of this while having exactly the same configuration, except for the IP.
**What you expected to happen**:
I would like to have a way to have only 1 policy , for multiple IPs on all the nodes. Maybe using an IP pool, or a formula, or any other way. This way, there is no need to duplicate configuration yamls, and also, makes nnce objects more scalable.
**How to reproduce it (as minimally and precisely as possible)**:
sample nncp yaml:

```yaml
apiVersion: nmstate.io/v1beta1
kind: NodeNetworkConfigurationPolicy
metadata:
  name: bond1-policy
spec:
nodeSelector: 
    kubernetes.io/hostname: compute-0.k8s.com 
  desiredState:
    interfaces:
    - name: bond1 
      description: Bond enslaving ens3f0 and ens3f1 
      type: bond 
      state: up 
      ipv4:
        address:
        - ip: 169.100.100.15
          prefix-length: 24
        dhcp: false 
        enabled: true 
      link-aggregation:
        mode: 802.3ad 
        options:
          miimon: '140' 
        slaves: 
        - ens3f0
        - ens3f1
      mtu: 1500 
```

- Create multiple yamls like above, for different nodes and different IPs, with same configuration. apply all of them
- oc get nncp , you should get x number of objects
- oc get nnce , you should get x^2 objects, where only x are successfully configured, the rest are mismatched because of the selector.

**Anything else we need to know?**:

**Environment**:
This is within Openshift 4.7 on baremetal environment.

phoracek:
Hello. This is indeed a valid use-case we've been looking into for a while. The API should be ready for this, this is one of the reasons we have `desiredState` in each individual NodeNetworkConfigurationEnactment - anticipating they will differ.

That being said, with exception of some API drafts, there was not work done on this front yet. Despite we anticipate this as a real use-case there were not many user requests for this.

It was discussed whether this API should be a part of core nmstate, but the answer is negative https://bugzilla.redhat.com/show_bug.cgi?id=1934514. This new API is, unlike the rest of nmstate, stateful.

[...]

PR #1088, opened 2022-05-23
proposal: Host IP pool with NNCP

qinqon:
Add a design proposal to include an IPAM mechanism at kubernetes-nmstate and IPI baremetal.

**Release note**:

```release-note
NONE
```


design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,75 @@
...
+### Non-Goals
+
+- This design does not cover IP address configuration of management NIC on day-0/1
+
+## Proposal
+## Implementation History
+
+- Day 2 ippool with kubernetes-nmstate is implemented.
+- The ippool library at kubernetes-nmstate is extracted or make it public.
+- Baremetal IPI is implemented adding `networkPolicy` field at install config.
```
maiqueb, line 42: I can't make any sense of this proposal with just this info.

maiqueb (changes requested):
I need more detail to understand what your proposal is.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Summary
+
+Provided a mechanism to configure automatically static IP configuration for secondary interfaces on existing and future Nodes
+
+## Motivation
+
+Using DHCP as the IP assignment mechanism has some drawbacks:
+- Introduce latencies on node restart.
+- Has problematic behavious when the DHCP server is miss configured.
```
rhrazdil, line 11: Well, isn't everything problematic, when misconfigured?

```suggestion
- Has problematic behaviours when the DHCP server is misconfigured.
```

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,75 @@
...
+- It depends on network setup that can be complex.
+
+To overcome this static configured interfaces taking the address from a
```
rhrazdil, line 14:
```suggestion
To overcome the above-mentioned drawbacks, statically configured interfaces taking the address from a
```

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,75 @@
...
+dynamic host IP pool is a solution. Since the interfaces are static there is no
+reconfiguration after some time or at node restart.
+
+This functionality was requested on U/S via https://github.com/nmstate/kubernetes-nmstate/issues/725
+
+Spike using whereabouts https://github.com/nmstate/kubernetes-nmstate/pull/1081
```
rhrazdil, line 20: I understand that this effort is just a spike, or poc, but the solution there seems to be in conflict with
this comment from the U/S issue: https://github.com/nmstate/kubernetes-nmstate/issues/725#issuecomment-1021906027

Shouldn't the desing of making this knmstate feature agnostic to the used IPAM plugin be part of the proposal?

rhrazdil (commented):
A few comments.
One concern about making the knmstate depend on whereabouts.

The Goals, User Stories and Non Goals are defined clearly, but I'm not sure I understand the Implementation History sections, that could be more verbose.

design-proposals/host-ip-pool.md
qinqon: I can remove the PoC reference, is just added to have a little of background, but maybe is not adding much.

qinqon:
@rhrazdil @maiqueb Looks like after a bad [force-push ](https://github.com/nmstate/kubernetes-nmstate/compare/bf1873194e1f1aa48642f014f2c98de7bcc1e033..e67958c0d39ae9e873f153edcf6857b2fd43723f)the whole proposal was removed, that's why the whol doc does not make any sense.

design-proposals/host-ip-pool.md
qinqon: Re-worked it

qinqon:
> I need more detail to understand what your proposal is.

I deleted the "proposal" chapter by mistake, https://github.com/nmstate/kubernetes-nmstate/pull/1088#issuecomment-1138261569

design-proposals/host-ip-pool.md
rhrazdil: I think the poc reference can remain here.

> The part that is going to be agnostic is the storage backend more than the IPAM itself.

This would be nice to mention somewhere, unless it already is in the proposal below that I'm just about to read.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+metadata:
+  name: traffic-1
+spec:
+  range:
+    cidr: 10.10.10.0/24
+    start: 1
+    end: 10
+    exclude: [4, 7]
+status:
+  allocated-offsets: [1, 2, 5, 6, 8]
```
maiqueb, line 109: Don't you need to know who owns each of these offsets ? How can you re-sync your pool ? How will you understand if an address is stale/live if not ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+    address-from-traffic-2: ippool.traffic-2.allocate
+desiredState:
+    interfaces:
+    - name: eth1
+      type: ethernet
+      state: up
+      ipv4:
+        address:
+        - ip: "{{ capture.address-from-traffic-1.ip }}"
+          prefix-length: "{{ capture.address-from-traffic-2.prefix-length }}"
```
maiqueb, line 188: I'm aware this is just an example, but I don't think this example make much sense - i.e. getting the IP address from a pool, but the prefix length from another.

The capture looks good.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+        dhcp: false
+        enabled: true
+    - name: eth2
+      type: ethernet
+      state: up
+      ipv4:
+        address:
+        - ip: "{{ capture.address-from-traffic-1.ip }}"
+          prefix-length: "{{ capture.address-from-traffic-2.prefix-length }}"
```
maiqueb, line 197: I think your `eth1` should get address / prefix length from pool 1, and `eth2` get address / prefix length from pool 2 .

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+The new field `platform.baremetal.networkPolicy` will contain the
+content of the day 2 `NodeNetworkConfigurationPolicy` and also the `ipPools`
+since there is not going to be multiple `networkPolicy` fields.
+
+### Implementation Details/Notes/Constraints
+
+#### Day 2
+
+Since kubernetes-nmstate has a distributed architecture each of the handler
+pods will calculate an allacatable ip from the `NodeNetworkIPPool` status and
```
maiqueb, line 271:
```suggestion
pods will calculate an allocatable ip from the `NodeNetworkIPPool` status and
```

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+
+The `image-customize-controller` [1] is a golang controller so it would be quite
+easy to import ippool library from kubernetes-nmstate and nmpolicy to render the
+`networkPolicy`.
+
+[1] https://github.com/openshift/image-customization-controller/blob/3ad18911d14dde1c23b660f99bb05e12a21c99e4/pkg/ignition/builder.go#L79
+
+### Risks and Mitigations
+
+Overtime from unexpected failure there are risk of having some IPs allocated
```
maiqueb, line 289: Do you mean downtime ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+but not used making the number of available IPs smaller and reaching the end
+sooner than expected, for that reseting the `status.allocated-offsets` should
+reconstruct the pool and bring back those IPs.
```
maiqueb, line 292: Is it too obvious to mention that the smaller the pool, the higher the chance of hitting this condition ? 

Consider adding something like "This risk gets higher for shorter IPPools" .

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Drawbacks
+
+The baremetal IPI Nmstate do a offline NetworkManager keyfile generation so
```
maiqueb, line 296:
```suggestion
The baremetal IPI Nmstate does a offline NetworkManager keyfile generation so
```

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+there is no network `currentState` that nmpolicy can reference to, so using
+nmpolicy just to reference the ippools may be too overkill.
```
maiqueb, line 298: Might be just me, but I honestly don't understand this one - what does the keyfile generation have to do with the network current state ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+- Glue `networkPolicy` field with IPAM + nmpolicy at [image-customization-controller](https://github.com/openshift/image-customization-controller)
+- Test baremetal IPI
+
+## Alternatives
+
+An alternative is to implement directly an IPAM feature at Nmstate with a 
+plugin system for ippool retrieval and allocation storage so at day 2 
+it will use the `NodeNetworkIPPool` and `NodeNetworkConfigurationEnactment` and 
+at baremetal IPI It will read the ippool from the openshift installer config 
+and store the allocations at the nodes file system.
```
maiqueb, line 324: This means each node would be pre-allocated with its own IP range slice, and would manage it, right ? 

Wouldn't that complicate the scale up / down scenarios ? AFAIU, it would impact each node's already existing range.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+**cluster admin** is a human user responsible for managing the cluster node
+networking.
+
+1. The cluster admin find that some IPs are not used anymore but still allocated
+2. The cluster admin reset `NodeNetworkIPPool` allocated status
+4. The cluster admin wait for `NodeNetworkIPPool` status to success
+
+NOTE: This is a manual garbage collector that is good enough for first implementation it can be converted into automatic in the future.
+
+### API Extensions day 2
```
maiqueb, line 90: I think we also need to see the enactment objects to better understand the proposal. At least I think I do.

maiqueb (commented):
Could you add some examples of how the enactment objects would look like ? 

Also elaborate a bit on the enactment / IPPool association ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
+# Host IP pooling with NNCPs.
+
+## Summary
+
+Provide a mechanism to assign static IP configuration for secondary interfaces on existing and future cluster nodes
```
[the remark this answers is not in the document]
qinqon: Replacing it with "scale up" and "scale down".

design-proposals/host-ip-pool.md
qinqon: Keeping the spike and rewording it.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+dynamic host IP pool is a solution. Since the interfaces are static there is no
+reconfiguration after some time or at node restart.
+
+This functionality was requested on U/S via https://github.com/nmstate/kubernetes-nmstate/issues/725
+
+Spike using whereabouts https://github.com/nmstate/kubernetes-nmstate/pull/1081
+
+### User Stories
+
+- As a cluster administrator, I want to configure static IPs on secondary NICs on a pool of nodes without creating one per each node.
```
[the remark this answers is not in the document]
qinqon: The feature is only for day 2 and secondary interfaces, this is a good first starting point, we can exapand it further to transform primary interfaces from DHCP to static or even use this same mechanism out of kubernetes-nmstate, but for now this design document is for secondary interfaces.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Proposal
+
+### Workflow Description (assign static address with IPPool)
+
+**cluster admin** is a human user responsible for managing the cluster node
+networking.
+
+1. The cluster admin create `NodeNetworkIPPool` resources with IP ranges and a names.
+2. The cluster admin configure node network interfaces statically but with IP from the `NodeNetworkIPPool` using a
```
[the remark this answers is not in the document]
qinqon: Done

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+`NodeNetworkConfigurationPolicy` referencing them.
+3. The cluster admin wait for the `NodeNetworkConfigurationPolicy` to success
+
+#### Variation (not free IP address at the NodeNetworkIPPool)
```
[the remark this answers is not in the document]
qinqon: Done

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+### Workflow Description (manual NodeNetworkIPPool clean up)
+
+**cluster admin** is a human user responsible for managing the cluster node
+networking.
+
+1. The cluster admin find that some IPs are not used anymore but still allocated
+2. The cluster admin reset `NodeNetworkIPPool` allocated status
+4. The cluster admin wait for `NodeNetworkIPPool` status to success
+
+NOTE: This is a manual garbage collector that is good enough for first implementation it can be converted into automatic in the future.
```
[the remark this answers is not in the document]
qinqon: There is an implemntation history, but it has no dates, I am not sure dates fits on a design document.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+
+### API Extensions day 2
+
+This proposal add a new CRD to configure and store IP pools (NodeNetworkIPPool) and some
+changes at the NNCPs to reference the ippool.
+
+Following is an example CR for the cluster wide CRD `NodeNetworkIPPool`:
```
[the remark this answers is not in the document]
qinqon: Maybe it's better to have `ClusterIPPool` ? it's a ip

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+apiVersion: nmstate.io/v1
+kind: NodeNetworkIPPool
+metadata:
+  name: traffic-1
+spec:
+  range:
+    cidr: 10.10.10.0/24
+    start: 1
+    end: 10
+    exclude: [4, 7]
```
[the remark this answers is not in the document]
qinqon: The whole index thing could be a premature optimization since here we are talking about a maginuted of hundreds so using the the whole IP address even for allocations is not an issue ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+status:
+  allocated-offsets: [1, 2, 5, 6, 8]
+  conditions:
+  - type: Available
+    status: True
+    reason: ReconciledPool
+    message: The pool is up to date
+  - type: Failure
+    status: False
+    reason: ReconciledPool
```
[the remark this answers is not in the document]
qinqon: This are not events this are conditions, and they denote a final state, a user can wait for a Failure condition and different user can wait for a Available condition:
```bash
kubectl wait --for=condition=Available
kubectl wait --for=condition=Failure
```

In this case we are saying that Failure is false since the pool is reconciled.

design-proposals/host-ip-pool.md
qinqon: Added the `NodeNetworkConfigurationEnactment` status API to the doc, now is more clear where is the reference.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: OK, fair enough - please add then the pod default network to the non-goals section. 

I would also introduction an `objectives` (or goals) section, after the motivation.

My mental model for these things is: 
- motivation
- objectives / goals
- non goals
- challenges (if any :) )
- implementation

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Workflow Description (assign static address with IPPool)
+
+**cluster admin** is a human user responsible for managing the cluster node
+networking.
+
+1. The cluster admin create `NodeNetworkIPPool` resources with IP ranges and a names.
+2. The cluster admin configure node network interfaces statically but with IP from the `NodeNetworkIPPool` using a
+`NodeNetworkConfigurationPolicy` referencing them.
+3. The cluster admin wait for the `NodeNetworkConfigurationPolicy` to success
```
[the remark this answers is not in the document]
maiqueb: I still see the wrong sentence .

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+2. The cluster admin wait for the `NodeNetworkConfigurationPolicy` to success so the IPs from removed node are free.
+
+### Workflow Description (manual NodeNetworkIPPool clean up)
+
+**cluster admin** is a human user responsible for managing the cluster node
+networking.
+
+1. The cluster admin find that some IPs are not used anymore but still allocated
+2. The cluster admin reset `NodeNetworkIPPool` allocated status
+4. The cluster admin wait for `NodeNetworkIPPool` status to success
```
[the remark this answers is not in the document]
maiqueb: For what it's worth, this is properly described later on. The user just deletes the indexes from the pool, and the reconcile loop will reconstruct it.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: I don't mean dates ; just feature blocks. I guess what you have covers this. Does it make sense to add "automating the garbage collector" as an optional last item on that list ?

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: would will you exclude ranges using indexes ?

design-proposals/host-ip-pool.md
maiqueb: Cool, I'll take a look later this week. Thanks.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: +1.  I don't think I understand what you're trying to say.

qinqon: Yep, that's it.

design-proposals/host-ip-pool.md
qinqon: I want to state that nmpolicy is designed to work in conjuntion with `currentState` so snippets from it can be inserted into the `desiredState`. But at baremetal IPI there is no `currentState` so if user try to put a desiredState with captures other than ones related to ippool they will not work.

design-proposals/host-ip-pool.md
qinqon: No, it means that at barmetal IPI the information about allocated IPs per node is store at the filesystem and that this may be converted after starting up into a NNCP that will create the NNCEs but re-using the allocation.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+### User Stories
+
+- As a cluster administrator, I want to configure static IPs on secondary NICs on a pool of nodes without creating one per each node.
+- As a cluster administrator, I want scale up the cluster with new nodes getting IPs from the policy.
+- As a cluster administrator, I want each of these static IPs to be unique, and taken from a defined range, and freed when unused.
+
+### Goals
+
+- Allow users to use NodeNetworkConfigurationPolicy to configure static cluster-unique IP addresses without hard-coding them.
+- If at all possible, we should pick a solution that could be utilized even in the IPI nmstate API.
```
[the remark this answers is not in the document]
qinqon: For baremetal IPI installation the openshift installer has a api that allow to pass Nmstate configuration that will be rendered into NetworkManager keyfiles. So it's really the IPI installer config networkData API.

https://github.com/openshift/installer/blob/969378dc498336d1ac29baa89f75324a44467d5d/pkg/types/baremetal/platform.go#L43

design-proposals/host-ip-pool.md
qinqon: Done

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+
+1. The cluster admin remove a node form a cluster already configured using `NodeNetworkIPool`
+2. The cluster admin wait for the `NodeNetworkConfigurationPolicy` to success so the IPs from removed node are free.
+
+### Workflow Description (manual NodeNetworkIPPool clean up)
+
+**cluster admin** is a human user responsible for managing the cluster node
+networking.
+
+1. The cluster admin find that some IPs are not used anymore but still allocated
```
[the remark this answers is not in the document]
qinqon: Rewording on a more explicit manner.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,349 @@
...
+2. The cluster admin reset `NodeNetworkIPPool` allocated status
```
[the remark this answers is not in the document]
qinqon: That's good point, we can Reconcile on status but we have to be carefull not to enter an infinite loop since the status change to change the ippool will also update the status conditions. 

Maybe an annotation is just fine ? At the end this is an emergency situation and in the future it maybe automatize.

What do you think ?

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: Reworded.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,368 @@
...
+
+### API Extensions baremetal IPI
+
+At baremetal IPI the Nmstate configuration is passed to openshift installer
+using the baremetal host `networkConfig` field [1].
+
+[1] https://github.com/openshift/installer/blob/ca8aacd3c931eefabfddafb49cb44b76f24653a6/docs/user/metal/install_ipi.md#describing-hosts
+
+To accommodate the ippools, capture and nmstate new fields will be needed there
+the following is an example of the install-config.yaml from openshift with
```
creydr, line 228: Wouldn't this then be day-0/1?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,368 @@
...
+
+1. The cluster admin creates `NodeIPPool` resources with IP ranges and a name.
+2. The cluster admin configures node network interfaces statically, with IPs from the `NodeIPPool` using a
+`NodeNetworkConfigurationPolicy` referencing them.
+3. The cluster admin wait for the `NodeNetworkConfigurationPolicy` to success
+
+#### Variation (exhausted range at the NodeIPPool)
+
+When the IP pool does not have an available IP address, the `NNCP` will fail 
+with a explicit message, so the range has to be
```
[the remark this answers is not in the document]
maiqueb: (... and was by now fixed).

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,368 @@
...
+2. The cluster admin waits for the NodeNetworkConfigurationPolicy to succeed, freeing the IPs from those interfaces.
+
+### Workflow Description (scaling up cluster)
+
+1.  The cluster admin adds a new node to the already configured cluster using IPools
+2. The cluster admin waits for the NodeNetworkConfigurationPolicy to succeed, causing the new node to take the address from the NodeIPPool
+
+### Workflow Description (scaling down cluster)
+
+1. The cluster admin removes a node from an already configured cluster using `NodeIPool`
```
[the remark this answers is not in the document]
qinqon: The wording was missleading, after some changes is more clear, it means that the node has a reference to an IP from a `NodeIPPool` so it has to wait for the `NodeNetworkConfigurationPolicy` to success both at scale up and scale down.

design-proposals/host-ip-pool.md
qinqon: Right

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: Adding it to the doc.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: Right, keeping just the Available condition with proper Reason and adding a pair of example with the two errors.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,368 @@
...
+
+**cluster admin** is a user responsible for managing the cluster node
+networking.
+
+### Workflow Description (assign static address with IPPool)
+
+1. The cluster admin creates `NodeIPPool` resources with IP ranges and a name.
+2. The cluster admin configures node network interfaces statically, with IPs from the `NodeIPPool` using a
+`NodeNetworkConfigurationPolicy` referencing them.
+3. The cluster admin wait for the `NodeNetworkConfigurationPolicy` to success
```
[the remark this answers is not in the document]
creydr: I guess "The cluster admin wait" -> "The cluster admin wait**s**"

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### Variation (exhausted range at the NodeIPPool)
+
+When the IP pool does not have an available IP address, the `NNCP` will fail 
+with a explicit message, so the range has to be increased and NNCP applied again.
```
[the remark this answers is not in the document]
maiqueb: but why would you want to support an extra flow that is error prone ? If both are error prone, I would strive to support one - not two.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,398 @@
...
+  name: ostest
+platform:
+  baremetal:
+    networkPolicy:
+      ipPools:
+        traffic-1:
+          cidr: 10.10.10.0/24
+          start: 1
+          end: 10
+          exclude: [4, 7]
```
maiqueb, line 272: are these indexes ?... or just the last byte of the IP in decimal format ? Because they look like the latter. 

How will this fit a CIDR with - let's say - 16 bits for the network ?

qinqon: This example is wrong since `NodeIPPool` was changed since and now start, end and exclude use full IP address, only the allocate-offsets use index since those will be a massive number of entries (we can have cluster with thousand nodes).

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,407 @@
...
+### API Extensions day 2
+
+This proposal adds a new CRD to configure and store IP pools (NodeIPPool) and some
+changes at the NNCPs to reference the ippool.
+
+Following is an example CR for the cluster wide CRD `NodeIPPool`:
+
+```yaml
+apiVersion: nmstate.io/v1
+kind: NodeIPPool
```
maiqueb, line 108: can we see an example with an IPv6 node ip pool for completeness ? 

I'm trying to get my head around the offsets you have defined below and I still don't fully understand them. Are they indexes within the pool or the last bytes of the IP address ? 

If they're the latter, if you have a prefix length < 24, you'll have to encode them in a number, which will make debugging harder.

I would also like to see a scenario for a /16 CIDR, with some exclude ranges in the middle, so I can make sense of them, and get a hold of how hard it will be to debug this when it goes wrong. The idea would be an exercise of mapping the pool's allocated offsets with the IPs in the enactment.

qinqon: They are indexes more than offsets but they are not the last bytes, I am adding some examples to clarify it.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+
+1. The cluster admin changes the network interface to remove the reference to the `NodeIPPool` (for example harcoding the IP or switching to DHCP).
+2. The cluster admin waits for the NodeNetworkConfigurationPolicy to succeed, freeing the IPs from those interfaces.
+
+### Workflow Description (scaling up cluster)
+
+1. The cluster admin adds a new node affected by the `NodeNetworkConfigurationPolicy`.
+2. The cluster admin waits for the `NodeNetworkConfigurationPolicy` to succeed, causing the new node to take the address from the `NodeIPPool`
+
+### Workflow Description (scaling down cluster)
```
AlonaKaplan, line 86: What about the flow-  `The cluster admin removes from the node an interface with IP from the `NodeIPPool`?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+
+1. The cluster admin removes a node that was taking an address from a `NodeIPPool`.
```
AlonaKaplan, line 88: Can we have more than one IP on the node from the same `NodeIPPool`?
Can differenct interface on the node take IPs from different `NodeIPPool`s?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+2. The cluster admin waits for the `NodeNetworkConfigurationPolicy` to succeed, thus freeing the IPs from the removed node.
+
+### Workflow Description (manual NodeIPPool clean up)
+
+1. The cluster admin finds some stale IPs, They are inconsistencies between `NodeIPPool.status` and `NodeNetworkConfigurationEnactment.status`.
+2. The cluster admin resets the `NodeIPPool` allocated status specifying an empty list `status.allocated-offsets: []`.
```
AlonaKaplan, line 94: What is `status.allocated-offsets`? It wasn't mentioned earlier in the doc.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+2. The cluster admin configures node network interfaces statically, with IPs from the `NodeIPPool` using a
+`NodeNetworkConfigurationPolicy` referencing them.
+3. The cluster admin waits for the `NodeNetworkConfigurationPolicy` to succeed
+
+#### Variation (exhausted range at the NodeIPPool)
+
+When the IP pool does not have an available IP address, the `NNCP` will fail
+with a explicit message, the `NodeIPPool` has to increase pool size.
+
+Modifying `NodeIPPool` means increasing CIDR, changing begin/end or exclude,
```
AlonaKaplan, line 72: What if an already allocated IP is not in the range after the change?

maiqueb: +1.

I would also like to see a list of scenarios for which update is possible / not possible.

Keep in mind I think the update scenario is not required, and I have advocated in the past for us to keep the pool's spec immutable.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+
+Following is an example CR for the cluster wide CRD `NodeIPPool`:
+
+```yaml
+apiVersion: nmstate.io/v1
+kind: NodeIPPool
+metadata:
+  name: traffic-1
+spec:
+  range:
```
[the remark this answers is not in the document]
maiqueb: Why would a pool provide addresses from multiple ranges @AlonaKaplan ? 

Assuming there are multiple ranges, If a node requests an address from a pool, from which of the ranges will it get the IP ? The first ? Round-robin ? Random ?...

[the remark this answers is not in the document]
AlonaKaplan: yes, randomly one of the ranges.
Regarding the usage, I"m not sure. But for example in the [cisco documentation](https://www.cisco.com/c/en/us/td/docs/net_mgmt/prime/network/4-2-3/user/guide/CiscoPrimeNetwork423UserGuide/ip-pool.pdf) there is a concept of assigning an ip address from multiple ranges (group of pools).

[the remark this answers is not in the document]
maiqueb: I don't know ... Let's say you have a range with ipv6, and another with ipv4. 

This would leave you with either an ipv4 or ipv6 address, depending on... randomness.

If we take this esoteric route (I find it strange) I would at least assure that all ranges belong to the same family.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+## Motivation
+
+Using DHCP as the IP assignment mechanism has some drawbacks:
+- Introduce latencies on node restart.
+- It presents a single point of failure in case the DHCP server is failing
+- It depends on network setup that can be complex.
+
+To overcome the above-mentioned drawbacks, statically configured interfaces taking the address from a
+dynamic host IP pool is a solution. Since the interfaces are static there is no
+reconfiguration after some time or at node restart.
```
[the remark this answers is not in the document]
qinqon: When we hardcode the ip address statically per node, that's there until the user change it, this should work similar to that, we are not implementing DHCP lease.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: I thought about this a bit more and I honestly don't understand why would a pool feature multiple ranges. It can, but, the same can be done with multiple pools. 

Plus, I think if you have multiple ranges, you'll also have to keep track on those allocations for the different ranges in the pool. I just don't see the motivation for it - thus, I wouldn't account for it **until** there's a use case that requires it.

design-proposals/host-ip-pool.md
qinqon: I will add that workflow

design-proposals/host-ip-pool.md
qinqon: > Can we have more than one IP on the node from the same `NodeIPPool`? 
Yep, multiple "captures" will allocate multiple IPs that can be then referenced on the `desiredState`

Can differenct interface on the node take IPs from different `NodeIPPool`s?
The NNCE has a new field at status that show what IPs where taken from what pools, this allow the garbage collector to return the IP to the correct pool.

design-proposals/host-ip-pool.md
qinqon: > What is `status.allocated-offsets`? It wasn't mentioned earlier in the doc.

It's later on mentiones on the API section, it's the offset from the "begin" on the range for allocated IPs.

design-proposals/host-ip-pool.md
qinqon: > What if an already allocated IP is not in the range after the change?

The modification is restricted to not exclude already allocated IPs.

> Keep in mind I think the update scenario is not required, and I have advocated in the past for us to keep the pool's spec immutable.

We can stick with NodeIPPool re-create at this design doc and adding update `NodeIPPool` feature later on. @AlonaKaplan @maiqueb what do you think ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+### API Extensions day 2
+
+This proposal adds a new CRD to configure and store IP pools (NodeIPPool) and some
+changes at the NNCPs to reference the ippool.
+
+Following is an example CR for the cluster wide CRD `NodeIPPool`:
+
+```yaml
+apiVersion: nmstate.io/v1
+kind: NodeIPPool
```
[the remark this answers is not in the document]
qinqon: > Are we planning to have the same kind of pool for VMs/pods in the future? If so, the name `NodeIPPool` is a problematic since we won't we able to ruse this crd.

Nope this is just for nodes that's why it has the `Node` preffix.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: @AlonaKaplan since every range will need it's own allocated info it does not gives any value define multiple of them at the same `NodeIPPool` we are not going to have a massive number of ranges when we talk about interfaces on nodes.

design-proposals/host-ip-pool.md
maiqueb: +1 to keeping the pool spec immutable.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+easy to import ippool library from kubernetes-nmstate and nmpolicy to render the
+`networkPolicy`.
+
+[1] https://github.com/openshift/image-customization-controller/blob/3ad18911d14dde1c23b660f99bb05e12a21c99e4/pkg/ignition/builder.go#L79
+
+### Risks and Mitigations
+
+If a bug prevents IPs deallocation the ippool can be exhausted and end up
+with network reconfiguration or scaling up failures. To fix those situations
+the field `status.allocated-offsets` at `NodeIPPool` can be set with an
```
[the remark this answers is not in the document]
qinqon: > What is the trigger to reconstruct the pool? Empty `status.allocated-offsets`list? 

Yes

> How do we know the pool actually needs a reconstruction and it's not just the the list is empty because no address is used? 

If no address is used the `ippools` field at NNCE will be empty and nothing will be done

> What component is responsible to reconstruct the pool?

Is distributed across the handlers, every handler try to allocate an address for it's interfaces, we do an optimistic locking with retry on failure like we do at NNCP status.

design-proposals/host-ip-pool.md
AlonaKaplan: When you re-create the pool. What will happen to the nodes pointing to the old deleted pool?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+        dhcp: false
+        enabled: true
+```
+
+The allocated IPs will be marked at the `NodeNetworkConfigurationEnactment`
+status and will be used to re-construct the nmstate on reboot and to discover
+if new allocation if needed or to de-allocate them the pool is no longer
+referenced, this is done on distributed fashion by each handler at their nodes.
+
+The `NNCE.status.ippool` information allows to reconstruct `NodeIPPool.status.allocated-offsets`
```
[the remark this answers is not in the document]
AlonaKaplan: I still don't get it. If you change the "begin" you should update the offsets, right. But you have the offset list, you know what change was done to the begin, so you can re-calculate the offsets. Why in this case do you need to know what node it using the IP?

design-proposals/host-ip-pool.md
qinqon: The steps are:
- Create a new NodeIPPool (it has to be in the range of previous one)
- Change NNCPs to reference the new NodeIPPool
- Already reserved IPs will be marked as so in the pool and the node that is missig the IP will get new one.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: In case we continue allowing `NodeIPPool` update, it's true that offset can be recalculated comparing old a new begin but since this work is going to be distribuged across handlers doing the logic of just registering the IPs they have looks simpler.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: and the NNCP will be marced as not available.

qinqon:
@AlonaKaplan @maiqueb I have mark the `NodeIPPool` as read-only on the doc, can you take a look now ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+spec:
+  range:
+    cidr: 10.10.10.0/24
+    start: 10.10.10.1
+    end: 10.10.10.10
+    exclude: ["10.10.10.4", "10.10.10.7"]
+status:
+  allocated-offsets: [1, 2, 5, 6, 8]
+  conditions:
+  - type: Available
```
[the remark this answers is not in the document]
qinqon: Here https://github.com/nmstate/kubernetes-nmstate/pull/1088/files#diff-f8b6fe96158c42df632deb84deee540aff93c8560a4c4d99fc6db6645797b2efR220-R225

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,486 @@
...
+The `NNCE.status.ippool` information allows to reconstruct `NodeIPPool.status.allocated-offsets`
+when `cidr` and `begin` values are modified.
+
+```yaml
+apiVersion: nmstate.io/v1
+kind: NodeNetworkConfigurationEnactment
+metadata:
+  name: static-ip.node01
+status:
+  ippool:
```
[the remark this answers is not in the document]
qinqon: Here https://github.com/nmstate/kubernetes-nmstate/pull/1088/files#diff-f8b6fe96158c42df632deb84deee540aff93c8560a4c4d99fc6db6645797b2efR239-R241

controllers/operator/nmstate_controller.go
```diff
@@ -191,7 +191,13 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1.NMState) error {
 
 	handlerTolerations := instance.Spec.Tolerations
 	if handlerTolerations == nil {
-		handlerTolerations = []corev1.Toleration{operatorExistsToleration}
+		handlerTolerations = []corev1.Toleration{
```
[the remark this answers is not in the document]
qinqon: Fixed

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: We need the reference to the pool at NNCE so the handler is able to reconstruct the offset at `NodeIPPool` if needed.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,502 @@
...
+marked as unknown, so users can wait on it.
+
+Following is a table with the "type=Available" condition states
+
+| reason              | status  | message |
+| --------------------| --------| ------ |
+| ReconcilingPool     | Unknown | The pool is being updated |
+| ExhaustedIPPool     | False   | The IP pool has being depleted, manual intervention needed to modify NodeIPPool.|
+| UnsyncronizedIPPool | False   | Failed re constructing the IP pool, it may be not in sync with nodes. |
+| OverlapingIPPool    | False   | The NodeIPPool XYZ is overlapping with this one. |
```
AlonaKaplan, line 225: We shouldn't get this. But maybe it can be useful to have. `duplicate ip address`, in case two nodes got the same address somehow.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
AlonaKaplan: But in case we want the pool to forget the IP, we don't want it to be reconstructed.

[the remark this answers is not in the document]
qinqon: The user IP will be forget by removing it from NNCE ippools at status, so it will not be bringed back.

design-proposals/host-ip-pool.md
AlonaKaplan: I don't think that allocating a new address to one of them is a good option. Both may be running nodes, we cannot just change the IP of one of them.
The tool should prevent getting to this situation, but if the situation somehow happened (due to a bug), it should report about it.

qinqon: Fair enough, adding this error, we can implement some kind of remediation in the future, this can also come handy if users are configuring statically the cluster and they configured the same ip address at a pair of nodes by mistake.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+if new allocation if needed or to de-allocate them the pool is no longer
+referenced, this is done on distributed fashion by each handler at their nodes.
+
+The `NNCE.status.ippool` information allows reserve those IPs again if the
+`NodeIPPool` is re-created after exhaustion, since fields on the `NodeIPPool.spec` section
+are read-only for simplicity.
+
+If the IPs reserved for a node for a specific pool want to be deallocated the
+user can change `NNCE.status.ippool` fields if the node is Ready the IP allocation
+will try to get a new one for the desiredState.
```
maiqueb, line 276: This doesn't read well: I don't think IPs want to be deallocated. They don't want anything :P 

```suggestion
The user can change the `NNCE.status.ippool` attribute to deallocate the IPs reserved for a node.

If the node is Ready,  the nmstate will attempt to get a new IP allocation for the desiredState.
```

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+offsets on a range [1], a snapshot of the bitmap can be stored as base64 at the
+`NodeIPPool` status.
+
+For the scale down scenario, the current kubernetes-nmstate architecture
+contains a controller for node that will Reconcile on node deletion, but since
+the nodes may be drained before deletion the pod running the controller
+will no longer be running, this means that the controller may need to run at
+control-plane nodes and watch for all the nodes.
+
+[1] https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/ovn/ipallocator/allocator.go
```
maiqueb, line 407: Does this section make sense here ? Like it seems to not fit - at all - between the `Baremetal IPI` and the `API Extensions baremetal IPI` ? It seems to be a simple general purpose implementation detail. 

It totally broke my concentration on anything related to baremetal.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+
+If a bug prevents IPs deallocation the ippool can be exhausted and end up
+with network reconfiguration or scaling up failures. To fix those situations
+the field `status.allocated-offsets` at `NodeIPPool` can be set with an
+empty list triggering a pool reconstruction, bringing back the
+stale IPs. This risk gets higher for shorter IPPools.
+
+If the `NodeIPPool` is exhausted the values from the `spec` fields can be
+changed to add more free IPs to the pool. Doing so involves that handlers will
+reconstruct `status.allocated-offsets` with the already reserved IPs at NNCE.
```
maiqueb, line 427: didn't we say this is bad ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+
+### Drawbacks
+
+Updating the pool is distributed across nodes by the nmstate-handler pod, this
+means that N nodes will try to update `NodeIPPool` and retry on conflict. Is
+not expected the IPAM mechanism will run very often so this should not be a
+problem.
```
maiqueb, line 434: famous last words :) 

I unfortunately don't see a way to mitigate this risk.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+
+Debugging the allocated offset can be a challange since it involve a convertion
+from the offset to the IP is pointing too depending on `NodeIPPool` spec.
```
maiqueb, line 437: Well, as long as there's a translation mechanism involved, we can always create a tool to help out.

It would make it definitely simpler. Consider adding it to the implementation schedule (very end of the list).

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+
+Fixing the "exhausted" issue updating the NodeIPPool involves a lot of logic to
+reconstruct the `allocated-offsets` that has to be carefully implemented and
+tested, what alternate would be to just delete `NodeIPPool` and create new one,
+so `NodeIPPool` is a read-only resource.
```
maiqueb, line 442: I honestly wouldn't even bother. Plan well in advance, use huge ranges if you plan on having lots and lots of hosts. 

But your current proposal seems more conservative than the previous option. I'm OK with it.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+
+The baremetal IPI Nmstate does an offline NetworkManager keyfile generation so
+there is no network `currentState` that nmpolicy can reference to, so using
+nmpolicy just to reference the ippools may be too overkill.
+
+If in the future nmstatectl is included at FCOS then the `networkPolicy` will
```
maiqueb, line 448: What's an FCOS ? (link to the fedora coreOS page please).

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+- Add automatic garbage collector for stale IPs
+
+## Alternatives
+
+An alternative is to implement directly an IPAM feature at Nmstate with a
+plugin system for ippool retrieval and allocation storage so at day 2
+it will use the `NodeIPPool` and `NodeNetworkConfigurationEnactment` and
+at baremetal IPI It will read the ippool from the openshift installer config
+and store the allocations at the nodes file system, later on it will be
+converted into proper NodeIPPool and NNCEs using different storage plugin.
```
maiqueb, line 478: I miss a comparison between the proposed solution and the alternative. 

Why did you choose one over the other ? What were its advantages over this one ?

maiqueb (changes requested):

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
qinqon: Is doable technically, but maybe is better to implemnent the mechanism with a annotation. 

```yaml
annotations:
  nmstate.io/reset-ipool: traffic1
```

We  can use this at NNCP (it will all nodes affected by it) or at NNCE (it will affect only the NNCE node).

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+- Metal3 IPAM: https://metal3.io/blog/2020/07/06/IP_address_manager.html
+  - It depends on CAPI (ClusterAPI)
+  - It's though for Metal3 baremetal.
+- Azure private addresses: https://docs.microsoft.com/en-us/azure/virtual-network/ip-services/private-ip-addresses
+  - Specific to azure
+  - Some code has to be put in place to consume it.
+
+### User Stories
+
+- As a cluster administrator, I want to configure static IPs on secondary NICs on a pool of nodes without creating one per each node.
```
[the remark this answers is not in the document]
qinqon: They are usually the NICs tha are not part of control-plane, so if we missconfigure them k8s is still working fine.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+When the IP pool does not have an available IP address, the `NNCP` will fail
+with a explicit message, the `NodeIPPool` has to increase pool size.
+
+Modifying `NodeIPPool` means increasing CIDR, changing begin/end or exclude,
+since allocations are stored as offsets they will have to be re-calculated from
+the full IP present at NNCEs.
+
+### Workflow Description (removing a reference to NodeIPPool)
+
+1. The cluster admin changes the network interface to remove the reference to the `NodeIPPool` (for example harcoding the IP or switching to DHCP).
```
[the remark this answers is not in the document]
qinqon: Well know that I think of a NNCP with captures cannot be modified, so the only way would be user deleting the NNCP, meaning that all ips are de-allocated.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+In this process the reason will be "ReconcilingPool" and `Available`
+marked as unknown, so users can wait on it.
+
+Following is a table with the "type=Available" condition states
+
+| reason              | status  | message |
+| --------------------| --------| ------ |
+| ReconcilingPool     | Unknown | The pool is being updated |
+| ExhaustedIPPool     | False   | The IP pool has being depleted, manual intervention needed to modify NodeIPPool.|
+| UnsyncronizedIPPool | False   | Failed re constructing the IP pool, it may be not in sync with nodes. |
```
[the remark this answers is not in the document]
qinqon: Maybe re-creating NNCPs, I am not sure to be honest

design-proposals/host-ip-pool.md
qinqon: A user can be already assining the same address statically at the same interface, or at different interfaces, we can detect those.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+| OverlapingIPPool    | False   | The NodeIPPool XYZ is overlapping with this one. |
```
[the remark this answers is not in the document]
qinqon: You mean add an admission webhook to check this ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+
+### API Extensions baremetal IPI
+
+At baremetal IPI the Nmstate configuration is passed to openshift installer
+using the baremetal host `networkConfig` field [1].
+
+[1] https://github.com/openshift/installer/blob/ca8aacd3c931eefabfddafb49cb44b76f24653a6/docs/user/metal/install_ipi.md#describing-hosts
+
+To accommodate the ippools, capture and nmstate new fields will be needed there
+the following is an example of the install-config.yaml from openshift with
```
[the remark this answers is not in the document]
qinqon: I am going to remove baremetal IPI and day 0 day 1 from here

design-proposals/host-ip-pool.md
qinqon: Going to rework the doc to remove baremetal IPI, that has to be done at a follow up.

design-proposals/host-ip-pool.md
qinqon: Removing baremetal IPI reference.

design-proposals/host-ip-pool.md
qinqon: Well there are not much alternatives after looking around.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: But how do they look like to the user (cluster-admin) ? I.e. is there a property on the interfaces that can be used to understand which ifaces are secondary ? 

If you don't mind, please define what a "secondary" interface is, to make things clear.

design-proposals/host-ip-pool.md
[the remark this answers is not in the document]
maiqueb: It is a good idea, which I think makes it more API driven, and less prone to errors.

Don't want to get in your way though, you can start small, and implement this after the MVP.

Just for future reference, this is how I see it happening (sure you think the same):
  - user would add the annotation
  - the controller would reconcile the enactments across all the cluster (not getting into detail of if each local agent resets its part of the pool or not)
  - once that's done, the controller removes the annotation.

If this happens, it would be interesting to throw an event / add a condition. 

I would favor the condition since that lives forever, while the event is deleted after a short while.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,503 @@
...
+metadata:
+  name: traffic-2
+spec:
+  range:
+    cidr: 10.10.11.0/24
+    start: 10.10.11.1
+    end: 10.10.10.11.10
+    exclude: ["10.10.11.4", "10.10.11.7"]
+status:
+  allocated-offsets: [2, 5, 6, 8]
```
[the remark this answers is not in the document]
qinqon: At the end I am changin the offset to start with "0" since begin + offset should give you the the IP.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Workflow Description (scaling down cluster)
+
+1. The cluster admin removes a node that was taking an address from a `NodeIPPool`.
+2. The cluster admin waits for the `NodeNetworkConfigurationPolicy` to succeed, thus freeing the IPs from the removed node.
+
+### Workflow Description (manual NodeIPPool clean up)
+
+1. The cluster admin finds some stale IPs, They are inconsistencies between `NodeIPPool.status` and `NodeNetworkConfigurationEnactment.status`.
+2. The cluster admin resets the `NodeIPPool` following instructions at "API Exntensions/Manual Garbage Collection" chapter.
```
maiqueb, line 89:
```suggestion
2. The cluster admin resets the `NodeIPPool` following instructions at "API Extensions/Manual Garbage Collection" chapter.
```

maiqueb: I would also add a link to that heading.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,442 @@
...
+3. The cluster admin waits for `NodeIPPool` status to succeed
+
+NOTE: This is a manual garbage collector that is good enough for first implementation it can be converted into automatic in the future.
```
maiqueb, line 92:
```suggestion
**NOTE**: This is a manual garbage collector that is good enough for first implementation it can be converted into automatic in the future.
```

maiqueb (changes requested):
Thanks. 

I have spotted a couple of typos, request for a link to be added.
Other than that, it's OK to me.

maiqueb (approved):
Thanks.

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,442 @@
...
+
+
+### Similar product
+
+- Metal3 IPAM: https://metal3.io/blog/2020/07/06/IP_address_manager.html
+  - It depends on CAPI (ClusterAPI)
+  - It's though for Metal3 baremetal.
+- Azure private addresses: https://docs.microsoft.com/en-us/azure/virtual-network/ip-services/private-ip-addresses
+  - Specific to azure
+  - Some code has to be put in place to consume it.
```
dankenigsberg, line 33: Can you add a reference to how assisted installer does this on day 1 ?

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,442 @@
...
+
+This functionality was requested on U/S via https://github.com/nmstate/kubernetes-nmstate/issues/725
+
+Although this proposal will present a solution with an agnostic storage
+mechanism, following is a spike done using whereabouts at the storage
+https://github.com/nmstate/kubernetes-nmstate/pull/1081
+
+
+### Similar product
+
```
dankenigsberg, line 27: @EdDev asks: how does bare-metal IT do this outside the context of k8s?

design-proposals/host-ip-pool.md
dankenigsberg: I'm told it's somewhere in https://github.com/openshift/assisted-service/blob/9bae4c3662e4755e80251284aba529318d76c8e9/docs/user-guide/restful-api-guide.md#additional-nmstate-configuration-examples

design-proposals/host-ip-pool.md
```diff
@@ -0,0 +1,442 @@
...
+for cluster nodes, supporting cluster scaling up and down.
+
+## Motivation
+
+Using DHCP as the IP assignment mechanism has some drawbacks:
+- Introduce latencies on node restart.
+- It presents a single point of failure in case the DHCP server is failing
+- It depends on network setup that can be complex.
+
+To overcome the above-mentioned drawbacks, statically configured interfaces taking the address from a
```
EdDev, line 15: The allocation and management of IP address on bare metal machines is a topic the IT industry had to tackle for the early days of computers.

I am sure DHCP and other options exists today which solve this challenge regardless of the kubernetes ecosystem.

I would suggest to examine existing enterprise and telco grade solutions that exist in the market, to make sure we are not re-inventing the wheel.

It is also not that trivial to claim that the DHCP drawbacks described above worth the investment of an in-house IP pool management. I would suggest first interviewing IT experts in this field before we commit to develop such a solution (which from first sight, looks very challenging and will require a lot of investment for the long run).

PR #1091, opened 2022-05-31
docs, examples: Migration DHCP -> static

qinqon:
Signed-off-by: Enrique Llorente <ellorent@redhat.com>

>
> /kind bug
> /kind enhancement

We are missing a feature at nmpolicy to generate a capture from a path so currentState can be referenced without filtering.

```yaml
capture: 
  dns: dns-resolver.running
```

Depends on:
- [x] https://github.com/nmstate/nmpolicy/pull/101
- [ ] https://github.com/kubevirt/kubevirtci/pull/816

**Release note**:

```release-note

```


phoracek:
Thanks. We should link to it from the README.md and in the user-guide

phoracek (changes requested):
Thanks. We should link to it from the README.md and in the user-guide

PR #1097, opened 2022-06-07
Nmpolicy v0.3.0

qinqon:

/kind enhancement

This PR bumps nmpolicy to v0.3.0 to have the capture path filter feature.

https://nmstate.io/nmpolicy/user-guide/102-policy-syntax.html#path-filter-pathexpression

**Release note**:

```release-note
Bump nmpooicy to v0.3.0
```

### DIFF
--- a/controllers/handler/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/handler/nodenetworkconfigurationpolicy_controller.go
@@ -328,7 +328,7 @@ func (r *NodeNetworkConfigurationPolicyReconciler) fillInEnactmentStatus(
 		return err
 	}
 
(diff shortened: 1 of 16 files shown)


phoracek:
@qinqon would you include an example of how to use this in knmstate user-guide too? So people don't have to dive into manuals. After all we keep examples of nmstate configurations, despite they are already in nmstate.io.

qinqon:
> @qinqon would you include an example of how to use this in knmstate user-guide too? So people don't have to dive into manuals. After all we keep examples of nmstate configurations, despite they are already in nmstate.io.

This is being done here https://github.com/nmstate/kubernetes-nmstate/pull/1091

PR #1106, opened 2022-07-11
infra, release: Move goflags to main

qinqon:
Having GOFLAGS as part of command does not work, this change move it to
the main part of the script after make so it does not affect how
knmstate is build.

```release-note
NONE
```


rhrazdil (approved):
Thanks
/lgtm
/approve

PR #1116, opened 2022-08-29
Refact probe timeouts

rhrazdil:

>
> /kind bug
> /kind enhancement

This PR chages the timeout logic, moving the retry responsibility
out of the probe functions. This simplifies the probe func interface,
as we don't need to pass a timeout parameter, and it's handled by the
caller.

Additionally, some comments seemed unclear. Second commit of this PR attempts to improve those.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/probe/probes.go
+++ b/pkg/probe/probes.go
@@ -46,9 +46,9 @@ var (
 )
 
 type Probe struct {
-	name    string
-	timeout time.Duration
-	run     func(client.Client, time.Duration) error
+	name      string
+	timeout   time.Duration
+	condition func(client.Client) wait.ConditionFunc
 }
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/probe/probes.go
```diff
@@ -77,174 +77,165 @@ func currentStateAsGJson() (gjson.Result, error) {
...
+	// [1] https://github.com/operator-framework/operator-sdk/blob/master/doc/user/client.md#non-default-client
+	currentConfig, err := config.GetConfig()
+	if err != nil {
+		return errors.Wrap(err, "getting config")
+	}
+
+	// Since we are going to retrieve Nodes default schema is good
+	// enough, also align timeout with poll timeout
+	currentConfig.Timeout = apiServerProbeTimeout
+	cli, err := client.New(currentConfig, client.Options{})
```
qinqon, line 105: Do we really need this one ? I supose that we have similar PollImmediate behind the scenes

pkg/probe/probes.go
```diff
@@ -77,174 +77,165 @@ func currentStateAsGJson() (gjson.Result, error) {
...
-		}
-		err = cli.Get(context.TODO(), types.NamespacedName{Name: metav1.NamespaceDefault}, &corev1.Namespace{})
-		if err != nil {
-			log.Error(err, "failed reaching the apiserver")
-			return false, nil
-		}
-		return true, nil
-	})
+// This probes use its own client to bypass cache
+func checkAPIServerConnectivity() error {
```
qinqon, line 94: Let's implement return `(bool, error)` here to be similar to `wait.ConditionFunc` so whenever we want to retry or stop is decied by the low level functions and make it easier to adapt to PollImmediate.

pkg/probe/probes.go
```diff
@@ -77,174 +77,165 @@ func currentStateAsGJson() (gjson.Result, error) {
...
+	runningNameServers := currentStateAsGJson.Get(runningServersGJsonPath).Array()
+	if len(runningNameServers) == 0 {
+		return fmt.Errorf("missing name servers at '%s' on %s", runningServersGJsonPath, currentStateAsGJson.String())
+	}
+	for _, runningNameServer := range runningNameServers {
+		r := &net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				d := net.Dialer{
+					Timeout: defaultDNSProbeTimeout,
```
qinqon, line 186: Do we need this Timeout ? we are going to poll

pkg/probe/probes.go
```diff
@@ -77,174 +77,165 @@ func currentStateAsGJson() (gjson.Result, error) {
...
+		},
+	}
+
+	for _, p := range externalConnectivityProbes {
+		err := wait.PollImmediate(time.Second, p.timeout, func() (bool, error) {
+			err := p.run(cli)
+			if err != nil {
+				return false, nil
+			}
+			return true, nil
```
qinqon, line 224: Let's create a "factory function" with cli that returns `wait.ConditionFunc` function expected by `PollImmediate`, this will simplify the polling code 

```golang
func pingCondition(_ client.Client) wait.ConditionFunc {
    return func() (bool, error) {
       ...
       return ping(defaultGw)
    }
}


func dnsCondition(_ client.Client) wait.ConditionFunc {
   return func() (bool, error) {
   ....
   }
}

{
			name:    "dns",
			timeout: defaultDNSProbeTimeout,
			condition:     dnsCondition,
		},

err := wait.PollImmediate(time.Second, p.timeout, p.condition(cli))
if err != nil {
...
}
`
```

test/e2e/handler/error_messages_formatting_test.go
```diff
@@ -73,14 +73,15 @@ func createPolicyAndWaitForEnactmentCondition(policy string, desiredState func()
 	waitForNodesReady()
 
 	By("Waiting for enactment to be failing")
-	policyconditions.EnactmentConditionsStatusEventually(
+	policyconditions.EnactmentConditionsStatusForPolicyEventually(
```
qinqon, line 76: This isn't related to the PR I think.

qinqon (changes requested):
Let's return `(bool, error)` at low level functions to follow `wait.ConditionFunc`

test/e2e/handler/error_messages_formatting_test.go
rhrazdil: It fixes a flake, I have moved it to a separate commit

pkg/probe/probes.go
rhrazdil: Fixed by setting the client timeout to 0 (to disable it)

pkg/probe/probes.go
rhrazdil: Done

pkg/probe/probes.go
rhrazdil: Removed

pkg/probe/probes.go
rhrazdil: Done

qinqon (approved):
/lgtm
/approve

PR #1118, opened 2022-09-02
Respect CR node selector for NMState handler

### LINKED ISSUE #1117: NMState Handler not deploying to Arm node on Multi-Arch Cluster
**What happened**:
When I installed the kubernetes-nmstate into my multi-arch cluster composed of arm64 and amd64 nodes, I noticed that the nmstate handler pods are not deploying to the arm64 node. I've tried adding nodeSelectors on the CR, however, the handler DaemonSet has a hardcoded nodeSelector that is forcing it to only go on the amd64 node.

The nmstate handler should have also deployed to the arm node

1. Have a multi-architecture cluster with amd64 and arm64 nodes
2. Install kubernetes-nmstate
3. Notice how only the amd64 nodes have the handler pod

I believe this is the line of code responsible for this:
https://github.com/nmstate/kubernetes-nmstate/blob/1216357c9081c4f36d591e9716860f81135b3568/controllers/operator/nmstate_controller.go#L197

Is there a valid reason as to why that nodeSelector is hardcoded to be the `goruntime.GOARCH`? Could that line be simply removed or be able to be replaced at least by the NMState CR's nodeSelectors?

**Environment**:

- kubernetes-nmstate image (use `kubectl get pods --all-namespaces -l app=kubernetes-nmstate -o jsonpath='{.items[0].spec.containers[0].image}'`): quay.io/nmstate/kubernetes-nmstate-handler:v0.72.0
- Kubernetes version (use `kubectl version`): 1.21.13
- OS (e.g. from /etc/os-release): ubuntu 20.04

MarkSpencerTan:
Signed-off-by: Mark Spencer Tan <msgtan@gmail.com>

>
/kind bug
> /kind enhancement

fixes #1117 

This PR fixes the handler's node selectors not being overridden properly with the CR's node selectors due to the `"kubernetes.io/arch": goruntime.GOARCH` node selector being added statically. This is problematic for Multi-Architecture clusters since the handler will only deploy on one of the arch types rather than all.

**Release note**:

```release-note
Fixed the nmstate handler's node selectors not being overridden properly with the CR's node selectors
```

### DIFF
--- a/controllers/operator/nmstate_controller.go
+++ b/controllers/operator/nmstate_controller.go
@@ -199,11 +199,11 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1.NMState) error {
 	}
 	archAndCRNodeSelector := instance.Spec.NodeSelector
 	if archAndCRNodeSelector == nil {
-		archAndCRNodeSelector = map[string]string{}
+		archAndCRNodeSelector = map[string]string{
+			"kubernetes.io/arch": goruntime.GOARCH,
+			"kubernetes.io/os":   "linux",
+		}
 	}
-	archAndCRNodeSelector["kubernetes.io/arch"] = goruntime.GOARCH
-	archAndCRNodeSelector["kubernetes.io/os"] = "linux"
-
 	handlerTolerations := instance.Spec.Tolerations
 	if handlerTolerations == nil {
 		handlerTolerations = []corev1.Toleration{operatorExistsToleration}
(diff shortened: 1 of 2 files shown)


qinqon (approved):
/lgtm
/approve

MarkSpencerTan:
Thanks @qinqon for merging my PR! Would you be able to create a release tag for this as well so we can use it in our environment?

Thanks so much

PR #1119, opened 2022-09-14
Add test exercising lldp configuration with nmpolicy

rhrazdil:

>
> /kind bug
> /kind enhancement

This PR bumps nmpolicy dependency to allow configuration of boolean values.
In the second commit, this feature is used in an e2e test that configures lldp with nmpolicy capture.
In the third commit, an example of enabling lldp on all ethernet interfaces that are up is added.

**Release note**:

```release-note
bump nmpolicy to v0.5.0
```

### DIFF
--- a/docs/examples/lldp.yaml
+++ b/docs/examples/lldp.yaml
@@ -1,0 +1,11 @@
+apiVersion: nmstate.io/v1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: lldp-enabled-on-eths
+spec:
+  capture:
+    ethernets: interfaces.type=="ethernet"
+    ethernets-up: capture.ethernets | interfaces.state=="up"
+    ethernets-lldp: capture.ethernets-up | interfaces.lldp.enabled:=false
+  desiredState:
+    interfaces: "{{ capture.ethernets-lldp.interfaces }}"
--- a/test/e2e/handler/lldp_with_nmpolicy_test.go
+++ b/test/e2e/handler/lldp_with_nmpolicy_test.go
@@ -1,0 +1,70 @@
+/*
+Copyright The Kubernetes NMState Authors.
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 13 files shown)


test/e2e/handler/lldp_with_nmpolicy_test.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+
+	interfacesWithLldpEnabledState := nmstate.NewState(`interfaces: "{{ capture.ethernets-lldp.interfaces }}"`)
+
+	BeforeEach(func() {
+		By("Enabling LLDP on up ethernet interfaces")
+		setDesiredStateWithPolicyAndCapture(lldpEnabledPolicyName, interfacesWithLldpEnabledState, configureLldpOnEthernetsCapture("true"))
+
+		DeferCleanup(func() {
```
qinqon, line 48: We can put all the AfterEach on the DeferCleanup right ? also look slike we are repeting the DeferCleanup at the AfterEach.

test/e2e/handler/lldp_with_nmpolicy_test.go
```diff
@@ -0,0 +1,78 @@
...
+			deletePolicy(lldpDisabledPolicyName)
+		})
+
+		By("Wait for policy to be ready")
+		policy.WaitForAvailablePolicy(lldpDisabledPolicyName)
+	})
+
+	It("should enable LLDP on all ethernet interfaces that are up", func() {
+		By("Wait for policy to be ready")
+		policy.WaitForAvailablePolicy(lldpEnabledPolicyName)
```
qinqon, line 66: Let's move this to BeforeEach, the It just check the values.

qinqon (changes requested):
Can we add an example of the LLDP ?

test/e2e/handler/lldp_with_nmpolicy_test.go
rhrazdil: Added an example yaml, moved AfterEach into DeferCleanup

test/e2e/handler/lldp_with_nmpolicy_test.go
rhrazdil: Done

docs/examples/lldp.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: nmstate.io/v1
```
qinqon, line 1: we should test this example

rhrazdil: we can do so once we have a release with the lldp.yaml, otherwise it break upgrade

qinqon: why we cannot do it on this PR ? there is  example test suite at test/e2e

qinqon (approved):
/lgtm
/approve

PR #1120, opened 2022-09-15
e2e,upgrade: fall back to current examples if needed

rhrazdil:

>
> /kind bug
> /kind enhancement

When testing upgrade, example yamls from previous tag are downloaded.
This causes an issue when adding a new example yaml, because it doesn't yet exist in previous tag.

This change falls back to current doc/examples path if an example yaml is not found in previous tag.
This allows to smoothly add new example yamls.

**Release note**:

```release-note
NONE
```


qinqon (commented):
/lgtm
/approve

PR #1132, opened 2022-12-22
go: Bump to v1.18

qinqon:
Looks like v1.17 is not at centos9 anymore, that's a problem for rust branch where we are using centos9 and also in the future.

**Release note**:

```release-note
Bump to go1.18
```


RamLavi (approved):
/lgtm

phoracek (approved):

PR #1135, opened 2023-01-04
docs: Pin ruby to 3.1

qinqon:
Looks like Jekyll has a pin to a old Liquid version that does not works with ruby 3.2, this change pins to ruby 3.1 so there is no issue.

**Release note**:

```release-note
NONE
```


phoracek (commented):
/lgtm
/approve

PR #1137, opened 2023-01-11
handler: Rollback always before Apply

qinqon:
/kind bug

Sometimes a pending checkpoint is lingering around, for example when a handler pod has being restarted before reaching Commit, also looks like the checkpoint feature is not used except for kubernetes-nmstate. This change do Rollback to remove the pending checkpoing before Apply since apply on pending checkpoing fails.

Closes https://bugzilla.redhat.com/show_bug.cgi?id=2116562
There is a bug at latest nmstate with a fix that has not reach centos stream 8 yet

**Release note**:

```release-note
Removing pending checkpoints before apply a state using rollback
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -166,6 +166,10 @@ func ApplyDesiredState(cli client.Client, desiredState shared.State) (string, er
 	// working fine after apply
 	probes := probe.Select(cli)
 
+	// Rollback before Apply to remove pending checkpoints (for example handler pod restarted
+	// before Commit)
+	nmstatectl.Rollback()
+
 	setOutput, err := nmstatectl.Set(desiredState, DesiredStateConfigurationTimeout)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cybertron:
Seems reasonable to me. At the end of a call to Apply the checkpoints should either have been rolled back or committed, so the only way one would be pending is if something got interrupted.

qinqon:
/retest

Let's check if centos stream 8 has latest libnmstate with the fix.

qinqon:
Still old libnmstate

```
python3-libnmstate              noarch    1.4.0-1.el8       appstream    268 k
```

qinqon:
Still latest nmstate `1.4.0-1.el8` is not a centos 8 stream

qinqon:
Old version:
```
1.4.0-1.el8
```

phoracek:
/lgtm
/approve

For Ben!

PR #1138, opened 2023-01-16
handler: Temporaly use Gris repo

qinqon:
Currently CI is blocked at nmstate bug [1] and the fix is not landing centos stream 8 since there is a problem at the nmstate supply chain [2].

This change use Gris dnf repo temporaly while the issue is fixed.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2158151 
[2] https://issues.redhat.com/browse/CS-1373

**Release note**:

```release-note
NONE
```


qinqon:
/cc @cybertron 

There is a problem at centos 8 stream

PR #1153, opened 2023-03-02
Probes: choose the default gw from the main routing table

fedepaol:

>
/kind bug
> /kind enhancement

The current logic relies on the first matching route entry with destination == 0.0.0.0/0, in order to find the default gateway to ping. When multiple routing table exist, we may have multiple entries with that destination, some of them may not be usable and even if they were, pinging that particular gateway doesn't assure the health of the node.

To overcome this, we narrow the filter to pick the next hop of the default route for the "main" routing table.
 
**Release note**:

```release-note
Fix the probe check when configuring multiple routing tables and they contain a default gateway that is not accessible.
```

### DIFF
--- a/pkg/probe/probes.go
+++ b/pkg/probe/probes.go
@@ -57,6 +57,7 @@ const (
 	defaultDNSProbeTimeout    = 120 * time.Second
 	apiServerProbeTimeout     = 120 * time.Second
 	nodeReadinessProbeTimeout = 120 * time.Second
+	mainRoutingTableID        = 254
 	ProbesTotalTimeout        = defaultGwRetrieveTimeout +
 		defaultDNSProbeTimeout +
 		defaultDNSProbeTimeout +
@@ -69,12 +70,19 @@ func currentStateAsGJson() (gjson.Result, error) {
 	if err != nil {
 		return gjson.Result{}, errors.Wrap(err, "failed retrieving current state")
 	}
-
-	currentState, err := yaml.YAMLToJSON([]byte(observedStateRaw))
+	res, err := yamlToGJson(observedStateRaw)
 	if err != nil {
-		return gjson.Result{}, errors.Wrap(err, "failed to convert current state to JSON")
+		return gjson.Result{}, errors.Wrap(err, "failed to convert the current state to GJson")
 	}
-	return gjson.ParseBytes(currentState), nil
+	return res, nil
+}
+
+func yamlToGJson(rawYaml string) (gjson.Result, error) {
+	json, err := yaml.YAMLToJSON([]byte(rawYaml))
+	if err != nil {
+		return gjson.Result{}, err
+	}
+	return gjson.ParseBytes(json), nil
 }
 
 func apiServerCondition(_ client.Client) wait.ConditionFunc {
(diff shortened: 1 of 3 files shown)


fedepaol:
My understanding is, gjson does not allow to have multiple queries. See https://github.com/tidwall/gjson/issues/64

pkg/probe/probes.go
```diff
@@ -127,25 +134,31 @@ func checkNodeReadiness(cli client.Client) (bool, error) {
...
+				return false
+			}
+			return true
+		},
+	)
+
+	if !found.Exists() {
 		msg := "default gw missing"
-		defaultGwLog := log.WithValues("path", defaultGwGjsonPath)
+		defaultGwLog := log.WithValues("path", "routes.running.next-hop-address")
```
qinqon, line 152: Can we add the table-id as part of the log ?

pkg/probe/probes.go
```diff
@@ -127,25 +134,31 @@ func checkNodeReadiness(cli client.Client) (bool, error) {
...
-	}
-	defaultGwGjsonPath := "routes.running.#(destination==\"0.0.0.0/0\").next-hop-address"
-	defaultGw := gjsonCurrentState.Get(defaultGwGjsonPath).String()
-	if defaultGw == "" {
+func defaultGw(currentState gjson.Result) (string, error) {
+	var found gjson.Result
+	currentState.Get("routes.running").ForEach(
+		func(_, v gjson.Result) bool {
+			// we want to pick the next hop related to the "main" table because we may have multiple tables
+			if v.Get("destination").String() == "0.0.0.0/0" && v.Get("table-id").Int() == 254 {
```
qinqon, line 142: Can we replace the "magic" number with a little const at top of probes.go ? like `mainRoutingTableID`, is the best way to document stuff.

pkg/probe/probes.go
fedepaol: done

pkg/probe/probes.go
fedepaol: done

fedepaol:
Added, and also I added an e2e as requested offline

andreaskaris:
omg, I can't believe that we ran into the same issue with 2 weeks difference and that I didn't check your PR first -_-
https://github.com/nmstate/kubernetes-nmstate/pull/1159

qinqon:
@fedepaol is this good to go ?

qinqon:
> omg, I can't believe that we ran into the same issue with 2 weeks difference and that I didn't find your PR first -_- I did find the syntax for defaultGwGjsonPath , though, so perhaps you can simplify your commit if your fancy: #1159 [d291173#diff-0c0168aef7c305629b107f3f2b5512406c3baa13146a6cce9463b244012634f1R147](https://github.com/nmstate/kubernetes-nmstate/commit/d29117392c0a4522bff21b413cadf9a1fafc32df#diff-0c0168aef7c305629b107f3f2b5512406c3baa13146a6cce9463b244012634f1R147)
> 
> gjson syntax that you were looking for:
> 
> a) that matches the first default route from table 254, table 0 or no table-id set (I dunno if table is always set):
> 
> ```
> defaultGwGjsonPath := fmt.Sprintf("[routes.running.#(table-id==~false)#,routes.running.#(table-id==%d)#].@flatten"+
> 		"|#(destination==\"0.0.0.0/0\").next-hop-address", mainRoutingTableID)
> ```
> 
> b) To match the first default route from table 254:
> 
> ```
> defaultGwGjsonPath := fmt.Sprintf("routes.running.#(table-id==%d)#" +
> 		"|#(destination==\"0.0.0.0/0\").next-hop-address", mainRoutingTableID)
> ```

Better keep the loops so we don't depend on gjson syntax (we may ditch it in the future)

PR #1165, opened 2023-04-04
Don't deploy UI Plugin resources if the manifests are missing

orenc1:

>
/kind bug
> /kind enhancement

If for whatever reason the UI manifests does not exist in the operator's filesystem, we don't want the operator to deploy UI-related resources, resulting in an error log.

**Release note**:

```release-note
Don't deploy UI Plugin resources if the manifests are missing
```

### DIFF
--- a/controllers/operator/nmstate_controller.go
+++ b/controllers/operator/nmstate_controller.go
@@ -147,7 +147,8 @@ func (r *NMStateReconciler) applyManifests(instance *nmstatev1.NMState, ctx cont
 	}
 
 	isOpenShift, err := cluster.IsOpenShift(r.APIClient)
-	if err == nil && isOpenShift {
+	_, errUIPluginPathExists := os.Stat(filepath.Join("openshift", "ui-plugin"))
+	if err == nil && isOpenShift && errUIPluginPathExists == nil {
 		if err = r.applyOpenshiftUIPlugin(instance); err != nil {
 			return ctrl.Result{}, errors.Wrap(err, "failed applying UI Plugin")
 		}


controllers/operator/nmstate_controller.go
```diff
@@ -147,7 +147,8 @@ func (r *NMStateReconciler) applyManifests(instance *nmstatev1.NMState, ctx cont
 	}
 
 	isOpenShift, err := cluster.IsOpenShift(r.APIClient)
-	if err == nil && isOpenShift {
+	_, errUIPluginPathExists := os.Stat(filepath.Join("openshift", "ui-plugin"))
+	if err == nil && isOpenShift && errUIPluginPathExists == nil {
```
dougsland, line 151: just a small question: should we notify users via log that ui-plugin is not installed ?

Example:
```
if "NMState version supports ui-plugin" && errUIPluginPathExists != nil {
    r.Log.Info("NMState ui-plugin is not set, to enable this feature, please consult NMState documentation");
}
```

orenc1: i'm not sure, because if at this point the manifests of the ui plugin are missing from the operator container, the users can't do much in order to deploy the ui plugin at this point, since the operator is expecting the manifests to be included in the container.

dougsland: thanks! I will let others in the group review and approve :)

qinqon:
/ok-to-test
/retest

There is something bad at centos 9 stream

dougsland:
The error doesn't seem related to the patch:
```
 GPG Keys are configured as: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
Public key for scl-utils-2.0.3-4.el9.x86_64.rpm is not installed. Failing package is: scl-utils-1:2.0.3-4.el9.x86_64
 GPG Keys are configured as: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: GPG check FAILED
Error: building at STEP "RUN dnf install -y golang-$(sed -En 's/^go +(.*+)$/\1/p' go.mod).*": while running runtime: exit status 1
```

cybertron:
There's a problem with the centos 9 repos. A workaround is proposed here https://github.com/nmstate/kubernetes-nmstate/pull/1168 so that will need to merge before this will pass.

qinqon:
/retest

The centos 9 stream RPM issue has being fixed upstream, but looks like we have e2e test issue at main

PR #1168, opened 2023-04-05
build: Add centos 9 stream bug workaround

qinqon:
Currently there is a bug at centos 9 stream that prevent installing packages from a container, this change add a workaround for it.

**Release note**:

```release-note
NONE
```


phoracek (approved):
/lgtm
/approve

PR #1175, opened 2023-04-14
cluster: Skip NM auto-profile for sec interfaces

qinqon:
Sometimes the eth2 interface comes up without explicitly saying so. This change copy the mechanism at openshift operator to prevent that [1].

[1] https://github.com/openshift/kubernetes-nmstate/blame/master/test/e2e/machineconfigs.yaml#L21

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -19,6 +19,9 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
         	  ./cluster/cli.sh ssh $node -- sudo nmcli con del $uuid
 	      fi
     done
+    for nic in $(nmcli --fields NAME,UUID -t con show | grep 'Wired Connection' | awk -F : '{print $2}'); do
+        nmcli con modify $nic match.interface-name "!$FIRST_SECONDARY_NIC, !$SECOND_SECONDARY_NIC";
+    done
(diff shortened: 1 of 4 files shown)


cluster/up.sh
```diff
@@ -19,6 +19,9 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
         	  ./cluster/cli.sh ssh $node -- sudo nmcli con del $uuid
 	      fi
     done
+    for nic in $(nmcli --fields NAME,UUID -t con show | grep 'Wired Connection' | awk -F : '{print $2}'); do
```
cybertron, line 22: I think this may need to be done on each reboot of a node because the in-memory Wired Connection profile is created new each time.

However, I think now there may be a better way to do this now. NM added a global config option to ignore nics. If you create a conf file that looks like this:

```[device-enp2s0]
match-device=interface-name:enp2s0
keep-configuration=no
allowed-connections=except:origin:nm-initrd-generator```

It should prevent NM from auto-configuring that interface.

qinqon: The wirde connection is not in-memory, NetworkManager persiste those connections I think, in fact CI is fine so I think we can merge it.

PR #1184, opened 2023-05-04
deps: Bump x/text to 0.9.0

mkowalski:

/kind bug

Bumping x/text dependency so that we get updated x/net. The latter is to address [GO-2023-1571](https://pkg.go.dev/vuln/GO-2023-1571)

The CVE is about HTTP/2 streams sent to the application. I'm aware that at this moment there is no direct exposure in this codebase, so if you feel like you want to drop this change just let me know and I'll simply drop this PR.

**Release note**:

```release-note
NONE
```


qinqon:
/retest
```
net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

PR #1191, opened 2023-06-07
Use ginkgo-2.1.5

cybertron:
Looking to pull in the summary fix from
https://github.com/onsi/ginkgo/commit/5cf23e264b07a4b6928b02798e594205db9eac89 so we always get summaries, even if there's only a single failure.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


cybertron:
/cc @dougsland 

This may help us track down repeated failures in ci jobs. This way you can just check the summary to see if the same tests are failing each time.

PR #1192, opened 2023-06-15
probe: Add support for systems without IPv4 gateway

mkowalski:

/kind enhancement

The current code assumes that there is always a default gateway using
IPv4 stack. Because of this, ping check will never succeed on a system
without one, e.g. single-stack IPv6.

This patch changes the behaviour so that we will use either of IPv4 or
IPv6 gateways effectively enabling IPv6-only systems.

**Release note**:

```release-note
Fix the probe check when running on a system without IPv4 default gateway (e.g. IPv6-only).
```

### DIFF
--- a/pkg/probe/probes.go
+++ b/pkg/probe/probes.go
@@ -51,6 +51,11 @@ type Probe struct {
 	condition func(client.Client, time.Duration) wait.ConditionFunc
 }
 
+type Route struct {
+	nextHop net.IP
+	iface   string
+}
+
 const (
 	defaultGwRetrieveTimeout  = 120 * time.Second
 	defaultGwProbeTimeout     = 120 * time.Second
@@ -135,20 +140,22 @@ func checkNodeReadiness(cli client.Client) (bool, error) {
 	return false, nil
 }
 
-func defaultGw(currentState gjson.Result) (string, error) {
-	var found gjson.Result
+func defaultGw(currentState gjson.Result) (Route, error) {
+	var found Route
 	currentState.Get("routes.running").ForEach(
 		func(_, v gjson.Result) bool {
 			// we want to pick the next hop related to the "main" table because we may have multiple tables
-			if v.Get("destination").String() == "0.0.0.0/0" && v.Get("table-id").Int() == mainRoutingTableID {
-				found = v.Get("next-hop-address")
+			if (v.Get("destination").String() == "0.0.0.0/0" || v.Get("destination").String() == "::/0") &&
+				v.Get("table-id").Int() == mainRoutingTableID {
+				found.nextHop = net.ParseIP(v.Get("next-hop-address").String())
+				found.iface = v.Get("next-hop-interface").String()
 				return false
 			}
 			return true
 		},
 	)
 
-	if !found.Exists() {
+	if found.nextHop == nil {
 		msg := "default gw missing"
 		defaultGwLog := log.WithValues("path", "routes.running.next-hop-address", "table-id", mainRoutingTableID)
 		defaultGwLogDebug := defaultGwLog.V(1)
@@ -157,9 +164,9 @@ func defaultGw(currentState gjson.Result) (string, error) {
 		} else {
 			defaultGwLog.Info(msg)
 		}
-		return "", errors.New(msg)
+		return Route{}, errors.New(msg)
 	}
-	return found.String(), nil
+	return found, nil
 }
 
 func pingCondition(cli client.Client, timeout time.Duration) wait.ConditionFunc {
@@ -188,8 +195,14 @@ func runPing(_ client.Client) (bool, error) {
 	return true, nil
 }
 
-func ping(target string) (string, error) {
-	cmd := exec.Command("ping", "-c", "1", target)
+func ping(target Route) (string, error) {
+	// If next hop is IPv6 link-local, we need to append an interface otherwise it is
+	// not clear which interface should be used for communication (e.g. ping test).
+	// As this syntax works always, we simply append it always.
+	//
+	// It is safe to ignore gosec error about concatenated strings as the Route struct
+	// is not directly taking any user input.
+	cmd := exec.Command("ping", "-I", target.iface, "-c", "1", target.nextHop.String()) // #nosec G204
 	var outputBuffer bytes.Buffer
 	cmd.Stdout = &outputBuffer
 	cmd.Stderr = &outputBuffer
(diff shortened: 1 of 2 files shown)


pkg/probe/probes.go
```diff
@@ -136,19 +136,26 @@ func checkNodeReadiness(cli client.Client) (bool, error) {
...
-				found = v.Get("next-hop-address")
+			if (v.Get("destination").String() == "0.0.0.0/0" || v.Get("destination").String() == "::/0") && v.Get("table-id").Int() == mainRoutingTableID {
+				// if next hop is IPv6 link-local, we need to append an interface otherwise it is
+				// not clear which interface should be used for communication (e.g. ping test)
+				found = v.Get("next-hop-address").String()
+				ip := net.ParseIP(found)
+				if ip.To4() == nil && ip.IsLinkLocalUnicast() {
+					iface := v.Get("next-hop-interface").String()
+					found = found + "%" + iface
+				}
```
mkowalski, line 151: To show clearly what I mean, please look at the following ping dumps

```
# ping fe80::b6fb:e4ff:fe51:782d
ping: Warning: IPv6 link-local address on ICMP datagram socket may require ifname or scope-id => use: address%<ifname|scope-id>
PING fe80::b6fb:e4ff:fe51:782d(fe80::b6fb:e4ff:fe51:782d) 56 data bytes
^C
--- fe80::b6fb:e4ff:fe51:782d ping statistics ---
4 packets transmitted, 0 received, 100% packet loss, time 3059ms
```

```
# ping fe80::b6fb:e4ff:fe51:782d%ens192
PING fe80::b6fb:e4ff:fe51:782d%ens192(fe80::b6fb:e4ff:fe51:782d%ens192) 56 data bytes
64 bytes from fe80::b6fb:e4ff:fe51:782d%ens192: icmp_seq=1 ttl=64 time=0.934 ms
64 bytes from fe80::b6fb:e4ff:fe51:782d%ens192: icmp_seq=2 ttl=64 time=0.772 ms
^C
--- fe80::b6fb:e4ff:fe51:782d%ens192 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1001ms
rtt min/avg/max/mdev = 0.772/0.853/0.934/0.081 ms
```

pkg/probe/probes.go
```diff
@@ -136,19 +136,27 @@ func checkNodeReadiness(cli client.Client) (bool, error) {
...
+			if (v.Get("destination").String() == "0.0.0.0/0" || v.Get("destination").String() == "::/0") &&
+				v.Get("table-id").Int() == mainRoutingTableID {
+				// if next hop is IPv6 link-local, we need to append an interface otherwise it is
+				// not clear which interface should be used for communication (e.g. ping test)
+				found = v.Get("next-hop-address").String()
+				ip := net.ParseIP(found)
+				if ip.To4() == nil && ip.IsLinkLocalUnicast() {
+					iface := v.Get("next-hop-interface").String()
+					found = found + "%" + iface
+				}
```
qinqon, line 152: If this works for ipv4 + ipv6 not link-local we can remove the condition to simplify code

```suggestion
				found = v.Get("next-hop-address").String()
				ip := net.ParseIP(found)
				iface := v.Get("next-hop-interface").String()
				found = found + "%" + iface
				
```

Also an extra-mile would be to return the "Route" struct instead of a string and consume it at `runPing` so the `found + "%" + iface`  composition is done there, I think is something specific of `ping` command

qinqon (changes requested):
Just nits

pkg/probe/probes.go
mkowalski: That's a good point, strings are ugly so I redesigned it to a struct `Route` that holds IP address and interface

pkg/probe/probes.go
```diff
@@ -188,8 +195,16 @@ func runPing(_ client.Client) (bool, error) {
...
 }
 
-func ping(target string) (string, error) {
-	cmd := exec.Command("ping", "-c", "1", target)
+func ping(target Route) (string, error) {
+	var dest string
+	// if next hop is IPv6 link-local, we need to append an interface otherwise it is
+	// not clear which interface should be used for communication (e.g. ping test)
+	dest = target.nextHop.String()
+	if target.nextHop.To4() == nil && target.nextHop.IsLinkLocalUnicast() {
```
cybertron, line 203: Just a quick followup on Quique's previous comment, it looks like ping with the interface does work fine on ipv4 too. Do we need this logic?

cybertron (commented):
/lgtm

One small question inline, but I don't think it has a huge effect on this patch. I'd be fine with it going in as-is.

pkg/probe/probes.go
mkowalski: LOL I completely missed this, I somehow had imprinted in my memory that it is only for link-local and nothing else. As it indeed works, I am removing this logic

pkg/probe/probes.go
```diff
@@ -188,8 +195,13 @@ func runPing(_ client.Client) (bool, error) {
 	return true, nil
 }
 
-func ping(target string) (string, error) {
-	cmd := exec.Command("ping", "-c", "1", target)
+func ping(target Route) (string, error) {
+	// If next hop is IPv6 link-local, we need to append an interface otherwise it is
+	// not clear which interface should be used for communication (e.g. ping test).
+	// As this syntax works always, we simply append it always.
+	dest := target.nextHop.String() + target.iface
```
cybertron, line 202: Did we lose the % separator here or is it included in one of those values?

mkowalski: Of course we did. I was wondering why all the tests failed and I was just starting to unfold their cryptic failure messages

pkg/probe/probes.go
mkowalski: Well apparently my memory was correct. After reading deeper (e.g. https://www.rfc-editor.org/rfc/rfc4007#section-11) the syntax with `%` is only correct for link-local IPv6. It may accidentally have worked for us when we tested manually, but in principle it does not work, i.e.

```
{"level":"error","ts":"2023-06-23T16:45:05.060Z","logger":"probe","msg":"error pinging default gateway -> output: ''","error":"failed running ping probe: cmd output: 'ping: 192.168.66.2%bond87: Name or service not known
': exit status 2","errorVerbose":"exit status 2
```

OTOH `ping -I interface address` works for IPv6 link-local too, so may be better if I move to this format

PR #1204, opened 2023-09-14
handler: Remove iface name normalization

qinqon:
nmstate 2.y.z renders correctly the interface name and add a single quotes if the name is not an string so marshaling it always result as an string. This change remove the code that normalize interface names for those scenarios since it's not longer needed.

**Release note**:

```release-note
Remove interface name normalization.
```


mkowalski (approved):
Always for removing unnecessary code, thanks!

mkowalski (approved):

mkowalski (approved):

mkowalski (approved):

PR #1206, opened 2023-09-19
deps: Bump go 1.19 and some deps

qinqon:
Bump to:
- golang 1.19
- controller-runtime 0.14.6
- k8s 1.26.3

There are changes at the code to fix linting, since there is new stuff from go1.19
There is no proper image for go 1.20 at quay.io so we stick with 1.19 for now
The k8s version is the oldest one that supports controller-runtime 0.14.6
The controller-runtime is the minimal we need for kube-addmission-webhook.

**Release note**:

```release-note
Bump: go 1.19, controller-runtime v0.14.6 and k8s 1.26.3
```


mkowalski:
/approve

Wow what a monster, thanks! Let me just see green CI and let's ship it

qinqon:
There is a proper failure after fixing the lints, looks like `PollUntilContextTimeout` return the error but `PollImmediate` was not returning the timeout error

At least e2e tests is fine.

PR #1212, opened 2023-10-19
Upgrade controller-runtime to v0.15

mkowalski:

/kind bug

We are upgrading controller-runtime to v0.15.2; because of breaking changes between v0.14 and v0.15, additional changes are needed in the way we use cache

**Release note**:

```release-note
NONE
```


mkowalski:
/hold

For some reason daemon set is missing. I feel like it needs manual test

PR #1215, opened 2023-10-24
handler: Retry enactment update

qinqon:

/kind bug

At some scenarios it takes some retries to update the enactment status if the network is not totally settle. This change add some retries to enactment status update.

```release-note
Retry enactment status update on failure.
```


pkg/enactmentstatus/status.go
```diff
@@ -34,12 +34,13 @@ import (
...
 var (
-	log = logf.Log.WithName("enactmentstatus")
+	log       = logf.Log.WithName("enactmentstatus")
+	allErrors = func(error) bool { return true }
 )
 
 func Update(cli client.Client, key types.NamespacedName, statusSetter func(*nmstate.NodeNetworkConfigurationEnactmentStatus)) error {
 	logger := log.WithValues("enactment", key.Name)
-	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	return retry.OnError(retry.DefaultRetry, allErrors, func() error {
```
jcaamano, line 43: I would add a comment explaining how temporary network unavailability needs to be accounted for when handling errors. Is this the only place where this problem can happen? Or is more generic handling needed? Isn't connectivity verified before this update is attempted? Or is this when we try to update after rollback?

mkowalski: Hey @jcaamano, AFAIR you were working on some bug when this was discovered. Do you have a Jira handle? We would simply link it in the commit so that whoever is interested at any time would `git blame` and read what's needed

jcaamano: This was something @qinqon tried out in context of https://issues.redhat.com//browse/OPNET-268

qinqon: @jcaamano @mkowalski added some comments to explain it.

PR #1228, opened 2024-02-21
Log setup steps when starting handler

mkowalski:
We have observed that under specific conditions starting handler may hang indefinitely. In order to help debugging those issues, we are adding some logging to report steps happening when handler starts. Thanks to this, whenever a process hands, we will see which component is not responding.

/kind enhancement

**Release note**:

```release-note
NONE
```


mkowalski:
This comes up as part of https://issues.redhat.com/browse/OCPBUGS-29266

PR #1232, opened 2024-04-03
lint: Use binary golangci-lint

qinqon:
There are some problems at the current golangci-lint version that can only be fixed bumping to new version that depends on new golang.

This change use directly the linter binary instead of "go run" so we don't have the deps problems.

```release-note
NONE
```


qinqon:
/override pull-kubernetes-nmstate-e2e-handler-k8s pull-kubernetes-nmstate-e2e-operator-k8s

PR #1237, opened 2024-04-29
api: add ovn bridge mappings to node network state

maiqueb:

>
/kind bug
> /kind enhancement

This PR exposes the node's ovn bridge mappings (already provided on `nmstatectl show`) via the NNS object.

**Release note**:

```release-note
Display OVN bridge mappings on the NodeNetworkState object
```

### DIFF
--- a/pkg/state/type.go
+++ b/pkg/state/type.go
@@ -28,6 +28,7 @@ type rootState struct {
 	Interfaces  []interfaceState `json:"interfaces"             yaml:"interfaces"`
 	Routes      *routes          `json:"routes,omitempty"       yaml:"routes,omitempty"`
 	DNSResolver *dnsResolver     `json:"dns-resolver,omitempty" yaml:"dns-resolver,omitempty"`
+	Ovn         *bridgeMappings  `json:"ovn,omitempty"          yaml:"ovn,omitempty"`
 }
 
 type routes struct {
@@ -101,4 +102,14 @@ func (r *routeState) UnmarshalJSON(b []byte) error {
 	r.Data["next-hop-interface"] = fields.NextHopInterface
 	r.routeFields = fields
 	return nil
+}
+
+type bridgeMappings struct {
+	PhysicalNetworkMappings []PhysicalNetworks `json:"bridge-mappings,omitempty" yaml:"bridge-mappings,omitempty"`
+}
+
+type PhysicalNetworks struct {
+	Name   string `json:"localnet" yaml:"localnet"`
+	Bridge string `json:"bridge" yaml:"bridge"`
+	State  string `json:"state,omitempty" yaml:"state,omitempty"`
 }
--- a/pkg/state/ovn_bridge_mappings_test.go
+++ b/pkg/state/ovn_bridge_mappings_test.go
@@ -1,0 +1,140 @@
+/*
+Copyright The Kubernetes NMState Authors.
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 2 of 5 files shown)


pkg/state/type.go
```diff
@@ -102,3 +103,13 @@ func (r *routeState) UnmarshalJSON(b []byte) error {
...
+
+type bridgeMappings struct {
+	PhysicalNetworkMappings []PhysicalNetworks `json:"bridge-mappings,omitempty" yaml:"bridge-mappings,omitempty"`
+}
+
+type PhysicalNetworks struct {
+	Name   string `json:"localnet" yaml:"localnet"`
+	Bridge string `json:"bridge" yaml:"bridge"`
+	State  string `json:"state,omitempty" yaml:"state,omitempty"`
+}
```
qinqon, line 115: This is not really needed since we don't need to filter it out, we just need an "ovn" field at desiredState level of type `json.RawMessage`

pkg/state/type.go
```diff
@@ -28,6 +28,7 @@ type rootState struct {
 	Interfaces  []interfaceState `json:"interfaces"             yaml:"interfaces"`
 	Routes      *routes          `json:"routes,omitempty"       yaml:"routes,omitempty"`
 	DNSResolver *dnsResolver     `json:"dns-resolver,omitempty" yaml:"dns-resolver,omitempty"`
+	Ovn         *bridgeMappings  `json:"ovn,omitempty"          yaml:"ovn,omitempty"`
```
qinqon, line 31:
```suggestion
	Ovn         json.RawMessage  `json:"ovn,omitempty"          yaml:"ovn,omitempty"`
```

qinqon: Now that we are a it maybe we can add the missing ones looking at https://docs.rs/nmstate/latest/nmstate/struct.NetworkState.html

We are missing:
- hostname
- route-rules
- ovs-db

So we can do the following
```golang
type rootState struct {
	Interfaces  []interfaceState `json:"interfaces"             yaml:"interfaces"`
	Routes      *routes          `json:"routes,omitempty"       yaml:"routes,omitempty"`
	DNSResolver *dnsResolver     `json:"dns-resolver,omitempty" yaml:"dns-resolver,omitempty"`
        Hostname         json.RawMessage  `json:"hostname,omitempty"          yaml:"ovn,omitempty"`
        RouteRules         json.RawMessage  `json:"route-rules,omitempty"          yaml:"ovn,omitempty"`
        OVSDb         json.RawMessage  `json:"ovs-db,omitempty"          yaml:"ovs-db,omitempty"`
        OVN         json.RawMessage  `json:"ovn,omitempty"          yaml:"ovn,omitempty"`
}
```

qinqon: An alternative is a custom encoder that will just bypass whatever we don't need to filter

qinqon (changes requested):
Let's generalize this PR and add the missing top level field with a transparent data type `json.RawMessage`.

pkg/state/type.go
maiqueb: It's not only filtering - it makes testing a lot easier, since I can use these structs to assert against. 

I prefer to 
```go
Expect(nodeBridgeMappings(node)).To(
				ConsistOf(state.PhysicalNetworks{Name: networkName, Bridge: bridgeName}))
```

instead of comparing against a string. 

But I won't insist.

pkg/state/type.go
maiqueb: I'd rather focus exclusively on the attributes that I need to add. And test only those attributes.

pkg/state/type.go
qinqon: We will consume a generated golang api in the furure so I am ok with this.

qinqon (approved):
/lgtm
/approve

maiqueb:
hm, I'm checking this locally and it passes.

Could it be we need to somehow integrate OVS in CI for this to work ?

maiqueb:
@qinqon ok, so this got merged.

When / how will it become available downstream ? What's the flow here ? ... Will it be automatically cherry-picked downstream by some automation, or do I need to do something further ? 

Thanks for your help and time.

mkowalski:
It will land in downstream automatically after _a few days_. Unless you have a critical case and need it in downstream immediately, there is nothing to do at all, just sit and watch

PR #1239, opened 2024-05-06
e2e: parametrize monitoring namespace

mkowalski:
This PR adds ability to manually configure monitoring namespace by using `MONITORING_NAMESPACE` env variable passed to Makefile, e.g. when executing

```
make test-e2e-handler
```

With this change we can run e2e tests in OpenShift clusters which use a different name than our hardcoded `monitoring`.

**Release note**:

```release-note
NONE
```


qinqon (approved):
/lgtm
/approve

PR #1248, opened 2024-06-14
e2e: don't require empty starting state for bridge mappings

mkowalski:
This commit changes the behaviour of the OVN bridge mapping test so that we do not require the initial state to be empty. Thanks to this we make e2e tests working correctly on machines whre OVN is already used and bridges are deployed.

Without this change the test will pass only if no OVN bridge mappings exist on the node. The assumption is quite strong and not really desired.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/nns_ovn.go
+++ b/test/e2e/handler/nns_ovn.go
@@ -36,7 +36,8 @@ var _ = Describe("[nns] NNS OVN bridge mappings", func() {
 
 	BeforeEach(func() {
 		for _, node := range nodes {
-			Expect(nodeBridgeMappings(node)).To(BeEmpty())
+			Expect(nodeBridgeMappings(node)).NotTo(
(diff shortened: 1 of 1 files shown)


test/e2e/handler/nns_ovn.go
```diff
@@ -35,10 +35,6 @@ var _ = Describe("[nns] NNS OVN bridge mappings", func() {
 	)
 
 	BeforeEach(func() {
-		for _, node := range nodes {
-			Expect(nodeBridgeMappings(node)).To(BeEmpty())
```
qinqon, line 39: Can we check here the opposite ? 
```golang
NotTo(ContainElement(state.PhysicalNetworks{Name: networkName, Bridge: bridgeName})))
```

qinqon (changes requested):

test/e2e/handler/nns_ovn.go
mkowalski: Yes, done

PR #1263, opened 2024-09-06
Openshift replace cert manager with openshift cert service

qinqon:
Use cert services at openshift cluster's instead of nmstate-cert-manager

TODO:
- [ ] RBAC cleanup

```release-note
Use cert services at openshift cluster's instead of nmstate-cert-manager
```


mkowalski:
/test pull-kubernetes-nmstate-e2e-operator-k8s

Maybe a flake, maybe not
```
Message: "daemonsets.apps \"nmstate-handler\" not found",
```

PR #1267, opened 2024-09-29
Use `nmstatectl apply` instead of `nmstatectl set`

cathay4t:
The `nmstatectl set` is deprecated in order to align with API method
name (`apply`). Should use `nmstatectl apply` instead.


qinqon:
/retest
/lgtm
/approve

Thanks @cathay4t

PR #1268, opened 2024-09-30
openshift: Ensure new webhook secret

qinqon:
The non openshift secret is not compatible with openshift generated version, this change a different one so they don't collide and also remove the old one.

```release-note
Use different webhook secret name for openshift.
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -181,7 +181,11 @@ spec:
       volumes:
         - name: tls-key-pair
           secret:
+{{- if not .IsOpenShift }}
             secretName: {{template "handlerPrefix" .}}nmstate-webhook
+{{- else }}
+            secretName: {{template "handlerPrefix" .}}openshift-nmstate-webhook
+{{- end }}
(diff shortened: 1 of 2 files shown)


deploy/handler/operator.yaml
```diff
@@ -391,7 +395,7 @@ metadata:
   name: {{template "handlerPrefix" .}}nmstate-webhook
   namespace: {{ .HandlerNamespace }}
   annotations:
-    service.beta.openshift.io/serving-cert-secret-name: {{template "handlerPrefix" .}}nmstate-webhook
+    service.beta.openshift.io/serving-cert-secret-name: {{template "handlerPrefix" .}}openshift-nmstate-webhook
```
mkowalski, line 398: This line also around `{{- if not .IsOpenShift }}`, I guess

qinqon: Not needed, this annotation is only used by openshift, so it's ok to have it there like this upstream.

mkowalski (approved):

PR #1277, opened 2024-12-19
nmpolicy: Use nmstatectl policy instead of go lib

qinqon:
The nmpolicies engine has being ported to nmstatectl policy, this change
replace the golang library by the nmstatectl policy so there is only one
implementation, later on the golang library will be archived.

Also this unblock support for numbers matching

**Release note**:

```release-note
Replace golang nmpolicy lib with nmstatectl policy command.
```


mkowalski:
/lgtm

We tested this change downstream (https://github.com/openshift/kubernetes-nmstate/pull/548) and it passes, so good to go there too.

PR #1279, opened 2025-01-07
e2e, handler: Simplify vlan_filtering 0 check

qinqon:
Looks like when disabling vlan filtering the bridge vlan show command is not working as expected but the system is correctly working, this change just check tha vlan filtering is 0 at the bridge and that the default pvid is 1, that's more than enough.

```release-note
NONE
```


rhrazdil (approved):

qinqon:
/override pull-kubernetes-nmstate-e2e-handler-k8s

Not related statistics job failed, we will have to look into it.

PR #1280, opened 2025-01-09
OCPBUGS-44596: Mount OVS directory instead of socket file

mkowalski:
When mounting the OVS DB socket directly as a file (`/run/openvswitch/db.sock`) instead of mounting the whole directory, handler container loses access to the OVS DB in case of OVS daemon restart on the host. This can only be fixed by restarting the container so that it mounts again the correct file.

To remediate this, we are mounting the whole `/run/openvswitch` directory so that DB connection can be gracefully restored by the handler itself.

Closes: OCPBUGS-44596

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -366,7 +366,7 @@ spec:
             - name: nmstate-lock
               mountPath: /var/k8s_nmstate
             - name: ovs-socket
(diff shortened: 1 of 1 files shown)


mkowalski:
/retest-required

Metric failed. Not sure if for real or not, but should not be related to this change

Issue #831, opened 2021-09-15
Add knobs to NNCP to tune the connectivity probes


qinqon:
**What happened**:
If the default gw want to be explicitly removed or ICMP is not working at next-hop-address the probes are not going to work as expected.

**What you expected to happen**:
Probes to run when needed towards what user want to check.

We need to add some knobs to NNCP to activate/deactivate ping and dns proba and also set the address to ping or resolv.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- `NodeNetworkState` on affected nodes (use `kubectl get nodenetworkstate <node_name> -o yaml`):
- Problematic `NodeNetworkConfigurationPolicy`:
- kubernetes-nmstate image (use `kubectl get pods --all-namespaces -l app=kubernetes-nmstate -o jsonpath='{.items[0].spec.containers[0].image}'`):
- NetworkManager version (use `nmcli --version`)
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Others:

mkowalski:
I did my research and clusters without default gateway should be 1st class citizens already. And idea to configure target IP for the ICMP probe seems like a good one. An idea to deactivate the probe seems like not a good one -- I like the fact that k-nmstate is the guardian of healthiness. If you don't want this feature, maybe you don't need k-nmstate in the end (as a workaround you can always ping 127.0.0.1 for a positive result)

PR #1281, opened 2025-01-09
design-proposal: dns probe name customize

qinqon:
At some users customizing globally the dns probe name to resolve is a must, this change add a design doc to do it at NMState CR.

Related but for default gw:
- https://github.com/nmstate/kubernetes-nmstate/issues/831

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design-proposals/customize-dns-probe-globally.md
+++ b/design-proposals/customize-dns-probe-globally.md
@@ -1,0 +1,72 @@
+# Globally customize DNS cluster health probe.
+
+## Summary
+
+Support customize kubernetes-nmstate DNS resolving probes at NMState CR.
+
+## Motivation
+
+At some cluster their network configuration is not compatible with DNS health
+probe the kubernetes-nmstate cluster health probes, so the NNCPs will fail.
+
+### User Stories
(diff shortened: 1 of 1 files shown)


design-proposals/customize-dns-probe-globally.md
```diff
@@ -0,0 +1,62 @@
+# Globally customize DNS cluster health probe.
+
+## Summary
+
+Support customize kubernetes-nmstate DNS resolving probes at NMState CR.
+
+## Motivation
```
phoracek, line 7: Please also list the alternatives in this document. e.g. making the probe smarter, and having this configured on NNCP-level

phoracek (changes requested):

design-proposals/customize-dns-probe-globally.md
qinqon: I will add an alternative about probe doing fallaback to a global name, about NNCP is stated as non goal and if needed can be done a new enhancement, we want this one to be very specific to reduce the number of knobs that we really need.

PR #1282, opened 2025-01-11
Added helm package oci registry support

supu2:
I've added helm package and oci registry support. People who don't want to use olm they can use helm.
/kind enhancement


qinqon (changes requested):

.github/workflows/release.yml
```diff
@@ -43,3 +46,19 @@ jobs:
...
+      uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Template Helm charts
+      run: |
+        make manifests
+        helm dependency update ./${{ env.CHART_DIR }}
+        helm package ./${{ env.CHART_DIR }} --destination ./charts --app-version ${{ steps.bump-semver.outputs.new_version }} --version ${{ steps.bump-semver.outputs.new_version }}
```
supu2, line 63: @qinqon  The version of chart and application verions override in that line.

deploy/helm/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: kubernetes-nmstate
+description: Declarative node network configuration driven through Kubernetes API.
+type: application
+version: 0.1.0
```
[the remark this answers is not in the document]
supu2: @qinqon  https://github.com/nmstate/kubernetes-nmstate/pull/1282/files#r1929588938

deploy/helm/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
```
mkowalski, line 1: A bunch of nits for the manifest. If we are going to build&push it to the public, let's have metadata complete.

Comparing with e.g. https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/Chart.yaml and looking at our current https://github.com/nmstate/kubernetes-nmstate/blob/main/deploy/bases/kubernetes-nmstate-operator.clusterserviceversion.yaml, what we can easily add is

* home
* icon
* keywords
* kubeVersion
* maintainers
* sources

deploy/helm/Chart.yaml
```diff
@@ -0,0 +1,6 @@
...
+name: kubernetes-nmstate
+description: Declarative node network configuration driven through Kubernetes API.
+type: application
+version: 0.1.0
+appVersion: "1.16.0"
```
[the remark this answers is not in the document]
mkowalski: If this is going to be overridden, let's have a value that does not resemble any real one, e.g. `0.0.1` or `99.99.99`. With `1.16.0` it's actually confusing to see at first sight that this value won't be preserved

deploy/helm/Chart.yaml
supu2: @mkowalski  updated chart file. I'm not sure kubeversion support  I hope nmstate works on  1.21.0

deploy/helm/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+home: https://github.com/nmstate/kubernetes-nmstate
+icon: https://nmstate.io/kubernetes-nmstate/assets/images/logo/fullcolor.png
+keywords:
+- nmstate
+- networking
+- NetworkManager
+maintainers:
+- email: ellorent@redhat.com
+  name: Enrique Llorente Pastora
+- email: phoracek@redhat.com
```
phoracek, line 16: @mkowalski would you mind if you were listed instead of me? It would better reflect the reality

mkowalski: This will happen. It's just that it's not the only place where we should remove you (e.g. there is also the OLM manifest) so I would merge this one as-is and then I will handle the metadata update myself

deploy/helm/templates/nmstate.io_v1_nmstate_cr.yaml
```diff
@@ -0,0 +1 @@
+../../../build/_output/manifests/nmstate.io_v1_nmstate_cr.yaml
```
mkowalski, line 1: Those are all dangling symlinks, i.e. when you just `git clone` they will not point to existing files, not even to the directory structure.

I would rather keep `deploy/` in the repository clean and have something like

```
mkdir -p deploy/helm/templates
cp ../../../build/_output/manifests/* deploy/helm/templates/
```
inside the pipeline, so after `make manifests`.

deploy/helm/Chart.yaml
phoracek: works for me

deploy/helm/templates/nmstate.io_v1_nmstate_cr.yaml
qinqon: @supu2 can you address this comments, if not we will close this PR.

qinqon:
@supu2 since we don't see the comments addressed we are going to close this, feel free to open it again in case you fix the comments.

PR #1285, opened 2025-01-17
test: add 'notsecret' keyword to the Bearer token

mkowalski:
Test platform scans all the logs and removes files that contain potential secrets. As bearer token is technically a secret, we are losing all the test logs.

This change adds `# notsecret` to the curl call so it will be ignored by the scanner.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


mkowalski:
Tested via https://github.com/openshift/kubernetes-nmstate/pull/550 and I am happy with the result

PR #1346, opened 2025-06-20
handler: sort NNCPs alphanumerically on requeue

mkowalski:
With this change, whenever there is a need to requeue existing NNCPs for reconciliation, instead of relying on their random order, they will get sorted alphanumerically.

This change is introduced to make reconciliation more deterministic. It should help users that create multiple NNCPs affecting the same node what is discouraged on its own, but happens.

Mandatory disclaimer - this is the first PR fully created using Cursor AI. It has been reviewed by human, but no code has been manually touched.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
When requeueing NNCPs, they will now be applied in the alphanumerical order instead of the random order.
```

## Summary by Sourcery

Sort NodeNetworkConfigurationPolicy requeue requests alphanumerically to ensure deterministic reconciliation

New Features:
- Apply alphanumeric sort on NNCP names when requeuing policies

Enhancements:
- Extract `allPolicies` mapping logic into its own method for clarity

Tests:
- Add unit tests to verify sorted order of policies
- Add unit test to assert empty result when no policies exist

### DIFF
--- a/controllers/handler/nodenetworkconfigurationpolicy_controller.go
+++ b/controllers/handler/nodenetworkconfigurationpolicy_controller.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -244,25 +245,7 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(_ context.Context, 
 }
 
 func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	allPolicies := handler.TypedMapFunc[*corev1.Node, reconcile.Request](
-		func(context.Context, *corev1.Node) []reconcile.Request {
-			log := r.Log.WithName("allPolicies")
-			allPoliciesAsRequest := []reconcile.Request{}
-			policyList := nmstatev1.NodeNetworkConfigurationPolicyList{}
-			err := r.Client.List(context.TODO(), &policyList)
-			if err != nil {
-				log.Error(err, "failed listing all NodeNetworkConfigurationPolicies to re-reconcile them after node created or updated")
-				return []reconcile.Request{}
-			}
-			for policyIndex := range policyList.Items {
-				policy := policyList.Items[policyIndex]
-				allPoliciesAsRequest = append(allPoliciesAsRequest, reconcile.Request{
-					NamespacedName: types.NamespacedName{
-						Name: policy.Name,
-					}})
-			}
-			return allPoliciesAsRequest
-		})
+	allPoliciesFunc := allPolicies(r.Client, r.Log)
 
 	// Reconcile NNCP if they are created/updated/deleted or
 	// Node is updated (for example labels are changed), node creation event
@@ -290,7 +273,7 @@ func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Man
 		source.Kind(
 			mgr.GetCache(),
 			&corev1.Node{},
-			handler.TypedEnqueueRequestsFromMapFunc[*corev1.Node](allPolicies),
+			handler.TypedEnqueueRequestsFromMapFunc[*corev1.Node](allPoliciesFunc),
 			onLabelsUpdatedForThisNode,
 		),
 	)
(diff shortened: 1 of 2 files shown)


controllers/handler/nodenetworkconfigurationpolicy_controller_test.go
```diff
@@ -215,4 +215,112 @@ var _ = Describe("NodeNetworkConfigurationPolicy controller predicates", func()
...
+				"10-numeric-policy",
+				"alpha-policy",
+				"beta-policy",
+				"zebra-policy",
+			}
+
+			Expect(requests).To(HaveLen(len(expectedOrder)))
+			for i, expectedName := range expectedOrder {
+				Expect(requests[i].NamespacedName.Name).To(Equal(expectedName))
+			}
```
qinqon: Let's use gomega WithTransform instead of iterating, so we see the full slice at errors instead of individual.

controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -243,8 +244,8 @@ func (r *NodeNetworkConfigurationPolicyReconciler) Reconcile(_ context.Context,
 	return ctrl.Result{}, nil
 }
 
-func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	allPolicies := handler.TypedMapFunc[*corev1.Node, reconcile.Request](
+func (r *NodeNetworkConfigurationPolicyReconciler) allPolicies() handler.TypedMapFunc[*corev1.Node, reconcile.Request] {
+	return handler.TypedMapFunc[*corev1.Node, reconcile.Request](
```
qinqon: Why we conert all policies into a method ? there is no dependency on the reonciler

mkowalski: I think actually there is. So if we keep the old

```
func (r *NodeNetworkConfigurationPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
	allPolicies := handler.TypedMapFunc[*corev1.Node, reconcile.Request](
```

then in the unit test you cannot simply use the `allPolicies` function because it exists only inside SetupWithManager(). So Cursor converted `allPolicies` into a method.

Now, why does it have to be a method of NodeNetworkConfigurationPolicyReconciler is because inside `allPolicies` we call `r.Client.List(context.TODO(), &policyList)` and if that was not a method, we would need to pass around the client.

I guess the main question/choice here is whether we want to `allPolicies` accept reconciler or accept a client. My personal view is that both of them are equally ugly. If you are more opinionated towards the other, I am happy to tell Cursor to rewrite

controllers/handler/nodenetworkconfigurationpolicy_controller_test.go
mkowalski: Done, rewritten by Cursor

controllers/handler/nodenetworkconfigurationpolicy_controller.go
mkowalski: Now, I asked the same question to Cursor because this PR is an exercise if it's as good to actually keep using it.

>Why we conert all policies into a method ? there is no dependency on the reonciler

And the answer as follows

> You're absolutely right to question this! Looking at the full implementation, the allPolicies method does have dependencies on the reconciler, but they're minimal:
> [...]
> The method uses:
> r.Log - for logging
> r.Client - for listing policies
> You're right that this could be refactored into a standalone function. Here's how it could look:
> ```
> func allPolicies(client client.Client, log logr.Logger) handler.TypedMapFunc[*corev1.Node, reconcile.Request] {
> 	return handler.TypedMapFunc[*corev1.Node, reconcile.Request](
> 		func(context.Context, *corev1.Node) []reconcile.Request {
> 			// ... implementation using the passed client and log
> 		})
> }
> ```
> Then in SetupWithManager:
> ```
> allPolicies := allPolicies(r.Client, r.Log)
> ```
> Benefits of making it standalone:
> More testable (easier to unit test without creating a full reconciler)
> Less coupling to the reconciler struct
> More explicit about dependencies
> Current approach benefits:
> Keeps all reconciler behavior encapsulated
> Clearer that it's part of the reconciler's functionality
> Consistent with other controller patterns
> Would you like me to refactor it to a standalone function? It's a valid architectural improvement that would make the code more modular and testable.

PR #1348, opened 2025-07-02
Proposal for status output when NNCP is not matching any nodes.

emy:

>
> /kind bug

/kind enhancement

This PR adds a proposal for discussion on what status we should return on `oc get nncp` in case a NNCP has not been applied to any node.

NONE

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design-proposals/status-for-nncp-not-matching-any-node.md
+++ b/design-proposals/status-for-nncp-not-matching-any-node.md
@@ -1,0 +1,192 @@
+# Status for NNCP not matching any node(s)
+
+## Summary
+
+This proposal is trying to figure out a way on how to handle the status response kubernetes-nmstate is reporting on
+`oc get nncp -oyaml` when the NNCP is not matching any node(s).
+
+## Motivation
+
+Users unfamiliar with the status reports of `oc get nncp -oyaml` might confuse the status report of `status.conditions[].type`
+(Available) and incorrectly assume that a NNCP has been properly applied to node(s) even though it is possibly not
+matching any node(s) `status.conditions[].message` (Policy does not match any node) and `status.conditions[].reason`
+(NoMatchingNode). Tickets like [these](https://issues.redhat.com/browse/OPNET-81) have already come up in the past and
+should therefore be addressed to provide a clear answer.
+
+### User Stories
+
+- As a cluster administrator, I want to be able to easily judge if an NNCP has successfully been rolled out by the status.
+
+### Goals
+
+- Clarify which `status.condition[]` list an applied NNCP not matching any node(s) should have.
+
+### Non-Goals
+
+- Changes to `status.condition[]` on anything else than applied NNCP.
+
+## Proposal
+
+### Solution 1: Add new status [completed/ignored]
+
+Adding a new status will indicate to the user that the NNCP is neither `progressing`, `degraded` nor `available`.
+That status could be called `completed` or `ignored`. `Completed` could indicate that a NNCP has applied but is not "running"
+therefore only `completed` and not `available`. `Ignored` could be used to indicate that the NNCP is valid therefore not
+`degraded` but is being ignored by all nodes
+for any reason (e.g. wrong node selectors).
+
+#### Pros
+
+- It is immediately visible to an admin that the NNCP has been applied successfully but has not affected any node.
+
+#### Cons
+
+- This proposal will add a new status. Scripts from cluster admins checking and relying on the status being (Available) will most likely
+  not work anymore and therefore need some changes.
+
+### Solution 2: Set status degraded
+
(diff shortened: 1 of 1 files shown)


design-proposals/status-for-nncp-not-matching-any-node.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Proposal
+
+### Solution 1: Add new status [completed/ignored]
+
+Adding a new status will indicate to the user that the NNCP is neither `progressing`, `degraded` nor `available`.
+That status could be called `completed` or `ignored`. `Completed` could indicate that a NNCP has applied but is not "running"
+therefore only `completed` and not `available`. `Ignored` could be used to indicate that the NNCP is valid therefore not
+`degraded` but is being ignored by all nodes
+for any reason (e.g. wrong node selectors).
```
mkowalski, line 36: I am personally against introducing concept of `Completed` next to `Available` because applying NNCP is an oneshot action. So saying "NNCP [...] is not running" is fundamentally wrong, because NNCP as a construct cannot be running, it can either apply or not apply. Think about it as an event, not a process.

`Ignored` next to `Available` and `Degraded` is much nicer though.

design-proposals/status-for-nncp-not-matching-any-node.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Pros
+
+- It is immediately visible to an admin that the NNCP has been applied successfully but has not affected any node.
+
+#### Cons
+
+- This proposal will add a new status. Scripts from cluster admins checking and relying on the status being (Available) will most likely
+  not work anymore and therefore need some changes.
```
mkowalski, line 45: Is that really a con? I mean, all that would change that NNCP which does not select any node would not be `Available` any more but rather `Ignored`. If there is a script someone uses and this script treats NNCP with no matching nodes as `Available` i.e. good, then it's a problem of this script. We can afford breaking this behaviour.

I would argue if we were modifying something that returns Error to now return OK, but the other way around is justified.

design-proposals/status-for-nncp-not-matching-any-node.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Solution 2: Set status degraded
+
+The status could be set to `degraded` to indicate the NNCP is not behaving as expected.
+
+#### Pros
```
mkowalski, line 51: Another pro is that it keeps our API simplified by not extending the list of possible states.

design-proposals/status-for-nncp-not-matching-any-node.md
```diff
@@ -0,0 +1,192 @@
...
+
+- It is immediately visible to an admin that the applied NNCP might not behave as expected.
+
+#### Cons
+
+- Setting the status to `degraded` could confuse cluster admins that rely on the status indicating an actual problem.
```
mkowalski, line 57: The assumption is always(?) that when you see something `Degraded`, you read the details. In case of NoMatchingNode you would see it immediately after doing `oc get nncp ... -o yaml`.

design-proposals/status-for-nncp-not-matching-any-node.md
```diff
@@ -0,0 +1,192 @@
...
+- It could also cause scripts of cluster admins to behave incorrectly. As they already could rely on the status being `available`
+  even though the NNCP does not match any node(s).
```
mkowalski, line 59: That is a valid point. One could be devil's advocate and say that if they currently treated NNCPs with NoMatchingNode as good ones (because they are `Available`), it's a chance for them to review those and either fix the selector, or delete NNCPs.

mkowalski:
/lgtm
/approve

Having reached lazy consensus, we are moving forward with this

design-proposals/status-for-nncp-not-matching-any-node.md
```diff
@@ -0,0 +1,192 @@
...
+
+- Clarify which `status.condition[]` list an applied NNCP not matching any node(s) should have.
+
+### Non-Goals
+
+- Changes to `status.condition[]` on anything else than applied NNCP.
+
+## Proposal
+
+### Solution 1: Add new status [completed/ignored]
```
qinqon, line 30: Let's add a Solution 2 where we just has a new Reason "NoMatchingNodes" 

So users can just do a `kubectl wait --for=condition=Available` and when it timeouts they will see that reason == NoMatchinNodes and will be clear what's the issue.

They can also wait first for progressing=False and then check avaiable=True <- if this is not the case they will see NoMatchingNodes.

btw I prefer this options since it means we don't add just another condition that complicate the conditions matrix.

qinqon (changes requested):

emy:
This feature was implemented in: #1362

PR #1349, opened 2025-07-03
netpol: use targeted policy for deny-all

mkowalski:
This commit changes the deny-all policy from targetting the whole namespace to targetting only labelled workload. This will help with cases when operator runs in a shared namespace and not its own unique.

We require this because with OLM deployment method we do not really forbid users from using a shared namespace.

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

## Summary by Sourcery

Scope the deny-all network policies to only the operator and operand workloads by using per-pod label selectors, and introduce the OPERATOR_NAMESPACE parameter across deployment configs, controller logic, and tests to support this change

New Features:
- Scope default-deny network policies to only labelled operator and operand pods rather than entire namespaces

Enhancements:
- Introduce and propagate OPERATOR_NAMESPACE for rendering targeted network policies in the reconciler

Deployment:
- Add OPERATOR_NAMESPACE environment variable to the operator deployment manifest

Tests:
- Set OPERATOR_NAMESPACE environment variable in unit tests to support targeted network policy logic

### DIFF
--- a/bundle/manifests/kubernetes-nmstate-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kubernetes-nmstate-operator.clusterserviceversion.yaml
@@ -63,7 +63,7 @@ metadata:
     categories: Networking
     certified: "false"
     containerImage: quay.io/nmstate/kubernetes-nmstate-operator:latest
-    createdAt: "2025-05-20T13:25:47Z"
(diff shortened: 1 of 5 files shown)


cybertron:
/lgtm

Looks sane, and passing CI.

PR #1358, opened 2025-08-01
Add a log level to nmstate CR

qinqon:

/kind enhancement

This PR adds conditional verbose logging support to the kubernetes-nmstate operator by introducing a new `LogLevel` field to the `NMStateSpec` API. This allows fine-grained control over log levels for debugging while maintaining clean production logs.

**Key Changes:**

- **API enhancement**: Adds `LogLevel` field with `info` (default, minimal output) and `debug` (verbose output) options
- **nmstatectl integration**: Makes `nmstatectl -vv` flag conditional based on debug mode setting  
- **Operator wiring**: Handler daemonset arguments and liveness probe commands are dynamically configured based on LogLevel
- **Runtime reconfiguration**: LogLevel changes trigger daemonset updates without requiring pod restarts or image rebuilds
- **Comprehensive testing**: Unit and e2e integration tests covering all LogLevel scenarios

**Use Cases:**
- Production environments: Use `info` level for minimal logging overhead
- Troubleshooting: Switch to `debug` level for detailed nmstate operation diagnostics  
- Development: Enable verbose logging without custom image builds

**Backward Compatibility:**
Existing deployments continue to work unchanged (defaults to `info` level). No breaking changes to existing APIs or behavior.

This implementation uses `LogLevel` with `info`/`debug` terminology instead of `Verbose` with `production`/`debug` for clearer semantics. The nmstatectl verbose flag uses `-vv` (instead of `-v`) for enhanced verbosity in debug mode, and liveness probe commands are updated accordingly.

**Release note**:

```release-note
Adds LogLevel field to NMState CR enabling runtime control of nmstate verbosity. Users can now switch between info (minimal logging) and debug (verbose logging) modes without requiring image rebuilds or pod restarts.
```

## Summary by CodeRabbit

- **New Features**
  - Configurable log level for NMState (spec.logLevel; default "info"; "debug" enables verbose). Handler arguments and readiness probes now reflect the selected log level.

- **Tests**
  - Added unit and e2e tests covering log-level behavior, handler args, and readiness-probe differences; updated test helpers for consistency.

- **Chores**
  - CLI logging flag default/help switched to "info".
  - Operator bundle timestamp refreshed.

- **Documentation**
  - Added "Enabling debug logging" to troubleshooting docs.

### DIFF
--- a/docs/user-guide/103-troubleshooting.md
+++ b/docs/user-guide/103-troubleshooting.md
@@ -141,6 +141,52 @@ it:
 kubectl delete nncp eth666
 ```
 
+## Enabling debug logging
+
+For advanced troubleshooting, you can enable verbose debug logging by configuring the `LogLevel` field in the NMState custom resource. This will provide more detailed output from nmstatectl operations, which can be helpful when diagnosing complex networking issues.
+
+### Setting debug log level
+
+To enable debug logging, update the NMState CR with `logLevel: debug`:
+
+```yaml
+apiVersion: nmstate.io/v1
+kind: NMState
+metadata:
+  name: nmstate
+spec:
+  logLevel: debug  # Valid values: "info" (default) or "debug"
+```
+
+Apply the configuration:
+
+```shell
+kubectl patch nmstate nmstate --type='merge' -p='{"spec":{"logLevel":"debug"}}'
+```
+
+### Verifying debug mode
+
+You can verify that debug mode is active by checking the handler DaemonSet pods:
+
+```shell
+kubectl logs -n nmstate -l component=kubernetes-nmstate-handler --tail=50
+```
+
+When debug mode is enabled, you should see more verbose output including:
+- Detailed nmstatectl command execution logs
+- Verbose nmstate operations with `-vv` flag
+- Additional debugging information from network configuration operations
+
+### Reverting to info log level
+
+To reduce log verbosity, revert to info log level:
+
+```shell
+kubectl patch nmstate nmstate --type='merge' -p='{"spec":{"logLevel":"info"}}'
+```
+
+**Note:** Debug logging significantly increases log volume and should primarily be used for troubleshooting purposes. Consider reverting to info level after resolving issues to maintain optimal performance.
+
 ## Continue reading
 
 This was the last article from the introduction series. You can continue reading


controllers/operator/nmstate_controller.go
```diff
@@ -319,6 +320,11 @@ func (r *NMStateReconciler) applyHandler(instance *nmstatev1.NMState) error {
 
 	probeConfig := instance.Spec.ProbeConfiguration
 
+	verboseArg := ""
+	if instance.Spec.Verbose == shared.VerboseLevelDebug {
+		verboseArg = "-vvv"
+	}
```
qinqon: This is wrong, the nmstate handler cmd verbose should be something like "-v production" while the 
readiness probe should be "-vvv" we should differenciante between the two at the template.

pkg/nmstatectl/nmstatectl.go
```diff
@@ -90,10 +96,13 @@ func Set(desiredState nmstate.State, timeout time.Duration) (string, error) {
 	var setDoneCh = make(chan struct{})
 	defer close(setDoneCh)
 
-	setOutput, err := nmstatectlWithInput(
-		[]string{"apply", "-v", "--no-commit", "--timeout", strconv.Itoa(int(timeout.Seconds()))},
-		string(desiredState.Raw),
-	)
+	args := []string{"apply"}
+	if debugMode {
+		args = append(args, "-v")
```
qinqon: Do we want to use "-v" or the more verbose option "-vvv" ?

pkg/nmstatectl/nmstatectl.go
```diff
@@ -33,8 +33,14 @@ import (
 	nmstate "github.com/nmstate/kubernetes-nmstate/api/shared"
 )
 
+var debugMode bool
+
 const nmstateCommand = "nmstatectl"
 
+func SetDebugMode(debug bool) {
+	debugMode = debug
+}
```
qinqon: I don't like global state, but doing othewise means that we have to pass a 
"nmstatectl" instance to the reconcilers

mkowalski: I agree with global here. It will help with other places where we want to consume this state. That way we won't need to refactor every single function to pass the log level

api/v1beta1/nmstate_types.go
```diff
@@ -58,6 +58,12 @@ type NMStateSpec struct {
 	// +kubebuilder:default:={}
 	// +optional
 	ProbeConfiguration NMStateProbeConfiguration `json:"probeConfiguration,omitempty"`
+	// Verbose defines the verbosity level for nmstate operations.
+	// Valid values are "production" (default, minimal output) and "debug" (verbose output for debugging).
+	// +kubebuilder:default:="production"
+	// +kubebuilder:validation:Enum=production;debug
+	// +optional
+	Verbose shared.VerboseLevel `json:"verbose,omitempty"`
```
mkowalski: Let's work on the wording here. I'd rather see `verbose: true / false` (not necessarily a boolean, this is up to you and what k8s suggests) or `level: verbose`

qinqon: so this knobs affects two things
- nmstatectl verbose level (-v, -vv, -vvv)
- handler golang trace level (debug, info, warning, error)

Maybe instead of abstract things away we need a pair of knobs:
- cliVerboseLevel(1, 2, 3)
- handlerTraceLevel(debug, info, warning, error)

mkowalski: Yes that sounds even better, to have a separate control for nmstatecli verbosity and for the regular process log levels

qinqon: well looking deeper at handler it's a lille more complicated since we have the zap logger args

```golang

// BindFlags will parse the given flagset for zap option flags and set the log options accordingly:
//   - zap-devel:
//     Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn)
//     Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
//   - zap-encoder: Zap log encoding (one of 'json' or 'console')
//   - zap-log-level: Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error',
//     or any integer value > 0 which corresponds to custom debug levels of increasing verbosity").
//   - zap-stacktrace-level: Zap Level at and above which stacktraces are captured (one of 'info', 'error' or 'panic')
//   - zap-time-encoding: Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'),
//     Defaults to 'epoch'.
func (o *Options) BindFlags(fs *flag.FlagSet) {
	// Set Development mode value
	fs.BoolVar(&o.Development, "zap-devel", o.Development,
		"Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). "+
			"Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)")

	// Set Encoder value
	var encVal encoderFlag
	encVal.setFunc = func(fromFlag NewEncoderFunc) {
		o.NewEncoder = fromFlag
	}
	fs.Var(&encVal, "zap-encoder", "Zap log encoding (one of 'json' or 'console')")

	// Set the Log Level
	var levelVal levelFlag
	levelVal.setFunc = func(fromFlag zapcore.LevelEnabler) {
		o.Level = fromFlag
	}
	fs.Var(&levelVal, "zap-log-level",
		"Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', "+
			"or any integer value > 0 which corresponds to custom debug levels of increasing verbosity")

	// Set the StrackTrace Level
	var stackVal stackTraceFlag
	stackVal.setFunc = func(fromFlag zapcore.LevelEnabler) {
		o.StacktraceLevel = fromFlag
	}
	fs.Var(&stackVal, "zap-stacktrace-level",
		"Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').")

	// Set the time encoding
	var timeEncoderVal timeEncodingFlag
	timeEncoderVal.setFunc = func(fromFlag zapcore.TimeEncoder) {
		o.TimeEncoder = fromFlag
	}
	fs.Var(&timeEncoderVal, "zap-time-encoding", "Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'.")
}
```

We are just using `zap-devel` do we really want to go beyond that ? if no then we may want to have
- handlerLogsDevelMode(true, false)
- nmstateCLIVerbose(1, 2, 3)

pkg/nmstatectl/nmstatectl_test.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func buildSetArguments(timeout time.Duration) []string {
+	args := []string{"apply"}
+	if debugMode {
+		args = append(args, "-vv")
+	}
+	args = append(args, "--no-commit", "--timeout", strconv.Itoa(int(timeout.Seconds())))
+	return args
+}
```
mkowalski: You are testing the test code here and not the real code. I.e. if I now go to `pkg/nmstatectl/nmstatectl.go` and change something, your test will still pass because you redefined `buildSetArguments()` here.

Is that intended? If yes, what is the use case?

deploy/handler/operator.yaml
```diff
@@ -437,7 +441,7 @@ spec:
               command:
               - bash
               - -c
-              - "nmstatectl show -vvv 2>&1"
+              - "nmstatectl show {{ .HandlerReadinessProbeExtraArg }} 2>&1"
```
mkowalski, line 446: Sorry I think we already discussed this in the past, but

1) Do we save output of this command anywhere? If yes, is there a chance this fails because of the size? If yes, will that mean we ~cannot start~ terminate the pod at all because this writer will fail?

2) In case of the actual issue with nmstatectl when this commands fails/errors, how feasible is it to collect the output of this probe? Does it land in the regular pod log? Or only in the k8s event log?

Basically I am trying to assess how easy and useful is to have `-vv` in the liveness probe. We most probably do want to have it, I just need to learn where it's stored and how to get it when needed

qinqon: We see that output at the kubelet logs, since it's the one running the probes

pkg/nmstatectl/nmstatectl_test.go
qinqon: Right this test does not mak any sense, will fix it.

mkowalski: This function `buildSetArguments()` is still copy-pasted into test code. Expected or not?

qinqon: Ahh right I have o fix th eother tests.

mkowalski (approved):

PR #1362, opened 2025-08-20
OPNET-81: Add Ignored status for NNCP when no node is matching the node selector.

emy:

>
> /kind bug

/kind enhancement

As per the design proposal in #1348 I have added a `Ignored` status to the NNCP.
This status is being set to ConditionTrue when there is no matching nodes for a given NNCP.
This is supposed to help admins by indicating the lack of nodes in a central point with high visibility.

**Release note**:

```release-note
The NNCP status will indicate status `Ignored` when no node is matching the node selector.
```

### DIFF
--- a/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -62,6 +62,7 @@ const (
 	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
 	NodeNetworkConfigurationPolicyConditionDegraded    ConditionType = "Degraded"
 	NodeNetworkConfigurationPolicyConditionProgressing ConditionType = "Progressing"
+	NodeNetworkConfigurationPolicyConditionIgnored     ConditionType = "Ignored"
 )
 
 var NodeNetworkConfigurationPolicyConditionTypes = [...]ConditionType{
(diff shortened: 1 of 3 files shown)


pkg/policyconditions/conditions.go
```diff
@@ -136,6 +154,12 @@ func SetPolicyFailedToConfigure(conditions *nmstate.ConditionList, message strin
 		nmstate.NodeNetworkConfigurationPolicyConditionConfigurationProgressing,
 		"",
 	)
+	conditions.Set(
+		nmstate.NodeNetworkConfigurationPolicyConditionIgnored,
+		corev1.ConditionFalse,
+		nmstate.NodeNetworkConfigurationPolicyConditionConfigurationProgressing,
```
mkowalski: This should be `nmstate.NodeNetworkConfigurationPolicyConditionFailedToConfigure`. You may also want to fix this in the condition above.

api/shared/nodenetworkconfigurationpolicy_types.go
```diff
@@ -62,6 +62,7 @@ const (
 	NodeNetworkConfigurationPolicyConditionAvailable   ConditionType = "Available"
 	NodeNetworkConfigurationPolicyConditionDegraded    ConditionType = "Degraded"
 	NodeNetworkConfigurationPolicyConditionProgressing ConditionType = "Progressing"
+	NodeNetworkConfigurationPolicyConditionIgnored     ConditionType = "Ignored"
```
mkowalski, line 65: You added a new const here, you should also add it to the list `var NodeNetworkConfigurationPolicyConditionTypes = [...]ConditionType{` just below

mkowalski:
@ai review

mkowalski:
@coderabbit review

PR #1381, opened 2025-10-23
Fix NNCP status loss after ungraceful reboot

mkowalski:
After an ungraceful cluster reboot, NodeNetworkConfigurationPolicy resources could lose their STATUS field because the handler relied entirely on Kubernetes watch events, which are not fired for existing resources when pods restart.

This fix implements two complementary mechanisms:

1. Startup reconciliation: Triggers 5 seconds after handler startup, enqueuing all NNCPs matching the node for immediate reconciliation.

2. Periodic reconciliation: Runs every 10 minutes to continuously ensure all matching NNCPs have status, providing ongoing resilience.

The implementation uses a channel-based event system integrated with controller-runtime's watch mechanism, ensuring non-blocking operation and graceful shutdown. Includes 10 new comprehensive unit tests covering all code paths, with full test coverage for success and failure scenarios.

Fixes: OCPBUGS-37809

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/OCPBUGS-37809-FIX.md
+++ b/OCPBUGS-37809-FIX.md
@@ -1,0 +1,212 @@
+# Fix for OCPBUGS-37809: NodeNetworkConfigurationPolicy not evaluated after ungraceful reboot
+
+## Summary
+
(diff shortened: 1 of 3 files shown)


mkowalski:
/hold

This is 100% AI work

PR #1383, opened 2025-10-24
Add retry logic for NNCP status update failures

mkowalski:
When status updates fail (e.g., due to webhook TLS certificate issues during cluster restart), return the error to trigger controller requeue with a 10-second delay instead of silently failing.

This ensures NNCP status is eventually updated after the webhook becomes available, preventing policies from being stuck in "Progressing" state when the actual network configuration has already succeeded.

The retry mechanism handles various failure scenarios:
- Webhook CA bundle not yet injected by service-ca-operator
- Webhook pods not ready during startup
- Temporary API server issues

Changes:
- Modified decrementUnavailableNodeCount() to return error instead of void
- Updated caller to check error and requeue with 10-second delay
- Preserved existing retry logic (cached client → non-cached client)

Relates to: OCPBUGS-37809, OCPBUGS-49722

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/rollback_test.go
+++ b/test/e2e/handler/rollback_test.go
@@ -206,7 +206,7 @@ var _ = Describe("rollback", func() {
 			Byf("Check that %s is rolled back", primaryNic)
 			Eventually(func() bool {
 				return autoDNS(nodes[0], primaryNic)
-			}, 480*time.Second, ReadInterval).Should(BeTrue(), "should eventually have auto-dns=true")
+			}, RollbackTimeout, ReadInterval).Should(BeTrue(), "should eventually have auto-dns=true")
 
 			Byf("Check that %s continue with rolled back state", primaryNic)
 			Consistently(func() bool {
@@ -234,7 +234,7 @@ var _ = Describe("rollback", func() {
 			Byf("Check that %s is rolled back", primaryNic)
 			Eventually(func() bool {
 				return autoDNS(nodes[0], primaryNic)
-			}, 480*time.Second, ReadInterval).Should(BeTrue(), "should eventually have auto-dns=true")
+			}, RollbackTimeout, ReadInterval).Should(BeTrue(), "should eventually have auto-dns=true")
 
 			Byf("Check that %s continue with rolled back state", primaryNic)
 			Consistently(func() bool {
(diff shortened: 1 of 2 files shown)


controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -520,16 +523,28 @@ func (r *NodeNetworkConfigurationPolicyReconciler) incrementUnavailableNodeCount
 
 func (r *NodeNetworkConfigurationPolicyReconciler) decrementUnavailableNodeCount(
 	policy *nmstatev1.NodeNetworkConfigurationPolicy,
-	generationKey string) {
+	generationKey string) error {
 	policyKey := types.NamespacedName{Name: policy.GetName(), Namespace: policy.GetNamespace()}
 	err := tryDecrementingUnavailableNodeCount(r.Client, r.Client, policyKey, generationKey)
 	if err != nil {
+		// If there are no unavailable nodes, this is not an error - the node was already processed
+		if err.Error() == "no unavailable nodes" {
```
qinqon: Let's create a proper golang error to match with it instead of a string

controllers/handler/nodenetworkconfigurationpolicy_controller.go
```diff
@@ -520,16 +523,28 @@ func (r *NodeNetworkConfigurationPolicyReconciler) incrementUnavailableNodeCount
...
+			r.Log.V(1).Info("No unavailable nodes to decrement, node already processed")
+			return nil
+		}
 		r.Log.Error(err, "error decrementing unavailableNodeCount with cached client, trying again with non-cached client.")
 		err = tryDecrementingUnavailableNodeCount(r.Client, r.APIClient, policyKey, generationKey)
 		if err != nil {
+			// Again, check if it's the "no unavailable nodes" case
+			if err.Error() == "no unavailable nodes" {
```
qinqon: ditto

controllers/handler/nodenetworkconfigurationpolicy_controller.go
qinqon: can we just return no error if "no unavailable" ? at the end is kind of a noop.

qinqon (changes requested):

controllers/handler/nodenetworkconfigurationpolicy_controller.go
mkowalski: Ok done

PR #1387, opened 2025-11-04
Prevent non mandatory mutating webhooks from making network configuration fail.

qinqon:

/kind bug

Reseting the NNCP status by cleaning up the status condition list force the NNCP status webhook to be called and set conditions to Unknown. This change directly set the NNCP status conditions to unknown at the controller, also this PR change the errorPolicy to Ignore on non mandatory mutating webhooks.

**Release note**:

```release-note
Prevent non mandatory mutating webhooks from making network configuration fail
```


mkowalski:
/retest

Deployment issue

```
  Warning  FailedMount  58s (x10 over 5m7s)  kubelet            MountVolume.SetUp failed for volume "tls-key-pair" : secret "nmstate-webhook" not found
```

PR #1389, opened 2025-11-04
Add slash command to run all CI pre-merge checks

emy:

>
> /kind bug

/kind enhancement

  Add new /ci-checks slash command that runs all CI verification checks
  except end-to-end tests to catch issues before they hit CI.

  The command runs the following checks sequentially:
  - Code generation verification (make check-gen)
  - Static code checks (make check - lint, vet, formatting)
  - Unit tests (make test/unit)
  - API unit tests (make test/unit/api)
  - Bundle verification (make check-bundle)
  - OCP bundle verification (make check-ocp-bundle) if available

  Features:
  - Automatically unsets IMAGE_REPO during execution to match CI environment
  - Continues running all checks even if some fail to provide complete picture
  - Auto-rollbacks bundle CSV file if only createdAt timestamp changed
  - Provides comprehensive summary with passed/failed checks and fix suggestions
  - Tracks progress with todo list for visibility

  This helps developers verify their changes locally before pushing to avoid
  CI failures.

  Co-Authored by: Claude

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,8 +118,6 @@ check: lint vet whitespace-check gofmt-check promlint-check
 format: whitespace-format gofmt
 
 pre-push: check test/unit
-	# We want to set these vars to not generate a csv with the local variables because it will always mismatch.
-	IMAGE_REPO=nmstate HANDLER_PULL_POLICY=IfNotPresent OPERATOR_PULL_POLICY=IfNotPresent $(MAKE) check-gen || exit 1
 
 vet:
 	GOFLAGS=-mod=mod go vet ./...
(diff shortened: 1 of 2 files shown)


mkowalski:
> @emy can we make it a proper make target and call it `pre-commit` <- that's a usual pattern, so developers has a actionable check to ensure minimal CI is fine and Claude can call pre-commit too.

We discussed it. I like the hook approach but pre-commit is a bit too expensive. Maybe pre-push or something like this? It's a command that runs longer than sub-second...

**Also** the current command misses check that ensures all the commit messages contain sign-off stanza.

qinqon:
> > @emy can we make it a proper make target and call it `pre-commit` <- that's a usual pattern, so developers has a actionable check to ensure minimal CI is fine and Claude can call pre-commit too.
> 
> We discussed it. I like the hook approach but pre-commit is a bit too expensive. Maybe pre-push or something like this? It's a command that runs longer than sub-second...
> 
> **Also** the current command misses check that ensures all the commit messages contain sign-off stanza.

pre-push sounds good to me, but pre-commit is kind of de facto standard at other projects.

emy:
Thanks for the comments @mkowalski @qinqon i'll add it with pre-push. I checked the run times and I agree it's def. too expensive. I already added the sign off part yesterday in the night but didn't commit that yet as I felt more was coming up after seeing the Gemini review.

Makefile
```diff
@@ -117,6 +117,15 @@ check: lint vet whitespace-check gofmt-check promlint-check
...
 
+pre-push:
+	@echo "Running pre-push CI checks..."
+	IMAGE_REPO=nmstate HANDLER_PULL_POLICY=IfNotPresent OPERATOR_PULL_POLICY=IfNotPresent $(MAKE) check-gen || exit 1
+	$(MAKE) check || exit 1
+	$(MAKE) test/unit || exit 1
+	$(MAKE) test/unit/api || exit 1
+	@# Check if bundle file was modified with only timestamp changes and rollback if so
+	@echo "✅ All pre-push checks passed!"
+
```
qinqon: We should implemnt this with Makefile target dependencies instead of this weird script.

emy: Fair point. I'll change that. Thanks for the feedback!

Makefile
```diff
@@ -117,6 +117,10 @@ check: lint vet whitespace-check gofmt-check promlint-check
 
 format: whitespace-format gofmt
 
+pre-push: check test/unit
+	# We want to set these vars to not generate a csv with the local variables because it will always mismatch.
+	IMAGE_REPO=nmstate HANDLER_PULL_POLICY=IfNotPresent OPERATOR_PULL_POLICY=IfNotPresent $(MAKE) check-gen || exit 1
+
```
qinqon, line 121:
```suggestion
pre-push: IMAGE_REPO=nmstate HANDLER_PULL_POLICY=IfNotPresent OPERATOR_PULL_POLICY=IfNotPresent
pre-push: check test/unit check-gen
```

qinqon (changes requested):

Makefile
emy: After talking to @mkowalski it was decided it's not needed to check bundle here.

PR #1390, opened 2025-11-06
Mount /etc/systemd/network in handler pod for interface alt-names

cgoncalves:

/kind enhancement

Add volume mount for /etc/systemd/network to allow the nmstate handler pod to write systemd .link files for interface alternative names persistence. Without this mount, alternative names would only be configured after the handler pod starts following a node reboot, rather than during early boot.

This change supports the interface alternative name feature introduced in nmstate 2.2.51, which uses systemd .link files for persistence instead of NetworkManager.

References:
- https://github.com/nmstate/nmstate/pull/2979
- https://nmstate.io/features/alt-name.html

**Release note**:

```release-note
The handler pod now mounts /etc/systemd/network to enable persistence of interface alternative names via systemd .link files. This ensures alternative names are available during early boot, not just after the handler pod starts following a node reboot. This enhancement supports the interface alternative name feature (https://nmstate.io/features/alt-name.html) introduced in nmstate 2.2.51.
```

### DIFF
--- a/deploy/handler/operator.yaml
+++ b/deploy/handler/operator.yaml
@@ -428,6 +428,8 @@ spec:
               mountPath: /var/k8s_nmstate
             - name: ovs-socket
               mountPath: /run/openvswitch
+            - name: systemd-network
+              mountPath: /etc/systemd/network
           securityContext:
             privileged: true
           readinessProbe:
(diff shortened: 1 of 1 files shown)


qinqon:
/hold
@cgoncalves can you add some e2e for this ? so we verify that the systemd mount works.

mkowalski:
/hold cancel

Agreed to merge

PR #1395, opened 2025-11-07
proposal: delete configuration on nncp delete

mkowalski:

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
NONE
```


design-proposals/delete-changes-on-nncp-delete.md
```diff
@@ -0,0 +1,188 @@
...
+
+1) We want users to have a consistent behavior between `nmstatectl gr` and NNCP deletion. The behavior when deleting NNCP should be aligned with what would happen if user generated manifest using `nmstatectl gr` and applied it as NNCP.
+
+2) We want kubernetes-nmstate to use its health probes to block removal of NNCP if that would break connectivity to the node.
+
+### Non-Goals
+
+1) We are not supporting removal/revert of the configuration when NNCP is modified.
+
+2) We are not taking snapshot of network configuration before NNCP was applied in order to revert to that snapshot. We are only deleting changes that specific NNCP applies.
```
cgoncalves, line 31: Say an interface is DHCP-configured, and an NNCP changes it to a static IP. What happens when the NNCP is deleted? Does it revert to DHCP or to no IP configuration at all?

Or an SR-IOV PF with 4 VFs. NNCP increases to 6 VFs and then NNCP is deleted. Outcome is 4 VFs or no VFs?

mkowalski: >Say an interface is DHCP-configured, and an NNCP changes it to a static IP. What happens when the NNCP is deleted? Does it revert to DHCP or to no IP configuration at all?

Forget about the whole "before" part because it's stateless so it has no memory. So I apply the following manifest (static IPv4 + static IPv6)

```
$ cat nmstate.ymla
interfaces:
- name: enp0s13f0u3u4u3
  type: ethernet
  state: up
  ipv4:
    address:
    - ip: 192.168.122.250
      prefix-length: 24
    enabled: true
  ipv6:
    address:
    - ip: 2001:db8::1:1
      prefix-length: 64
    enabled: true
```

Then I run `gr` on that and what I get is

```
interfaces:
- name: enp0s13f0u3u4u3
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: true
    address: []
    auto-dns: true
    auto-gateway: true
    auto-routes: true
  ipv6:
    enabled: true
    dhcp: true
    autoconf: true
    address: []
    auto-dns: true
    auto-gateway: true
    auto-routes: true
```

**BUT** if before the first manifest I already had static IP (so I only changed one static to another static) I would end up with the same gr-ed manifest

>Or an SR-IOV PF with 4 VFs. NNCP increases to 6 VFs and then NNCP is deleted. Outcome is 4 VFs or no VFs?

You will never get "back" to 4 because nothing in the system remembers what that was before

qinqon:
No longer required by users.

fbaudin:
This is still required by end users. NMSTATE is not behaving like other Operators. The whole configuration needs to be reflected in the CRDs, so NMSTATE would really be intend based.

PR #1399, opened 2025-11-19
Add VRF route persistence test using vrf-name

emy:

>
> /kind bug

/kind enhancement

Add e2e test to verify VRF routes specified by vrf-name persist across node reboots. This tests the new vrf-name parameter for routes that allows referencing VRF route tables by interface name instead of numeric table-id.

The test includes:
   - Basic verification test for vrf-name route configuration
   - Reboot persistence test to ensure routes survive node restart

Depeds-on: nmstate 2.2.55

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/handler/states.go
+++ b/test/e2e/handler/states.go
@@ -237,6 +237,34 @@ func vrfAbsent(vrfID string) nmstate.State {
 `, vrfID))
 }
 
+// vrfRouteWithVrfName creates a VRF interface and a route using vrf-name instead of table-id.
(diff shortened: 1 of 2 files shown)


emy:
Handler e2e fails because the required nmstate version is [2.2.55](https://github.com/nmstate/nmstate/releases/tag/v2.2.55) and above.

qinqon:
> Handler e2e fails because the required nmstate version is [2.2.55](https://github.com/nmstate/nmstate/releases/tag/v2.2.55) and above.

We need to see this test passing to merge it, although would be good to skip it if nmstate version is no the expected, the nmstate version can be check looking at the NNS

PR #1407, opened 2025-12-02
Add configurable NNCP retry strategy via NMState CR

emy:

>
> /kind bug

/kind enhancement

The recently added NNCP retry added some delay into the CI (due to the negative tests which will result in the NNCP retrying for minutes each time). So I added a configurable backoff strategy which keeps the same defaults but offers configuration. 

**Release note**:

```release-note
The NNCP backoff strategy can be configured through the NMState CR now. (Default values stay initial delay 1s, max. delay 30s, 5 retries)
```


qinqon:
After some offline discussion we don't want to add more complexity to the installation interface, we should suffer little wI time increase, but nothing is for free.

PR #1414, opened 2025-12-19
metrics: introduce metrics based on the NNS

mkowalski:
This PR introduces a small set of metrics based on the NodeNetworkState, i.e. `kubernetes_nmstate_network_interfaces` and
`kubernetes_nmstate_routes`. Those track number of network interfaces and number of routes for every node in the cluster where nmstate handler pods are running.

This creates a skeleton for k-nmstate to create metrics based on the effective state of the network on the system and not only based on the content of NNCP.

In the future it may be used to create statistics about network topologies of clusters.

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
New metrics `kubernetes_nmstate_network_interfaces` and `kubernetes_nmstate_routes` are introduced to track number of network interfaces and number of routes configured per node.
```

### DIFF
--- a/cmd/handler/main.go
+++ b/cmd/handler/main.go
@@ -83,6 +83,8 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 
 	metrics.Registry.MustRegister(monitoring.AppliedFeatures)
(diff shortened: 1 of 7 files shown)


mkowalski:
image
image

PR #1416, opened 2025-12-19
test: add cleanup in BeforeEach to avoid duplicated policies

mkowalski:
This change attempts to remediate an issue where test fails because of a duplicated policy, i.e.

```
nodenetworkconfigurationpolicies.nmstate.io "test-policy" already exists
```

which may potentially be an outcome of a missing cleanup in some other test. With this change we will always try to create the policy first before starting a test to ensure a clean starting state.

>
/kind bug
> /kind enhancement

**Release note**:

```release-note
NONE
```


mkowalski:
/hold

Not solving the issue I hoped it would solve

PR #1423, opened 2026-01-14
design: Add netplan backend design proposal

qinqon:

>
> /kind bug
> /kind enhancement

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,13 +24,6 @@ OPERATOR_IMAGE_TAG ?= latest
 OPERATOR_IMAGE_FULL_NAME ?= $(IMAGE_REPO)/$(OPERATOR_IMAGE_NAME):$(OPERATOR_IMAGE_TAG)
 OPERATOR_IMAGE ?= $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE_FULL_NAME)
 
-KUBE_RBAC_PROXY_NAME ?= origin-kube-rbac-proxy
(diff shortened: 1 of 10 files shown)


design-proposals/netplan-backend-support.md
```diff
@@ -0,0 +1,541 @@
...
+4. Handler pods initialize the netplan backend on startup
+5. When a NodeNetworkConfigurationPolicy is created, the handler uses the netplan D-Bus API to apply network configuration
+6. Network state is reported back through the same netplan backend
+
+### Workflow Description (Switching Back to NMState Backend)
+
+1. The cluster administrator edits the NMState CR and sets `spec.backend: nmstate` (or removes the backend field entirely)
+2. The operator updates the handler DaemonSet environment variables
+3. Handler pods restart and initialize the nmstate backend
+4. Subsequent policies are applied via nmstatectl
```
mkowalski: We need to be specific on defining what happens to the existing NNCPs when you switch backend at runtime. I don't know yet how to solve it in the most elegant way but also I don't think it's a real issue (because I don't think any cluster admin changes the underlying OS of the nodes), ever.

Just to say that if you have a bunch of NNCPs created with backend 1, then switching to backend 2 will make syntax of this NNCPs wrong, so an attempt to reconcile them will create a gazillion of errors.

Maybe it's as simple as storing `spec.backend` into `status` of NNCP and only reconcile those that have matching value? Anything else would go into `Ignored` or similar state.

Issue #1418, opened 2025-12-24
nmstate metrics pod unhealthy. Keeps crashing.


kristvanbesien:
I deployed the kubernetes-nmstack operator, following the instructions listed at:
https://github.com/nmstate/kubernetes-nmstate/releases/tag/v0.85.1

Not everything appears to deploy fine:

```
% kubectl get pods -n nmstate
NAME                                    READY   STATUS             RESTARTS       AGE
nmstate-cert-manager-745bf8f4fc-vhrsr   1/1     Running            0              12m
nmstate-handler-64kb2                   1/1     Running            0              12m
nmstate-handler-lhc8b                   1/1     Running            0              12m
nmstate-handler-zgjng                   1/1     Running            0              12m
nmstate-metrics-74ddd9c697-47wvd        1/2     CrashLoopBackOff   5 (118s ago)   5m21s
nmstate-operator-68dcbf8c55-vfdzt       1/1     Running            0              12m
nmstate-webhook-58cc4b5957-tnds5        1/1     Running            0              12m
```

Further investigation turned up that the problem is that the metrics pod contains a container that uses an image from origin:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nmstate-metrics
  namespace: nmstate
 (...)
spec:
  replicas: 1
  selector:
    matchLabels:
      name: nmstate-metrics
  template:
(...)
    spec:
      containers:
(...)
      - name: kube-rbac-proxy
        image: quay.io/openshift/origin-kube-rbac-proxy:4.10.0
(...)
```

And here is the problem: Images from Origin are only good for the intel architecture, and my cluster runs on arm...

I manged to fix this by changing the value of the KUBE_RBAC_PROXY_IMAGE in the operator.yaml, to quay.io/brancz/kube-rbac-proxy:v0.20.1

So please consider for the next release to make sure that all image references point to multi arch images.

qinqon:
Thanks, we will look into this, in fact I think with newere controller-runtime versions we don't heed the kube rba proxy image

mkowalski:
Hi Krist, just an admin thing - as Red Hat please open us OCPBUGS rather than GitHub issue so that we can track it better. GH issues fall under cracks and we don't actually track the progress of those in the same way we track bugs in Jira.

On a more technical note, the issue belongs to the ART team and they should make sure that `origin-kube-rbac-proxy` is a multi-arch image. For downstream productized images this is the case, for origin it may be not.

[...]

PR #1426, opened 2026-01-23
metrics: Replace kube-rbac-proxy with built-in authentication

qinqon:
/kind enhancement

Replace the kube-rbac-proxy sidecar in the nmstate-metrics Deployment with controller-runtime's built-in WithAuthenticationAndAuthorization filter for metrics authentication and authorization.

The existing RBAC permissions for tokenreviews and subjectaccessreviews in the handler ClusterRole are sufficient for this functionality.

Closes https://github.com/nmstate/kubernetes-nmstate/issues/1418

```release-note
Replace kube-rbac-proxy container with WithAuthenticationAndAuthorization
```


pkg/metrics/filters/filters.go
```diff
@@ -0,0 +1,139 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package filters provides authentication and authorization filters for the metrics server.
+// This is a local implementation of the controller-runtime metrics filters for projects
+// that don't have the filters package vendored.
+package filters
```
mkowalski: I honestly don't know because I never implemented this, but if we are using controller-runtime "mechanism" why do we need this whole file copy-pasted here? Shouldn't it be just a library with functions that you vendor and consume?

PR #1428, opened 2026-01-29
NO-JIRA: Bump ginkgo version used when no ginkgo run command is set on GINKGO env

emy:

>
> /kind bug
> /kind enhancement

When the `GINKGO` environment variable is empty and make test/unit or an e2e test is run the ginkgo versions mismatch because it's still using the old 2.11.0 version:

```make test/unit 
~ %  cd api && GOFLAGS=-mod=mod go run github.com/onsi/ginkgo/v2/ginkgo@v2.11.0 --junit-report=junit-api-unit-test.xml -r -keep-going --randomize-all --randomize-suites --race --trace  ./...
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.11.0
  Mismatched package versions found:
    2.22.1 used by names, shared, v1beta1
```
Bumping the version makes this warning go away.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,8 @@ export KUBECTL ?= ./cluster/kubectl.sh
 KUBECTL ?= ./cluster/kubectl.sh
 OPERATOR_SDK_VERSION ?= 1.37.0
 
-GINKGO = GOFLAGS=-mod=mod go run github.com/onsi/ginkgo/v2/ginkgo@v2.22.1
+GINKGO_VERSION ?= v2.22.1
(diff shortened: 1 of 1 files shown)


mkowalski:
/approve
/lgmt

There might have been a reason not to upgrade downstream, but we will handle it there

mkowalski:

```
Handler E2E Test Suite: [It] NodeNetworkState when desiredState is configured with a linux bridge up with a port with disabled VLAN should have the linux bridge at currentState with vlan_filtering 0 and no default vlan range configured expand_more
```

This issue happens more often than it should. One day it will explode.

PR #1430, opened 2026-02-06
fix(handler): Pending/NotPending returning wrong condition count

qinqon:
/kind bug

The Pending() and NotPending() methods were incorrectly using progressing() instead of pending(), returning wrong counts.

**Release note**:

```release-note
The Pending() and NotPending() methods were incorrectly using progressing() instead of pending(), returning wrong counts.
```

### DIFF
--- a/pkg/enactmentstatus/conditions/counter.go
+++ b/pkg/enactmentstatus/conditions/counter.go
@@ -96,10 +96,10 @@ func (c ConditionCount) NotProgressing() int {
 	return c.progressing().false()
 }
 func (c ConditionCount) Pending() int {
-	return c.progressing().true()
+	return c.pending().true()
 }
 func (c ConditionCount) NotPending() int {
-	return c.progressing().false()
+	return c.pending().false()
 }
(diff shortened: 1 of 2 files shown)


pkg/enactmentstatus/conditions/counter.go
```diff
@@ -96,10 +96,10 @@ func (c ConditionCount) NotProgressing() int {
 	return c.progressing().false()
 }
 func (c ConditionCount) Pending() int {
```
mkowalski, line 98: /gemini help

mkowalski:
/lgtm
/approve

If passes CI, good

pkg/enactmentstatus/conditions/counter.go
mkowalski: @gemini-code-assist, find all the places in the codebase that rely on this function

mkowalski: @gemini-code-assist, thank you. You were as useless as I would expect.

PR #1439, opened 2026-02-16
fix: use Set/Delete instead of Inc/Dec for applied features gauge

qinqon:
/kind bug

The NNCE metrics controller used `Inc()`/`Dec()` with an `oldNNCEs` map to
track feature gauge transitions. This is fundamentally broken when
multiple NNCEs change the same feature simultaneously (the normal case
-- a policy targets all nodes), as `Inc`/`Dec` accumulates errors across
concurrent reconciliations.

Replace the `Inc()`/`Dec()` approach with `Set(1)`/`Delete()`, following the
same pattern used by the NNS metrics controller (`updateNodeInterfaceMetrics`).
On each reconciliation the full NNCE list is fetched, a deduplicated feature
set is built, and absolute gauge values are set. An `oldFeatures` set tracks
previously-seen feature names so stale metrics are cleaned up when a feature
disappears.

Also removes the now-dead `Stats.Subtract()` helper from `nmstatectl` and
fixes the e2e test to expect metric deletion instead of a zero value.

NONE

**Release note**:
```release-note
Fix applied features gauge double-counting by using absolute Set/Delete instead of incremental Inc/Dec
```

### DIFF
--- a/test/e2e/handler/metrics_test.go
+++ b/test/e2e/handler/metrics_test.go
@@ -137,7 +137,7 @@ routes:
 					}).
 						WithPolling(time.Second).
 						WithTimeout(2 * time.Second).
-						Should(HaveKeyWithValue(monitoring.AppliedFeaturesOpts.Name+`{name="dhcpv4-custom-hostname"}`, "0"))
+						ShouldNot(HaveKey(monitoring.AppliedFeaturesOpts.Name + `{name="dhcpv4-custom-hostname"}`))
 				})
 			})
 		})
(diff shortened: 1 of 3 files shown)


test/e2e/handler/metrics_test.go
```diff
@@ -137,7 +137,7 @@ routes:
 					}).
 						WithPolling(time.Second).
 						WithTimeout(2 * time.Second).
-						Should(HaveKeyWithValue(monitoring.AppliedFeaturesOpts.Name+`{name="dhcpv4-custom-hostname"}`, "0"))
```
mkowalski, line 140: This looks like the old test was different than the new one. Why?

qinqon: Because now we are just using Delete when count is zero

PR #1453, opened 2026-03-11
Use centralized TLS profile if exists

mkowalski:

>
> /kind bug
/kind enhancement

We start using centralized TLS profile fetched from the k8s-apiserver
configuration if available. As a result, in those environments we no
longer manage TLS configuration ourselves but consume global
configuration.

Thanks to this we no longer need to follow the latest recommendations as
per cipher suites, available curves and so on.

K8s-apiserver is now the source of truth.

**Release note**:

```release-note
* Kube-rbac-proxy is not used any more.
* If TLS profile is available on the platform (now only implemented for OpenShift), it will be used for webhook and metrics server. Configuration will be dynamically updated when the profile changes.
```

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -58,8 +58,7 @@ func FetchOpenShiftTLSOpts(cfg *rest.Config, scheme *runtime.Scheme) (func(*tls.
 
 	isOCP, err := IsOpenShift(kclient)
 	if err != nil {
-		log.Info("Warning: could not determine if running on OpenShift, assuming not")
-		return nil, nil
+		return nil, fmt.Errorf("could not determine if running on OpenShift: %w", err)
 	}
 	if !isOCP {
 		return nil, nil


pkg/cluster/cluster.go
```diff
@@ -39,3 +47,35 @@ func IsOpenShift(kclient client.Client) (bool, error) {
...
+func FetchOpenShiftTLSOpts(scheme *runtime.Scheme) (func(*tls.Config), error) {
+	cfg := ctrl.GetConfigOrDie()
+	kclient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, fmt.Errorf("failed creating client for TLS profile detection: %w", err)
+	}
+
+	isOCP, err := IsOpenShift(kclient)
+	if err != nil {
+		log.Info("Warning: could not determine if running on OpenShift, assuming not")
```
qinqon: Consider logging at Error/Warning level, or retrying since this runs at startup:
```go
if err != nil {
    log.Error(err, "could not determine if running on OpenShift, assuming not")
    return nil, nil
}
```
Alternatively, consider whether this should actually return the error to fail fast rather than silently degrading.

pkg/cluster/cluster.go
```diff
@@ -39,3 +47,35 @@ func IsOpenShift(kclient client.Client) (bool, error) {
 
 	return true, nil
 }
+
+// FetchOpenShiftTLSOpts detects OpenShift and fetches the TLS profile for
+// secure serving. Returns nil on non-OpenShift clusters.
+func FetchOpenShiftTLSOpts(scheme *runtime.Scheme) (func(*tls.Config), error) {
+	cfg := ctrl.GetConfigOrDie()
```
qinqon: nit: `ctrl.GetConfigOrDie()` is also called inside `ctrl.NewManager()` in `createManager`. Consider accepting the rest config as a parameter to avoid the duplication and the second potential panic point.

cmd/handler/main.go
```diff
@@ -128,17 +134,43 @@ func mainHandler() int {
...
+	platformTLSOpts, err := cluster.FetchOpenShiftTLSOpts(scheme)
+	if err != nil {
+		setupLog.Error(err, "unable to fetch TLS configuration")
+		return generalExitStatus
+	}
+
+	// Compose the TLS opts applied to all TLS-enabled servers.
+	tlsOpts := composeTLSOpts(platformTLSOpts)
+
+	mgr, err := createManager(tlsOpts, platformTLSOpts != nil /*isOpenShift*/)
```
qinqon: `isOpenShift` is derived from `platformTLSOpts != nil`. If `FetchOpenShiftTLSOpts` ever returns `nil` then `isOpenShift` would incorrectly be `false` , disabling `filters.WithAuthenticationAndAuthorization` on the metrics endpoint.

Consider returning the OpenShift detection result separately from the TLS opts, e.g.:
```go
isOpenShift, platformTLSOpts, err := cluster.FetchOpenShiftTLSOpts(scheme)
```

cmd/handler/main.go
```diff
@@ -167,19 +199,68 @@ func setupHandlerLockIfNeeded() (*flock.Flock, error) {
...
-// createManager creates and configures the controller manager
-func createManager() (manager.Manager, error) {
+// composeTLSOpts returns TLS options for all TLS-enabled servers.
+// Always disables HTTP/2 (CVE-2023-39325), and adds the platform profile opts if present.
+func composeTLSOpts(tlsOpts func(*tls.Config)) func(*tls.Config) {
+	return func(c *tls.Config) {
+		if tlsOpts != nil {
+			tlsOpts(c)
+		}
+		c.NextProtos = []string{"http/1.1"}
```
qinqon, line 212: nit: Add a comment about this being enforced for CVE-2023-39325

mkowalski: Human or AI ? The comment is literally in the same green block

cmd/handler/main.go
mkowalski: >isOpenShift is derived from platformTLSOpts != nil

Where does it come from?

In upstream code we have 

```
func IsOpenShift(kclient client.Client) (bool, error) {
	// if the cluster has the securityContextConstraint resource of the group security.openshift.io, then it is most likely an OCP/OKD cluster
	_, err := kclient.RESTMapper().ResourcesFor(securityv1.SchemeGroupVersion.WithResource("securitycontextconstraints"))

[...]
```

and in downstream we have 

```
func IsOpenShift(kclient client.Client) (bool, error) {
	return true, nil
```

>If FetchOpenShiftTLSOpts ever returns nil

If you look at the definition of `func FetchOpenShiftTLSOpts` in `cluster.go L53`, it seems to me that if you run on OpenShift, you can never end up with empty `tlsOpts` because if `FetchAPIServerTLSProfile` errors, then we return an error and not `nil`.

pkg/cluster/cluster.go
mkowalski: Good point

pkg/cluster/cluster.go
mkowalski: Good point

cmd/handler/main.go
qinqon: Fair enough

cmd/handler/main.go
qinqon: We can close this

PR #1467, opened 2026-03-24
fix(handler): eliminate API server dependency at startup for TLS config

qinqon:
/kind bug

Eliminates API server calls at handler startup for TLS configuration, preventing crash loops when network connectivity is temporarily unavailable.

The handler binary calls the API server at startup to detect OpenShift (`IsOpenShift()` → `GET /api`) and fetch the TLS security profile (`FetchAPIServerTLSProfile()`). When the default network interface is being reconfigured (e.g., bond removal in e2e tests), the API server is temporarily unreachable, causing a 30s timeout → handler crash → restart loop (12 restarts observed in [build 2036339976228048896](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/1465/pull-kubernetes-nmstate-e2e-handler-k8s/2036339976228048896)).

The fix:
1. **`IS_OPENSHIFT` env var** — Operator passes platform detection to all handler-deployed pods, eliminating the `IsOpenShift()` API discovery call
2. **TLS profile ConfigMap** — On OpenShift, the operator fetches the TLS profile and writes it to a ConfigMap. Webhook and metrics pods mount it and read from file. Kubelet caches ConfigMap volumes locally, so they survive container restarts without API server connectivity
3. **Handler skips TLS profile** — The handler DaemonSet doesn't serve platform-specific TLS endpoints (webhook and metrics are separate Deployments), so it skips TLS profile loading entirely
4. **Operator-driven TLS profile updates** — The `SecurityProfileWatcher` is removed from webhook/metrics pods. Instead, the operator watches the `APIServer` CR for TLS profile changes, updates the ConfigMap, and rolls webhook/metrics Deployments automatically via a `nmstate.io/tls-profile-hash` annotation on the pod template

On container restart (the key scenario):
1. Pod starts with API server connectivity → kubelet mounts ConfigMap
2. Bond removal disrupts connectivity → handler crashes
3. Kubelet restarts container → volumes already cached locally
4. Handler reads config from file → no API server call → starts successfully

TLS profile change flow (OpenShift only):
1. Admin changes TLS profile on `APIServer` CR
2. Operator's watch triggers reconcile
3. Operator re-fetches TLS profile, updates ConfigMap, re-renders Deployments with new hash annotation
4. Kubernetes rolls webhook/metrics Deployments with updated TLS config

Openshift test:
- https://github.com/openshift/kubernetes-nmstate/pull/739

**Release note**:

```release-note
Fix handler crash loop when default network interface is temporarily reconfigured by eliminating API server dependency at startup for TLS configuration.
```

### DIFF
--- a/cmd/handler/main.go
+++ b/cmd/handler/main.go
@@ -68,11 +68,13 @@ import (
 	nmstatelog "github.com/nmstate/kubernetes-nmstate/pkg/log"
 	"github.com/nmstate/kubernetes-nmstate/pkg/monitoring"
 	"github.com/nmstate/kubernetes-nmstate/pkg/nmstatectl"
-	nmstatetls "github.com/nmstate/kubernetes-nmstate/pkg/tls"
 	"github.com/nmstate/kubernetes-nmstate/pkg/webhook"
 )
 
-const generalExitStatus int = 1
+const (
+	generalExitStatus    int    = 1
+	tlsProfileConfigPath string = "/etc/nmstate/tls/profile.json"
+)
 
 const (
 	defaultRetriesUntilFail = 5
@@ -143,37 +145,36 @@ func mainHandler() int {
 
 	cfg := ctrl.GetConfigOrDie()
 
-	// Detect OpenShift and fetch TLS profile early, before creating the
-	// manager, so both the metrics server and webhooks share the same config.
-	platformTLSOpts, err := cluster.FetchOpenShiftTLSOpts(cfg, scheme)
-	if err != nil {
-		setupLog.Error(err, "unable to fetch TLS configuration")
-		return generalExitStatus
+	// Detect OpenShift from env var set by the operator, avoiding an API
+	// server call that can fail when network connectivity is temporarily
+	// unavailable (e.g. during default interface reconfiguration).
+	isOpenShift := cluster.IsOpenShiftFromEnv()
+
+	var platformTLSOpts func(*tls.Config)
+	// Only load TLS profile on OpenShift for components that serve TLS
+	// (webhook and metrics). The handler DaemonSet uses default TLS.
+	// Read from a ConfigMap-mounted file to avoid API server calls.
+	// TLS profile changes are handled by the operator: it updates the
+	// ConfigMap and rolls the Deployments via a hash annotation.
+	if isOpenShift && !environment.IsHandler() {
+		opts, _, err := cluster.FetchTLSProfileFromFile(tlsProfileConfigPath)
+		if err != nil {
+			setupLog.Error(err, "unable to load TLS configuration from file")
+			return generalExitStatus
+		}
+		platformTLSOpts = opts
 	}
 
 	// Compose the TLS opts applied to all TLS-enabled servers.
 	tlsOpts := composeTLSOpts(platformTLSOpts)
 
-	mgr, err := createManager(cfg, tlsOpts, platformTLSOpts != nil /*isOpenShift*/)
+	mgr, err := createManager(cfg, tlsOpts, platformTLSOpts != nil)
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
 		return generalExitStatus
 	}
 
-	// Create a cancellable context so that controllers (e.g. the TLS
-	// security profile watcher) can trigger a graceful shutdown.
-	ctx, cancel := context.WithCancel(ctrl.SetupSignalHandler())
-	defer cancel()
-
-	// On OpenShift, watch for TLS profile changes on processes that serve
-	// TLS (webhook and metrics) and trigger a graceful shutdown so they
-	// restart with the new configuration.
-	if platformTLSOpts != nil && (environment.IsWebhook() || environment.IsMetricsManager()) {
-		if err := setupTLSProfileWatcher(mgr, cancel); err != nil {
-			setupLog.Error(err, "unable to set up TLS profile watcher")
-			return generalExitStatus
-		}
-	}
+	ctx := ctrl.SetupSignalHandler()
 
 	if err := setupControllersByEnvironment(mgr, tlsOpts); err != nil {
 		return generalExitStatus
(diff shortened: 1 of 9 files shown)


cmd/handler/main.go
```diff
@@ -139,18 +139,30 @@ func mainHandler() int {
...
+		}
+		platformTLSOpts = opts
+		initialTLSProfileSpec = &spec
 	}
 
 	// Compose the TLS opts applied to all TLS-enabled servers.
 	tlsOpts := composeTLSOpts(platformTLSOpts)
 
-	mgr, err := createManager(cfg, tlsOpts, platformTLSOpts != nil /*isOpenShift*/)
+	mgr, err := createManager(cfg, tlsOpts, isOpenShift)
```
mkowalski: This slightly changes the behaviour, it's not only a refactor change.

I want to be able to have `platformTLSOpts` configured even if I am not OpenShift.

I want to be able to have empty `platformTLSOpts` even if I am OpenShift.

qinqon: Make sense let me check what we can do.

qinqon: Good point. Fixed in 27617adc9 — `createManager` now uses `platformTLSOpts \!= nil` again to decide whether to apply auth/authz filters, keeping the two concerns independent.

cmd/handler/main.go
```diff
@@ -139,18 +139,30 @@ func mainHandler() int {
...
+	// unavailable (e.g. during default interface reconfiguration).
+	isOpenShift := cluster.IsOpenShiftFromEnv()
+
+	var platformTLSOpts func(*tls.Config)
+	var initialTLSProfileSpec *configv1.TLSProfileSpec
+	// Only load TLS profile on OpenShift for components that serve TLS
+	// (webhook and metrics). The handler DaemonSet uses default TLS.
+	// Read from a ConfigMap-mounted file to avoid API server calls.
+	if isOpenShift && !environment.IsHandler() {
+		opts, spec, err := cluster.FetchTLSProfileFromFile("/etc/nmstate/tls/profile.json")
```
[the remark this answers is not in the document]
qinqon: Makes sense. Extracted to a `tlsProfileConfigPath` constant in 681a76b45.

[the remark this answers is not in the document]
qinqon: Already addressed — the path is now a package-level constant `tlsProfileConfigPath` at `cmd/handler/main.go:77`.

qinqon:
⚠️ Conflict resolution created 1 unexpected new commit(s) instead of resolving within the rebase flow.

Expected: 0 commits (original structure preserved)
Got: 1 commits (new commits added)

Please review the commit history and manually squash or rebase to preserve the original commit structure.

qinqon:
Rebased commit 2585d99 on main and pushed (conflicts resolved).

pkg/cluster/cluster.go
```diff
@@ -79,3 +88,27 @@ func FetchOpenShiftTLSOpts(cfg *rest.Config, scheme *runtime.Scheme) (func(*tls.
...
+		return nil, nmstatetls.TLSProfileSpec{}, fmt.Errorf("failed parsing TLS profile from %s: %w", path, err)
+	}
+
+	tlsOpts, unsupportedCiphers := nmstatetls.NewTLSConfigFromProfile(spec)
+	if len(unsupportedCiphers) > 0 {
+		log.Info("TLS configuration contains unsupported ciphers that will be ignored",
+			"unsupportedCiphers", unsupportedCiphers)
+	}
+
+	return tlsOpts, spec, nil
```
mkowalski, line 80: /gemini looks like caller does not consume "spec" from this function. Why is it needed?

pkg/cluster/cluster.go
```diff
@@ -50,6 +52,13 @@ func IsOpenShift(kclient client.Client) (bool, error) {
...
+// IsOpenShiftFromEnv returns true if the IS_OPENSHIFT environment variable
+// is set to "true". The operator sets this variable on all handler-deployed
+// pods so they can skip the API server discovery call at startup.
+func IsOpenShiftFromEnv() bool {
+	return os.Getenv("IS_OPENSHIFT") == "true"
+}
+
 // FetchOpenShiftTLSOpts detects OpenShift and fetches the TLS profile for
 // secure serving. Returns nil on non-OpenShift clusters.
 func FetchOpenShiftTLSOpts(cfg *rest.Config, scheme *runtime.Scheme) (func(*tls.Config), error) {
```
mkowalski: /gemini isn't "FetchOpenShiftTLSOpts" a dead function in this code right now?

pkg/cluster/cluster.go
mkowalski: /gemini isn't it overengineering? If someone needs it in the future, they can always add it?

controllers/operator/nmstate_controller.go
```diff
@@ -361,6 +389,23 @@ func (r *NMStateReconciler) applyHandler(ctx context.Context, instance *nmstatev
...
+		}
+		tlsJSON, err := json.Marshal(tlsProfileSpec)
+		if err != nil {
+			return fmt.Errorf("failed serializing TLS profile: %w", err)
+		}
+		data.Data["TLSProfileJSON"] = string(tlsJSON)
+		data.Data["TLSProfileHash"] = fmt.Sprintf("%x", sha256.Sum256(tlsJSON))
+	}
+
 	return r.renderAndApply(ctx, instance, data, "handler", true)
```
mkowalski, line 409: /gemini, "handler/operator.yaml" is applied before "handler/tls_config.yaml" so webhook and metrics pod will be created before the ConfigMap with TLS config exists. Explain how it's supposed to work and not fail if webhook and metrics try to mount the configmap as volume

mkowalski: @gemini-code-assist "handler/operator.yaml" is applied before "handler/tls_config.yaml" so webhook and metrics pod will be created before the ConfigMap with TLS config exists. Explain how it's supposed to work and not fail if webhook and metrics try to mount the configmap as volume

qinqon:
CI failures on commit 6657266:

### Related Failures (1)

| Check | Summary |
|-------|--------|
| `pull-kubernetes-nmstate-e2e-handler-k8s` | The fix for the unused imports in `pkg/cluster/cluster.go` was already applied and amended into the commit in the previo... |

Pushed a fix for the related failure.

qinqon:
## CI Failure Analysis — commit cb45598

**Total failures**: 1 (1 unrelated)

---

<details>
<summary>⚠️ Unrelated: <code>pull-kubernetes-nmstate-e2e-handler-k8s</code> — flaky test</summary>

### Error
````
```
[FAILED] Timed out after 120.003s.
Test didn't return to initial state on node node02
Expected
    <map[string]string | len:5>: {"eth0": "up", "eth1": "up", "eth2": "up", "flannel-v6.1": "up", "lo": "up"}
to equal
    <map[string]string | len:6>: {"cni0": "up", "eth0": "up", "eth1": "up", "eth2": "up", "flannel-v6.1": "up", "lo": "up"}
In [AfterEach] at: main_test.go:142
```
````

### Root Cause
The test "should keep the default IP address after node reboot" passed, but the AfterEach cleanup in main_test.go timed out because node02's `cni0` bridge interface was not recreated within 120s after the linux bridge was removed and the node was rebooted. The `cni0` interface is a CNI artifact that gets recreated when pods are scheduled, making its restoration timing-dependent.

### Recommendation
Retest with `/retest pull-kubernetes-nmstate-e2e-handler-k8s`. The cni0 interface restoration after node reboot is timing-dependent and not deterministic.

</details>

File: cmd/handler/main.go [hunk 6]
```diff
@@ -143,37 +145,36 @@ func mainHandler() int {
...
+	// unavailable (e.g. during default interface reconfiguration).
+	isOpenShift := cluster.IsOpenShiftFromEnv()
+
+	var platformTLSOpts func(*tls.Config)
+	// Only load TLS profile on OpenShift for components that serve TLS
+	// (webhook and metrics). The handler DaemonSet uses default TLS.
+	// Read from a ConfigMap-mounted file to avoid API server calls.
+	// TLS profile changes are handled by the operator: it updates the
+	// ConfigMap and rolls the Deployments via a hash annotation.
+	if isOpenShift && !environment.IsHandler() {
```
mkowalski, line 159: The `!environment.IsHandler` here means that `/metrics` endpoint on the handler pod is losing this configuration. Is that ok or not?

Technically, if we serve something via HTTPs we have to respect this configuration. If we serve something only via HTTP then we do not have to move it to HTTPs (yet).

mkowalski:
/lgtm

As discussed in https://redhat-internal.slack.com/archives/CP7329Z5Z/p1778675352529169, let's disable metrics on handler pod as a subsequent PR

PR #1496, opened 2026-04-14
Move developer documentation to dedicated guide

emy:

>
> /kind bug

/kind enhancement

  Reorganize developer documentation by creating a structured developer guide
  in docs/content/developer-guide/ with the following pages:

  - 101-getting-started.md: Setup and prerequisites

  - 102-testing.md: Unit tests, e2e tests, and static checks

  - 103-local-cluster.md: Local development cluster setup

  - 104-code-generation.md: Code generation and manifests

  - 105-advanced.md: Advanced topics and troubleshooting

  Update CONTRIBUTING.md to reference the new guide instead
  of duplicating detailed developer instructions. This improves discoverability
  and maintainability of developer documentation.

**Release note**:

```release-note
NONE
```


docs/content/developer-guide/101-getting-started.md
```diff
@@ -0,0 +1,87 @@
...
+title: "Getting Started"
+weight: 10
+type: docs
+---
+
+This page covers the prerequisites and basic building instructions for kubernetes-nmstate.
+
+## Prerequisites
+
+- Go 1.21 or later
```
mkowalski: I would not put any version here because you will become responsible for updating it all the time. Smart people can read `go.mod` and infer what's the required version. Less smart people can rely on their IDE to figure it out

docs/content/developer-guide/103-local-cluster.md
```diff
@@ -0,0 +1,59 @@
...
+title: "Local Development Cluster"
+weight: 30
+type: docs
+---
+
+This page covers setting up and using a local development cluster for kubernetes-nmstate.
+
+## Overview
+
+For detailed information about deploying a local virtualized cluster, see the [local virtualized cluster guide]({{< relref "/deployment/local-cluster" >}}).
```
mkowalski, line 15: I would like to see in this section very clearly marked that we are blindly using kubevirtci (or whatever else that is). Otherwise we will be (sooner or later) accused of breaking things up which are not really our things.

Make it clear that this guide is just a courtesy so that the reader does not need to learn about kubevirtci or devscripts or whatnot

docs/content/developer-guide/104-code-generation.md
```diff
@@ -0,0 +1,56 @@
...
+make gen-rbac
+
+# Run all generators
+make generate
+
+# Verify generated code is up to date
+make check-gen
+```
+
+**Important**: Always run `make generate` after modifying types in `api/v1` or `api/v1beta1`, or after changing controller RBAC markers.
```
mkowalski, line 28: You may also note that there is a Claude skill responsible for validating if API changes are correct. But OK to never mention that.

docs/content/developer-guide/104-code-generation.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Manifests and Deployment
```
mkowalski, line 32: In this section you need to make clear that "manifests and deployment" refer to the OLM framework because it's not a standard way in upstream projects.

By default people may think it's Helm what we are doing, but we don't do Helm yet (or ever)

qinqon:
@emy I think we have break the webpage 

https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/1496/pull-kubernetes-nmstate-docs/2043981380903243776/artifacts/gh-pages/index.html

This is from the docs CI logs, at the end you have an URL you can visit.

emy:
@qinqon Strange. On local it all worked. Is there any special things I need to be aware of that is different when CI is deploying it?

emy:
After some investigation on the build process locally I checked up with the preview links on the other pulls and noticed these have all the same problem. (Example: https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/nmstate_kubernetes-nmstate/1474/pull-kubernetes-nmstate-docs/2041896240894971904/artifacts/gh-pages/index.html)

PR #1504, opened 2026-04-22
Add periodic resync to operator reconciler

mkowalski:

/kind bug

**Problem**: The operator only reconciles on NMState CR, Deployment, or DaemonSet changes. If the `openshift.io/node-selector: ""` namespace annotation is removed externally, the reconciler never fires and pods remain stuck in Pending state on clusters with `defaultNodeSelector` configured.

**Fix** (belt and suspenders, per review):
- Returns `ctrl.Result{RequeueAfter: 5 * time.Minute}` from `Reconcile`, adding a periodic resync similar to what CNO does with its 3-minute resync timer
- Watches the handler namespace via `Watches` + `EnqueueRequestsFromMapFunc`, mapping its events to NMState reconcile requests so external modifications are reverted immediately
- Restricts the manager cache for Namespace objects to `metadata.name == HANDLER_NAMESPACE`, so the operator does not cache/watch every namespace in the cluster

This ensures that externally modified resources are restored to their desired state — immediately on namespace events, and at worst within the resync period for anything else.

Ref: https://redhat.atlassian.net/browse/OCPBUGS-67277

Inspired by CNO's `ResyncPeriod` approach. The 5-minute interval is conservative enough to not cause excessive API server load but responsive enough to fix the issue within a reasonable time window.

**Release note**:
```release-note
The NMState operator now watches its handler namespace and periodically reconciles its managed resources every 5 minutes, ensuring that externally modified namespace annotations (such as openshift.io/node-selector) are restored. This fixes an issue where pods could get stuck in Pending state on clusters with defaultNodeSelector configured if the annotation was removed.
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -25,11 +25,13 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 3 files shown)


qinqon (changes requested):
@mkowalski I would do a belt and suspenders and do both, Namespace reconcile (you have to be sure that we filter the cache to just our namespace) and the 5 min thing you have already do, this way we have the best of both words.

qinqon:
@mkowalski let's re-purpose this PR to watch for the kubernetes-nmstate namespace instead.

mkowalski:
@qinqon done — kept the 5 min periodic resync and added a `Watches` on the handler namespace mapping events to NMState reconcile requests. The manager cache is filtered with a `metadata.name == HANDLER_NAMESPACE` field selector for Namespace objects, so we only cache/watch our own namespace. Added unit tests for the mapping function.

This message was generated using AI. Please verify before acting on it.

mkowalski:
Hey @qinqon, let's try moving this one forward. I applied your last comment and hopefully we are ready to go

mkowalski:
/approve

Human review, looks good.

PR #1505, opened 2026-04-22
Strip verbose fields from VLAN interfaces in large NodeNetworkState

mkowalski:

/kind bug

When a node has a large number of interfaces (e.g., 1500+ VLANs), the NodeNetworkState object exceeds the etcd request size limit (1.5 MB), causing `etcdserver: request is too large` errors. This breaks NNS updates entirely, which cascades into empty NNCP status/reason fields.

This PR extends the existing `filterOut()` function to strip verbose fields from VLAN interfaces when the total interface count exceeds 500. For each VLAN interface, only essential fields are preserved:

| Kept | Stripped |
|------|---------|
| `name` | `mac-address` |
| `type` | `mtu` |
| `state` | `lldp` |
| `ipv4` | `ethtool` |
| `ipv6` | `accept-all-mac-addresses` |
| `vlan` (id + base-iface) | `wait-ip`, `description`, etc. |

This reduces per-VLAN size from ~300 bytes to ~100 bytes. For 1500 VLANs, that is ~450 KB → ~150 KB, safely under the etcd limit.

**Key design decisions**:
- Only VLAN interfaces are affected — ethernet, bridge, bond, etc. retain all fields
- Filtering only activates above 500 interfaces — small deployments are completely unaffected
- IP addresses (ipv4/ipv6) are preserved since they are essential for network observability
- The threshold is a constant that can be tuned if needed

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-62470

The etcd limit is 1.5 MB per request. A VLAN interface with full nmstate output is ~250-300 bytes of YAML. At 1500 VLANs, the NNS easily exceeds the limit. The upstream nmstate library also has `MAX_SUPPORTED_INTERFACES = 1000` but that is a separate issue (desired state limit, not observed state).

**Release note**:
```release-note
NodeNetworkState now strips verbose fields (mac-address, mtu, lldp, ethtool) from VLAN interfaces when the total interface count exceeds 500, preventing etcd request size limit errors on nodes with large numbers of VLANs. IP addresses and VLAN identity fields are preserved.
```

### DIFF
--- a/pkg/state/filter.go
+++ b/pkg/state/filter.go
@@ -220,6 +220,50 @@ func isUnmanaged(ifaceData map[string]interface{}) bool {
 	return ifaceData["state"] == "ignore"
 }
 
+const (
+	// interfaceCountThreshold is the number of interfaces above which
+	// verbose fields are stripped from VLAN interfaces to reduce the
+	// NodeNetworkState object size and avoid exceeding the etcd request
+	// size limit (1.5 MB).
+	interfaceCountThreshold = 500
+
+	// vlanInterfaceType is the nmstate type string for VLAN interfaces.
+	vlanInterfaceType = "vlan"
+)
+
+// vlanEssentialFields is the set of fields to preserve on VLAN interfaces
+// when the interface count exceeds the threshold. All other fields are
+// stripped to reduce serialized size.
+var vlanEssentialFields = map[string]struct{}{
+	"name":  {},
+	"type":  {},
+	"state": {},
+	"ipv4":  {},
+	"ipv6":  {},
+	"vlan":  {},
+}
+
+// stripVerboseVlanFields removes non-essential fields from VLAN interfaces
+// when the total interface count exceeds the threshold. This prevents the
+// NodeNetworkState from exceeding the etcd request size limit on nodes
+// with large numbers of VLANs.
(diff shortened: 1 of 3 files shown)


qinqon:
@mkowalski can we fix CI first before review it ? I think it's related.

mkowalski:
I think no because it failed in the test case with small amount of VLANs

```
   STEP: Verifying VLAN interfaces have all fields (below threshold) - /tmp/knmstate/kubernetes-nmstate/test/e2e/handler/nns_vlan_filtering_test.go:256 @ 06/24/26 14:02:46.953
  [FAILED] Expected
      <string>: ignore
  to equal
      <string>: up
```

Anyway will need to look someday because I want this in

qinqon:
> I think no because it failed in the test case with small amount of VLANs
> 
> ```
>    STEP: Verifying VLAN interfaces have all fields (below threshold) - /tmp/knmstate/kubernetes-nmstate/test/e2e/handler/nns_vlan_filtering_test.go:256 @ 06/24/26 14:02:46.953
>   [FAILED] Expected
>       <string>: ignore
>   to equal
>       <string>: up
> ```
> 
> Anyway will need to look someday because I want this in

I think is just the e2e tests there since it expect "state: up" for the vlans but since the tests is creating htem with ip link NNS has "state: ignore" for the vlans.

PR #1514, opened 2026-05-15
tls: propagate cluster TLS adherence policy

mkowalski:
Read spec.tlsAdherence from the OpenShift APIServer resource alongside the TLS security profile.

Propagate the adherence policy through the generated TLS ConfigMap, include it in the rollout hash, and watch it for changes so dependent deployments are restarted when either the profile or adherence policy changes.

Missing or empty tlsAdherence is tolerated as the zero value to support clusters where the TLSAdherence feature gate is disabled or the field is absent.

AI-Model: Claude Opus 4.7

>
> /kind bug
/kind enhancement

**Release note**:

```release-note
For OpenShift clusters we now parse TLSAdherencePolicy and include it in the nmstate-tls-config configmap
```

### DIFF
--- a/cmd/handler/main.go
+++ b/cmd/handler/main.go
@@ -68,12 +68,14 @@ import (
 	nmstatelog "github.com/nmstate/kubernetes-nmstate/pkg/log"
 	"github.com/nmstate/kubernetes-nmstate/pkg/monitoring"
 	"github.com/nmstate/kubernetes-nmstate/pkg/nmstatectl"
+	nmstatetls "github.com/nmstate/kubernetes-nmstate/pkg/tls"
 	"github.com/nmstate/kubernetes-nmstate/pkg/webhook"
 )
 
 const (
-	generalExitStatus    int    = 1
-	tlsProfileConfigPath string = "/etc/nmstate/tls/profile.json"
+	generalExitStatus      int    = 1
+	tlsProfileConfigPath   string = "/etc/nmstate/tls/profile.json"
+	tlsAdherenceConfigPath string = "/etc/nmstate/tls/adherence"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


cmd/handler/main.go
```diff
@@ -163,6 +165,24 @@ func mainHandler() int {
...
+		// kubernetes-nmstate already honors the cluster TLS profile
+		// regardless of the adherence policy, so this is logged for
+		// observability only — but the value is also exposed via
+		// ShouldHonorClusterTLSProfile for any future code path that
+		// needs to gate behavior on it.
+		adherence, err := cluster.FetchTLSAdherenceFromFile(tlsAdherenceConfigPath)
+		if err != nil {
+			setupLog.Error(err, "unable to load TLS adherence from file")
+			return generalExitStatus
+		}
```
qinqon, line 181: This should not make handler fail, if it cannot read the file we are just using to print it.

pkg/tls/tls.go
```diff
@@ -107,6 +141,18 @@ type tlsSecurityProfile struct {
...
+// "" (no opinion) — this keeps callers safe on clusters where the feature
+// gate is disabled or the field is not yet present.
+func parseTLSAdherence(obj map[string]interface{}) TLSAdherencePolicy {
+	value, found, err := unstructured.NestedString(obj, "spec", "tlsAdherence")
+	if err != nil || !found {
+		return ""
+	}
+	return TLSAdherencePolicy(value)
+}
+
```
qinqon, line 155: Do the proper thing with the error return it and take action or at least print it.

pkg/tls/controller.go
```diff
@@ -35,15 +35,22 @@ import (
 )
 
 // SecurityProfileWatcher watches the APIServer object for TLS profile changes
-// and triggers a callback when the profile changes.
+// and triggers a callback when the profile or the cluster-wide tlsAdherence
+// policy changes.
```
qinqon, line 39: I don't see the SecurityProfileWatcher being use at all, maybe we should delete it ?

qinqon (changes requested):

pkg/tls/controller.go
mkowalski: ACK, will check. I thought it's used somewhere but it's been long ago

qinqon:
Leet's keep the changes at the fork branch, since we still don't need that and this PR just print a log, it has no effect on TLS.

Issue #1523, opened 2026-05-29
CI Failure: periodic-knmstate-e2e-handler-k8s-latest / The `periodic-knmstate-e2e-handler-k8s-latest` CI job


qinqon:
Periodic CI job **periodic-knmstate-e2e-handler-k8s-latest** failed in run [2060179517325971456](https://storage.googleapis.com/kubevirt-prow/logs/periodic-knmstate-e2e-handler-k8s-latest/2060179517325971456/build-log.txt).

## Summary

The `periodic-knmstate-e2e-handler-k8s-latest` CI job failed with 2 test failures in the linux bridge E2E tests. The first test ("linux bridge up with disabled VLAN") timed out waiting for a NodeNetworkConfigurationPolicy to reach a terminal state during `BeforeEach`. This left a stale bridge (`br68`) on `node03`, which then caused the second test ("linux bridge up / vlan_filtering 1") to fail in the global `AfterEach` cleanup because the node's interface state didn't match the initial snapshot.

## Root Cause

The causal chain is:

1. **Trigger**: The latest commit `a5f1ecbc3` bumped kubevirtci from `k8s-1.32` to `k8s-1.34` (and the KUBEVIRTCI_TAG from `2602051029-2d22c9bb` to `2605280038-4074cc0d`). This is the first CI run on the new infrastructure version. Additionally, this periodic job runs with the latest bleeding-edge nmstate version from copr (as indicated by the job name `periodic-knmstate-e2e-handler-k8s-latest`).

2. **Primary failure**: The test at `test/e2e/handler/simple_bridge_and_bond_test.go:150-152` calls `updateDesiredStateAndWait(linuxBrUpWithDisabledVlan(bridge1))` in its `BeforeEach`. This calls through to `policy.WaitForPolicy()` (`test/e2e/policy/conditions.go:208-219`) which polls `StatusForPolicyEventually` with a 600-second `RollbackTimeout` waiting for the NNCP to reach a terminal condition (Available, Degraded, or Ignored). The policy **never reached a terminal state** within 600 seconds, causing the timeout at `conditions.go:210`. This indicates that the network configuration application (creating a linux bridge with disabled VLAN on the secondary NICs) hung or was extremely slow on the new k8s-1.34 / latest-nmstate environment.

[...]

PR #1526, opened 2026-06-02
automation: detach emulated igb before handler e2e

oshoval:

/kind bug

Prevent handler e2e runs from interacting with the emulated igb NIC by detaching only worker devices that match the expected igb PCI ID before tests start. This keeps the CI path aligned with the manual mitigation and fails fast if the target device detection is ambiguous.

See https://github.com/nmstate/kubernetes-nmstate/issues/1523#issuecomment-4590140581 seems there is kernel bug that cause spin lock we saw only here. Until we fix it from the root, lets dettach the igb.

cc @qinqon 

**Release note**:

```release-note
None
```

### DIFF
--- a/automation/check-patch.e2e-k8s.sh
+++ b/automation/check-patch.e2e-k8s.sh
@@ -62,6 +62,8 @@ main() {
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
     make cluster-sync
 
+    ./hack/detach-igb-workers.sh
+
     make E2E_TEST_TIMEOUT=2h E2E_TEST_ARGS="--no-color --output-dir=$ARTIFACTS --junit-report=junit.functest.xml" test-e2e-handler
 }
 
--- a/hack/detach-igb-workers.sh
+++ b/hack/detach-igb-workers.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -euo pipefail
+
+workers="$(./cluster/kubectl.sh get nodes -l node-role.kubernetes.io/worker -o jsonpath='{.items[*].metadata.name}')"
+if [ -z "${workers// }" ]; then
+    echo "No worker nodes found via label node-role.kubernetes.io/worker"
+    exit 1
+fi
+
+read -r -d '' remote_script <<'EOF' || true
+set -euo pipefail
+
+matches="$(
+for ifpath in /sys/class/net/*; do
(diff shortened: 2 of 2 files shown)


oshoval:
closing 
see https://github.com/nmstate/kubernetes-nmstate/pull/1525

qinqon:
Hey @oshoval, I dug into this since it overlaps with #1525, and I think this approach will hit the same wall my earlier `install igb /bin/true` attempt did: **it won't survive the node reboots the handler e2e suite performs.**

The detach uses a runtime PCI driver unbind:
```sh
echo "$pci" > /sys/bus/pci/devices/$pci/driver/unbind
```
That's an in-memory sysfs operation, not persistent. And `hack/detach-igb-workers.sh` runs **once**, after `make cluster-sync` and before the suite starts — it does not re-run during the tests.

The handler suite reboots nodes in several specs (`default_bridged_network`, `default_ovs_bridged_network`, `vrf`, `nncp_cleanup`, `bonding_default_interface`) via `restartNodeWithoutWaiting` → `sudo halt --reboot` (`test/e2e/handler/utils.go:334`). So the sequence becomes:

1. Detach runs → eth3 gone → `main_test.go` captures initial state **without** eth3.
2. A reboot spec runs `halt --reboot`.
3. kubevirtci boots each node from an immutable external kernel+initrd (`qemu -kernel/-initrd`); that initrd loads igb and PCI coldplug **re-binds** it → **eth3 reappears**.
4. Nothing re-detaches → `main_test.go` "didn't return to initial state" (len:7 with eth3 vs initial len:6) → suite fails.

This is exactly the eth3-reappears-after-reboot failure I traced in the failing periodic builds.

In #1525 I worked around it by installing a systemd oneshot unit on the node **rootfs** (`WantedBy=multi-user.target`) that runs `modprobe -r igb` on **every** boot, so it re-fires after each in-suite reboot. The rootfs is the only thing that persists, since the kernel/initrd are immutable provider files.

If you'd prefer to keep the PCI-unbind path (it's cleaner and your fail-fast-on-ambiguous-detection check is nice), it'd just need to be made reboot-persistent — e.g. wrap the detach in a systemd unit that re-runs each boot. Happy to reconcile the two PRs into one if you want. cc @qinqon

oshoval:
nice find, anyhow closed, lets continue with yours, as long as it bullet proof, its better and simpler

PR #1535, opened 2026-06-09
build: consolidate multi-arch container build scripts — replace buildah with native podman

### LINKED ISSUE #1532: build: consolidate multi-arch container build scripts — replace buildah with native podman
## Problem

The multi-arch container build infrastructure uses a dispatch matrix of 3 scripts based on `${IMAGE_BUILDER}.${HOST_OS}`:

| `IMAGE_BUILDER` | `HOST_OS` | Script | Actual tool |
|---|---|---|---|
| `podman` | `linux` | `hack/build-push-container.podman.linux.sh` | **buildah** |
| `podman` | `darwin` | `hack/build-push-container.podman.darwin.sh` | **podman** |
| `docker` | `linux` | `hack/build-push-container.docker.linux.sh` | **docker buildx** |
| `docker` | `darwin` | (missing) | **FAILS** |

Issues:

1. **`podman.linux.sh` uses buildah, not podman** — misleading name, adds an unnecessary explicit CLI dependency. Modern podman (4.x+) natively supports `podman build --manifest` and `podman manifest push --all`.
2. **`podman.darwin.sh` already uses native podman** — proving it works. The only difference from the Linux variant is the QEMU setup (not needed on macOS since podman machine handles it).
3. **No `docker.darwin.sh`** — Docker Desktop on macOS would fail silently with "file not found".
4. **Duplicated logic across all scripts** — same skip checks, TLS verify, QEMU setup, arch loop, manifest push.

## Proposed Fix

Consolidate the podman linux and darwin scripts into a single `hack/build-push-container.podman.sh`, replacing `buildah` with native `podman` commands and conditionally calling QEMU setup only when needed (Linux cross-arch builds). The Makefile dispatch would change from `${IMAGE_BUILDER}.${HOST_OS}` to just `${IMAGE_BUILDER}`.

Specifically:

- **Merge `podman.linux.sh` + `podman.darwin.sh`** into one `podman.sh` using native `podman` commands (as the darwin script already does), with QEMU setup gated on `uname -s` + cross-arch detection
- **Optionally add `docker.darwin.sh`** or merge `docker.linux.sh` into a unified `docker.sh` (docker buildx works the same on both platforms)
- **Update Makefile** dispatch from `hack/build-push-container.${IMAGE_BUILDER}.${HOST_OS}.sh` to `hack/build-push-container.${IMAGE_BUILDER}.sh`
- **Remove `HOST_OS` variable** from Makefile if no longer needed elsewhere

### `hack/build-push-container.podman.linux.sh` (uses buildah)

```bash
buildah rmi ${IMAGE} 2>/dev/null || true
buildah manifest rm ${IMAGE} 2>/dev/null || true
buildah manifest create ${IMAGE}
for arch in $ARCHS; do
    buildah bud \
        --manifest ${IMAGE} \
        --arch ${arch} --build-arg TARGETARCH=${arch} $@ --tag ${IMAGE}.${arch}
done
buildah manifest push --all ${IMAGE} docker://${IMAGE} --tls-verify=${TLS_VERIFY}
```

### `hack/build-push-container.podman.darwin.sh` (uses native podman — same commands work on Linux)

```bash
podman rmi ${IMAGE} 2>/dev/null || true
podman manifest rm ${IMAGE} 2>/dev/null || true
podman manifest create ${IMAGE}
for arch in $ARCHS; do
    podman build \
        --manifest ${IMAGE} \
        --arch ${arch} --build-arg TARGETARCH=${arch} $@ --tag ${IMAGE}.${arch} ./
done
podman manifest push --all ${IMAGE} docker://${IMAGE} --tls-verify=${TLS_VERIFY}
```

The darwin script already proves native podman commands work for multi-arch builds. The linux script can adopt the same approach, adding only the QEMU binfmt setup for cross-arch builds.

## References

- PR #1530 added s390x to ARCHS and removed `hack/init-buildx.sh` (docker buildx is now stable)
- Issue #1531 exposed the copr chroot detection bug in multi-arch builds

/kind enhancement

qinqon:
Fixes #1532

/kind enhancement

The multi-arch container build infrastructure dispatched to one of several
scripts keyed on `${IMAGE_BUILDER}.${HOST_OS}`. That layout had several
problems:

- `build-push-container.podman.linux.sh` actually invoked `buildah`, not
  `podman` — a misleading name that also added an unnecessary explicit CLI
  dependency. Modern podman (4.x+) natively supports `podman build --manifest`
  and `podman manifest push --all`, which the existing darwin script already
  used successfully.
- There was no `docker.darwin.sh`, so Docker Desktop on macOS failed with a
  "file not found" error.
- The per-OS variants duplicated the same skip checks, TLS verify, QEMU setup,
  arch loop and manifest push logic.

This PR consolidates the scripts:

- Merges the podman linux and darwin variants into a single
  `hack/build-push-container.podman.sh` built on native `podman` commands. The
  QEMU `binfmt` registration is now gated on Linux cross-architecture builds
  only; on macOS the podman machine already provides the emulation.
- Renames `docker.linux.sh` to `hack/build-push-container.docker.sh` and applies
  the same Linux gate, so Docker Desktop on macOS (which provides its own
  emulation) no longer runs the `qemu-user-static` container and the docker
  path works on both platforms.
- Simplifies the Makefile dispatch from `${IMAGE_BUILDER}.${HOST_OS}.sh` to
  `${IMAGE_BUILDER}.sh` and drops the now-unused `HOST_OS` variable.

- The new `podman.sh` is derived from the previous darwin script (native
  podman), so git tracks it as a rename; the buildah-based `podman.linux.sh` is
  removed. `docker.sh` is a rename of `docker.linux.sh` with only the QEMU gate
  changed.
- Behavior is unchanged for the common Linux CI path: cross-arch builds still
  run `hack/qemu-user-static.sh` before building.
- The dead `IMAGES=${IMAGE}` assignment (never referenced) was dropped while
  merging the podman scripts.
- Verified the Makefile dispatch resolves to the existing scripts via
  `make -n push-handler IMAGE_BUILDER=podman` and
  `make -n push-operator IMAGE_BUILDER=docker`, and that both scripts pass
  `bash -n`.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -183,13 +181,13 @@ handler: SKIP_PUSH=true
 handler: push-handler
 
 push-handler:
-	SKIP_PUSH=$(SKIP_PUSH) SKIP_IMAGE_BUILD=$(SKIP_IMAGE_BUILD) IMAGE=${HANDLER_IMAGE} hack/build-push-container.${IMAGE_BUILDER}.${HOST_OS}.sh ${HANDLER_EXTRA_PARAMS} --build-arg GO_VERSION=$(GO_VERSION) -f build/Dockerfile
+	SKIP_PUSH=$(SKIP_PUSH) SKIP_IMAGE_BUILD=$(SKIP_IMAGE_BUILD) IMAGE=${HANDLER_IMAGE} hack/build-push-container.${IMAGE_BUILDER}.sh ${HANDLER_EXTRA_PARAMS} --build-arg GO_VERSION=$(GO_VERSION) -f build/Dockerfile
```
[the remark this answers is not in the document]
qinqon: > If `IMAGE_BUILDER` is not set or cannot be auto-detected (e.g., neither `podman` nor `docker` is running), the recipe will fail with a confusing `No such file or directory` error when trying to execute `hack/build-push-container..sh`. Adding a guard to check if `IMAGE_BUILDER` is set provides a much clearer error message.

Addressed. Rather than duplicating the inline guard in both `push-handler` and `push-operator`, I added a single shared `require-image-builder` prerequisite target:

```make
require-image-builder:
	@test -n "$(IMAGE_BUILDER)" || { echo "Error: IMAGE_BUILDER is not set and could not be auto-detected (neither podman nor docker is running/available)." >&2; exit 1; }
```

`push-handler` now depends on it (`push-handler: require-image-builder`), so it fails fast with the clear message instead of executing the broken `hack/build-push-container..sh` path. Verified: `make push-handler IMAGE_BUILDER=` prints the error and exits 1, while `IMAGE_BUILDER=podman` passes.

Makefile
```diff
@@ -183,13 +181,13 @@ handler: SKIP_PUSH=true
...
 
 operator: SKIP_PUSH=true
 operator: push-operator
 
 push-operator:
-	SKIP_PUSH=$(SKIP_PUSH) SKIP_IMAGE_BUILD=$(SKIP_IMAGE_BUILD) IMAGE=${OPERATOR_IMAGE} hack/build-push-container.${IMAGE_BUILDER}.${HOST_OS}.sh --build-arg GO_VERSION=$(GO_VERSION) -f build/Dockerfile.operator
+	SKIP_PUSH=$(SKIP_PUSH) SKIP_IMAGE_BUILD=$(SKIP_IMAGE_BUILD) IMAGE=${OPERATOR_IMAGE} hack/build-push-container.${IMAGE_BUILDER}.sh --build-arg GO_VERSION=$(GO_VERSION) -f build/Dockerfile.operator
```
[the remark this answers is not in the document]
qinqon: > Similarly, add a guard to check if `IMAGE_BUILDER` is set before executing the build script to avoid a confusing file-not-found error.

Addressed. `push-operator` now depends on the same shared `require-image-builder` target introduced for `push-handler` (`push-operator: require-image-builder`), so both recipes share one guard and one error message instead of duplicating it. As a prerequisite target it also runs only once per `make push` invocation. Verified the failure path with `make push-operator IMAGE_BUILDER=` (clear error, exit 1).

hack/build-push-container.podman.sh
```diff
@@ -14,10 +14,16 @@ fi
...
 ARCHS=${ARCHS:-$(go env GOARCH)}
 
+# QEMU binfmt registration is only needed for Linux cross-architecture
+# builds. On macOS the podman machine already provides the emulation, so
+# skip the setup there.
+if [ "${ARCHS}" != "$(go env GOARCH)" ] && [ "$(uname -s)" == "Linux" ]; then
+    hack/qemu-user-static.sh
+fi
+
 podman rmi ${IMAGE} 2>/dev/null || true
```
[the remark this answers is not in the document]
qinqon: > Since `set -o nounset` is active, if the `IMAGE` environment variable is not set, the script will crash with a generic `unbound variable` error. Adding an explicit assertion using the shell parameter expansion pattern `: "${IMAGE:?...}"` provides a clear, actionable error message.

Addressed. Added `: "${IMAGE:?IMAGE environment variable must be set}"`. I placed it right after the `SKIP_IMAGE_BUILD` early-exit (rather than just before `podman rmi`) so it fails before the QEMU setup runs, while still allowing `SKIP_IMAGE_BUILD=true` to short-circuit without requiring `IMAGE`. I also applied the same assertion to the sibling `hack/build-push-container.docker.sh`, which shares the identical `IMAGE` contract under `nounset`, to keep both scripts consistent. Verified both now print `IMAGE: IMAGE environment variable must be set` when `IMAGE` is unset, and that `SKIP_IMAGE_BUILD=true` still exits 0 without `IMAGE`.

qinqon:
/oompa fix the commit message prefix from "Fix #1532: build" to  refactor(build):

Makefile
```diff
@@ -179,17 +177,20 @@ generate: gen-k8s gen-crds gen-rbac
 manifests:
 	GOFLAGS=-mod=mod go run hack/render-manifests.go -handler-prefix=$(HANDLER_PREFIX) -handler-namespace=$(HANDLER_NAMESPACE) -operator-namespace=$(OPERATOR_NAMESPACE) -handler-image=$(HANDLER_IMAGE) -operator-image=$(OPERATOR_IMAGE) -handler-pull-policy=$(HANDLER_PULL_POLICY) -monitoring-namespace=$(MONITORING_NAMESPACE) -kube-rbac-proxy-image=$(KUBE_RBAC_PROXY_IMAGE) -operator-pull-policy=$(OPERATOR_PULL_POLICY) -input-dir=deploy/ -output-dir=$(MANIFESTS_DIR)
 
+require-image-builder:
+	@test -n "$(IMAGE_BUILDER)" || { echo "Error: IMAGE_BUILDER is not set and could not be auto-detected (neither podman nor docker is running/available)." >&2; exit 1; }
```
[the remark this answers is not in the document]
qinqon: > `require-image-builder` only checks that `IMAGE_BUILDER` is non-empty. If a user/export sets `IMAGE_BUILDER` to an unsupported value (or a typo), the Makefile will try to execute a non-existent script ... Consider validating that the corresponding script exists (and is executable) here so failures are immediate and self-explanatory.

Good call. `require-image-builder` now also asserts the dispatch script exists and is executable, so an unsupported or typo'd `IMAGE_BUILDER` fails immediately with an actionable message instead of a generic "No such file or directory":

```make
require-image-builder:
	@test -n "$(IMAGE_BUILDER)" || { echo "Error: IMAGE_BUILDER is not set ..." >&2; exit 1; }
	@test -x "hack/build-push-container.$(IMAGE_BUILDER).sh" || { echo "Error: unsupported IMAGE_BUILDER '$(IMAGE_BUILDER)': hack/build-push-container.$(IMAGE_BUILDER).sh not found or not executable (expected 'podman' or 'docker')." >&2; exit 1; }
```

Verified:
- `make push-handler IMAGE_BUILDER=foo` -> `Error: unsupported IMAGE_BUILDER 'foo': ... not found or not executable (expected 'podman' or 'docker').` (exit 1)
- `IMAGE_BUILDER=podman` -> passes the guard.

qinqon:
Thanks for the additional review — follow-ups addressed on this branch:

**`Makefile` (`require-image-builder`)** — resolved in the inline thread: the target now also verifies the dispatch script exists and is executable, so an unsupported/typo'd `IMAGE_BUILDER` fails fast with a clear message.

**Copilot low-confidence suppressed comments (3)** — all valid; the build scripts are now safe to run directly (outside `make`) under `set -o nounset`:
- `hack/build-push-container.podman.sh` & `hack/build-push-container.docker.sh`: default the optional `SKIP_IMAGE_BUILD`/`SKIP_PUSH` vars at the top (`${VAR:-false}`), so a direct invocation reaches the explicit `IMAGE` assertion instead of aborting with an "unbound variable" error.
- `hack/build-push-container.podman.sh`: the registry check now uses `${IMAGE_REGISTRY:-}`, so it no longer trips `nounset` when `IMAGE_REGISTRY` is unset.

@qinqon — acknowledged; updating the commit message prefix from `Fix #1532: build:` to `refactor(build):`.

Verification: changes are Makefile/shell-only (no Go touched), exercised locally with `bash -n`, direct script runs, and `make require-image-builder` for valid/invalid/empty `IMAGE_BUILDER`. Go `lint`/`unit` are covered by CI.

qinqon:
/oompa the commit title prefix didn't change with the changes I ask for.

qinqon:
@qinqon — you're right, and my earlier comment claiming the prefix was "being updated" was wrong: it didn't change, and I'm sorry for the noise.

The reason is structural, not an oversight on the retry. The commit subject on this PR isn't authored in the worktree — the oompa automation generates it when it squashes the work into a single commit, as `Fix #<issue>: <issue title>` (the `Fix #<n>:` prefix is fixed), and every review round folds new changes in with `git commit --amend --no-edit`, which preserves that subject. The agent running inside the PR worktree is explicitly blocked from `git commit` / `--amend` / `push` (the automation owns those), so it can't rewrite the subject from here — and any manual rewrite would just be regenerated by the squash step on the next run.

So `Fix #1532:` → `refactor(build):` has to change on the automation side — the squash message construction (`gitSquashCommits` in oompa's `pkg/agent/git_ops.go`), e.g. deriving a conventional-commit type/scope instead of the hardcoded `Fix #<n>:` prefix. As the operator you can also amend the subject directly on this PR; the target would be:

`refactor(build): consolidate multi-arch container build scripts — replace buildah with native podman`

(keeping the existing body and the `Signed-off-by` / `Assisted-by` trailers).

qinqon:
Rebased commit 1df4875 on main and pushed.

hack/build-push-container.docker.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+PLATFORM=""
+
+for arch in $ARCHS; do
+    PLATFORM=${PLATFORM}linux/$arch,
+done
+
+# Remove last ','
+PLATFORM=${PLATFORM::-1}
```
mkowalski: `${PLATFORM::-1}` (negative-length substring expansion) requires bash >= 4.2. macOS ships bash 3.2 and `#!/usr/bin/env bash` resolves to it on a stock Mac, so this line aborts with `PLATFORM: substring expression < 0` — i.e. exactly the Docker-Desktop-on-macOS path this PR is meant to enable. The POSIX trailing-comma strip works everywhere:

```suggestion
PLATFORM=${PLATFORM%,}
```

(Carried over from the old `docker.linux.sh`, where it was harmless because that script never ran on darwin.)

hack/build-push-container.podman.sh
```diff
@@ -2,22 +2,35 @@
 
 set -x -o errexit -o nounset -o pipefail
 
+# Default optional environment variables so the script can be run directly
+# (outside of the Makefile) without tripping `set -o nounset`.
+SKIP_IMAGE_BUILD=${SKIP_IMAGE_BUILD:-false}
+SKIP_PUSH=${SKIP_PUSH:-false}
```
mkowalski, line 8: The `SKIP_*` defaults make direct (non-make) invocation possible, but on Linux a cross-arch direct run still breaks one step later: `hack/qemu-user-static.sh` does `${IMAGE_BUILDER} run --rm --privileged ...`, and outside make `IMAGE_BUILDER` is unset (the Makefile exports it), so it expands to nothing and fails with a confusing `run: command not found`. Each script knows its builder, so defaulting (and exporting, for the child script) here would complete the hardening:

```suggestion
SKIP_PUSH=${SKIP_PUSH:-false}
# Builder forwarded to hack/qemu-user-static.sh; default it so the script
# also works when invoked directly instead of through the Makefile.
export IMAGE_BUILDER=${IMAGE_BUILDER:-podman}
```

Same applies to `hack/build-push-container.docker.sh` with `docker`.

mkowalski (commented):
Two findings from a closer pass on the consolidated scripts: a macOS portability bug in the docker path, and a small gap in the direct-run hardening. Details inline.

hack/build-push-container.docker.sh
mkowalski: @emy should have a final say if this is reasonable

hack/build-push-container.podman.sh
mkowalski: That one sounds like a fair suggestion

hack/build-push-container.docker.sh
qinqon: > `${PLATFORM::-1}` (negative-length substring expansion) requires bash >= 4.2. macOS ships bash 3.2 and `#!/usr/bin/env bash` resolves to it on a stock Mac, so this line aborts with `PLATFORM: substring expression < 0` ... The POSIX trailing-comma strip works everywhere: `PLATFORM=${PLATFORM%,}`

Fixed — switched to `PLATFORM=${PLATFORM%,}`. Since `PLATFORM` is built as `linux/$arch,` per iteration it always ends in exactly one comma, so `%,` is an exact equivalent of the old `::-1` but POSIX (works on macOS's bash 3.2). Added an inline note pointing at the bash 4.2 requirement so it doesn't get "simplified" back. Verified for 1/2/3 archs:

- `amd64` -> `linux/amd64`
- `amd64 arm64` -> `linux/amd64,linux/arm64`
- `amd64 arm64 s390x` -> `linux/amd64,linux/arm64,linux/s390x`

hack/build-push-container.podman.sh
qinqon: > on Linux a cross-arch direct run still breaks one step later: `hack/qemu-user-static.sh` does `${IMAGE_BUILDER} run --rm --privileged ...`, and outside make `IMAGE_BUILDER` is unset ... defaulting (and exporting, for the child script) here would complete the hardening ... Same applies to `hack/build-push-container.docker.sh` with `docker`.

Fixed in both scripts. Confirmed `hack/qemu-user-static.sh` runs `${IMAGE_BUILDER} run --rm --privileged ...` (lines 25/28) and uses only `set -xe` (no `nounset`), so an unset `IMAGE_BUILDER` would expand to empty and fail with `run: command not found` on a direct cross-arch Linux run. Added right after the `SKIP_*` defaults:

- `hack/build-push-container.podman.sh`: `export IMAGE_BUILDER=${IMAGE_BUILDER:-podman}`
- `hack/build-push-container.docker.sh`: `export IMAGE_BUILDER=${IMAGE_BUILDER:-docker}`

Defaulted to each script's own builder and `export`ed so the child `qemu-user-static.sh` inherits it; an explicit `IMAGE_BUILDER` (e.g. from the Makefile) still wins. Verified: unset -> child process sees the per-script default; a preset value is preserved.

qinqon:
Rebased commit 37ad0e4 on main and pushed.

qinqon (commented):
Change the commit message title prefix to comply with "refactor(build):"

qinqon:
@qinqon — understood, and I know this is the third time it's come up, so let me be straight about the blocker rather than repeat a promise: I can't change the commit title from inside the PR worktree.

The subject isn't written here — the oompa automation generates it in `gitSquashCommits` as `Fix #<issue>: <issue title>` (issue #1532's title is `build: consolidate …`, which is why it renders as `Fix #1532: build:`), and every review round keeps it via `git commit --amend --no-edit`. The PR-side agent is explicitly barred from `git commit` / `--amend` / `push`, and there's no commit-subject override it can set (only `.pr-body.md`, which controls the PR *body*, not the title). So from here the `Fix #1532:` prefix can't be turned into `refactor(build):` — and with no code change to push this round, anything I amended locally wouldn't reach the PR anyway.

Two ways to actually land it:
1. **Automation (durable):** have `gitSquashCommits` honor a conventional-commit prefix already present on the issue title (and move `Fixes #<n>` into the body) instead of always prepending `Fix #<n>:`.
2. **One-off (operator):** amend this commit's subject directly to:
   `refactor(build): consolidate multi-arch container build scripts — replace buildah with native podman`
   (leaving the body and the `Signed-off-by` / `Assisted-by` trailers as-is).

mkowalski:
/lgtm

---
🤖 *This comment was generated by AI on behalf of @mkowalski, who has reviewed it.*

PR #1530, opened 2026-06-04
build: add s390x to multi-arch image builds

ashokpariya0:
Add linux/s390x to the set of architectures built for the handler and operator container images. This enables kubernetes-nmstate to run on s390x arch.

**Is this a BUG FIX or a FEATURE?**:

> Uncomment only one, leave it on its own line:
>
> /kind enhancement

**What this PR does / why we need it**:
Currently the images published to quay.io/nmstate only include amd64 and arm64
manifests:
- https://quay.io/repository/nmstate/kubernetes-nmstate-handler?tab=tags
- https://quay.io/repository/nmstate/kubernetes-nmstate-operator?tab=tags

After this change, both images will include s390x in their manifest lists,
enabling kubernetes-nmstate deployment on s390x arch/cluster.

Changes:
- Consolidate ARCHS definition in automation/check-patch.setup.sh
- Remove duplicate ARCHS exports from publish.sh and release.sh
- Remove hack/init-buildx.sh (docker buildx is now stable)

**Special notes for your reviewer**:

**Release note**:

```release-note
Add s390x architecture support to kubernetes-nmstate handler and operator container images.
```

PR #1538, opened 2026-06-09
build: gracefully handle missing Copr chroot for s390x

mkowalski:
The `nmstate/nmstate-git` Copr project does not have a
`centos-stream-9-s390x` build target, which causes the multi-arch
container build to fail for s390x when `NMSTATE_SOURCE=git`.

PR #1530 added s390x to the `ARCHS` list, but the runtime stage runs
`install-nmstate.git.sh` which unconditionally tries:

```
dnf copr enable -y nmstate/nmstate-git "centos-stream-9-$(uname -m)"
```

Inside the s390x container (qemu emulation), `uname -m` returns `s390x`,
and that Copr chroot does not exist → build fails.

**Fix:** Try to enable the Copr repo, and if the chroot is not available,
fall back to the distro-provided nmstate package. This keeps s390x images
building (with distro nmstate) while amd64 and arm64 continue using the
Copr git builds as before.

Failing job: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-knmstate-e2e-handler-k8s-latest/2064165606369792000

Once the Copr project adds `centos-stream-9-s390x`, that arch will
automatically start picking up the git builds with no further changes.

---
🤖 *This PR was generated by AI on behalf of @mkowalski, who has reviewed it.*

### DIFF
--- a/build/install-nmstate.git.sh
+++ b/build/install-nmstate.git.sh
@@ -2,9 +2,16 @@
 
 dnf install -b -y dnf-plugins-core
 
-# Specify the centos-stream-9 chroot explicitly because dnf copr auto-detection
-# picks epel-9-<arch> on CentOS Stream 9, which does not exist for aarch64 in
-# the nmstate/nmstate-git Copr project.
-dnf copr enable -y nmstate/nmstate-git "centos-stream-9-$(uname -m)"
+arch=$(uname -m)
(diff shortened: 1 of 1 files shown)


qinqon:
Supeseed by:
- https://github.com/nmstate/kubernetes-nmstate/pull/1539

PR #1539, opened 2026-06-09
automation: force archs just for publish/release

qinqon:
After [1] now all the automation e2e where trying to build for all the archs, that's  wrong since for e2e we just need the image for the arch kuberntes is going to run on.

[1] https://github.com/nmstate/kubernetes-nmstate/pull/1530

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.setup.sh
+++ b/automation/check-patch.setup.sh
@@ -23,5 +23,4 @@ export E2E_LOGS=${TMP_PROJECT_PATH}/test_logs/e2e
 export ARTIFACTS=${ARTIFACTS-$tmp_dir/artifacts}
 mkdir -p $ARTIFACTS
 
-export ARCHS="${ARCHS:-amd64 arm64 s390x}"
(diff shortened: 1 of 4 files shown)


mkowalski:
Clean approach — e2e jobs only need the native arch image, multi-arch builds belong in publish/release.

Verified:
- All scripts that source `check-patch.setup.sh` (e2e, unit tests) will now fall back to `ARCHS=$(go env GOARCH)` from `build-push-container.podman.linux.sh` → just `amd64`. Correct for CI.
- `publish.sh` and `release.sh` source the new `publish-release.env` → get the full `amd64 arm64 s390x` set. Correct for releases.

Note: the publish/release path will still hit the missing `centos-stream-9-s390x` Copr chroot when building with `NMSTATE_SOURCE=git` (same pre-existing issue from #1530), but that is out of scope here and can be addressed separately.

/lgtm

---
🤖 *This comment was generated by AI on behalf of @mkowalski, who has reviewed it.*

qinqon:
/retest

Don't know why this is not runnning

PR #1550, opened 2026-07-09
OCPBUGS-98067: Validate maxUnavailable via CRD CEL rule

emy:

>
/kind bug
> /kind enhancement

---
**What**

Move maxUnavailable validation from controller logic to a CRD-level CEL validation rule, so invalid values are rejected at admission time with a clear message instead of silently falling back to a default during reconciliation.

**Why**

Previously an invalid maxUnavailable (e.g. 0, 0%, or a value scaling to zero nodes) was only caught at reconcile time and quietly defaulted. A value that scales to zero makes the rollout treat its limit as reached on the first node and deadlock. Validating at the API boundary gives users immediate, actionable feedback and prevents the bad value from ever being persisted.

**Changes**

- CEL validation rule on NodeNetworkConfigurationPolicy.spec.maxUnavailable: accepts a positive integer or a percentage between 1% and 100%. Rejects 0, 0%, negative values, non-numeric strings, and percentages > 100%. Absolute counts remain uncapped by design.
- Runtime handling (ScaledMaxUnavailableNodeCount): accepts both percentage strings ("50%") and bare-integer strings ("5", treated as an absolute node count). Any value that still resolves to a non-positive count — only reachable for policies persisted before the CRD upgrade — returns a terminal InvalidMaxUnavailableError.
- Terminal failure path: an invalid value fails the enactment (Failing=True, reason FailedToConfigure) with a clear message and no requeue — the desired state is never applied and no fallback value is silently used.
- Regenerated CRDs and OLM bundle manifests.

**Release note**:

```release-note
Reject invalid maxUnavailable values (0, 0%, negatives, or percentages over 100%) at API admission with a clear error message.
```

### DIFF
--- a/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -46,7 +46,7 @@ type NodeNetworkConfigurationPolicySpec struct {
 
 	// MaxUnavailable specifies percentage or number
 	// of machines that can be updating at a time. Default is "50%".
-	// The computed value must result in at least 1 node being available for updates.
+	// The computed value must result in at least 1 node being allowed to undergo updates at a time.
 	// If set to 0 or a percentage that rounds to 0, the policy will fail with FailedToConfigure.
 	// +optional
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
--- a/api/shared/nodenetworkconfigurationpolicy_types.go
+++ b/api/shared/nodenetworkconfigurationpolicy_types.go
@@ -46,6 +46,8 @@ type NodeNetworkConfigurationPolicySpec struct {
 
 	// MaxUnavailable specifies percentage or number
 	// of machines that can be updating at a time. Default is "50%".
+	// The computed value must result in at least 1 node being available for updates.
+	// If set to 0 or a percentage that rounds to 0, the policy will fail with FailedToConfigure.
 	// +optional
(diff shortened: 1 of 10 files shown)


qinqon:
@emy why we don't implement this as a CEL validation of  > 1 for MaxUnavailable ?

mkowalski:
> @emy why we don't implement this as a CEL validation of > 1 for MaxUnavailable ?

You cannot make CEL validation stricter once you introduced API because your already existing objects won't be forward-compatible. Correct me where I am wrong

qinqon:
> > @emy why we don't implement this as a CEL validation of > 1 for MaxUnavailable ?
> 
> You cannot make CEL validation stricter once you introduced API because your already existing objects won't be forward-compatible. Correct me where I am wrong

well same thing will happend if we do it at the controller, the old NNCP will have NNCP status with failed to configure, doing it at CEL level is more aparent (maybe customer's script are not checking the status so they have a fail possitive).

mkowalski:
Yeah actually over all those years I never saw anyone using `maxUnavailable: 0`. Whatever you wish, CEL is fine. If there is really someone with value `0` we can handle him like snowflake.

qinqon:
> Yeah actually over all those years I never saw anyone using `maxUnavailable: 0`. Whatever you wish, CEL is fine. If there is really someone with value `0` we can handle him like snowflake.

Let's go with CEL it will make issue more aparent and there wil not be false possitive of users not checking status but finding that networking is not "networketing".

PR #1571, opened 2026-08-13
OCPBUGS-74261: Fix NNCP MaxUnavailableLimitReached deadlock with unavailable-slot audit

mkowalski:

/kind bug

Fixes the NNCP `MaxUnavailableLimitReached` deadlock family (OCPBUGS-74261, OCPBUGS-90538). `UnavailableNodeCountMap` is maintained by blind `+1`/`-1` around each apply; any interruption between the two (ungraceful reboot, handler crash, a failed status write right after the node reconfigured its own networking) leaves a ghost slot. Once the stale count reaches `maxUnavailable`, every enactment is refused forever (`Pending`/"Waiting for progressing nodes to finish") and the only runtime path that clears the map — full policy success — is unreachable. Recreating the NNCP was the only recovery. #1427/#1542 heal some cases at handler startup, but nothing heals a ghost slot created while the handler keeps running, the startup `List` was unretried (it fails silently exactly in the post-reboot window it targets), and the broadened `!IsAvailable` startup decrement could over-free slots held by other nodes.

This PR makes the counter self-auditing — enactments are the source of truth, the counter is only the lock:

- **Release before success** (`cbd8607cf`): the reconcile now decrements the slot *before* marking the enactment `Available`, with a ~30s authoritative retry budget. A failed release leaves the enactment truthfully `Progressing`; the poisonous `Available`+held-slot state is structurally unreachable. An enactment already `Progressing` for the current generation skips re-claiming (it holds the slot from an interrupted reconcile).
- **Audit-on-block** (`8db393bb0`, `e1356f75f`): when a claim is refused, the counter is recomputed from live holders (current-generation enactments with `Progressing=True` and a heartbeat younger than `NMSTATE_ENACTMENT_STALE_THRESHOLD`, default 15m > worst-case apply cycle) and repaired set-to-truth — never blind decrements — guarded by a 30s grace window on the (previously dead) `LastUnavailableNodeCountUpdate` field. Blocked reconciles requeue within 90–120s, so recovery is bounded even on a quiet cluster. Policies already deadlocked before an upgrade heal on their first blocked reconcile (nil timestamp = no grace).
- **Startup reclaim** (`79f3b907b`, `0dbee3c0c`): on handler start, this node's own still-`Progressing` enactments (provably dead) are marked `Pending` ("interrupted by handler restart"), their retry counts reset, and their policies audited. The initial `List` is retried ~2min. All blind-decrement cleanup (`cleanStaleUnavailableCounts` and friends) is deleted, eliminating the `!IsAvailable` over-free.
- **Progressing-write failure handling** (`d7677db8a`): if the `Progressing` condition cannot be persisted after a successful claim, the slot is released and the reconcile requeues — closing the one path where an invisible holder could be repaired away by another node's audit.
- **Tests**: unit specs for every audit semantics case (ghost repair, live holder, staleness, grace, generation change, idempotency), reconcile-level regression tests replaying the deadlock via fake-client interceptors, and an e2e test killing a handler mid-apply and asserting the policy converges without NNCP recreation (`614c63000`, `57034697f`).

No API schema changes; `maxUnavailable` semantics for healthy clusters are unchanged (the audit runs only when a claim is refused).

- The audit deliberately repairs only the *current* generation's map entry; stale old-generation entries remain until the policy's success-path map reset (cosmetic, follow-up candidate).
- `applyDesiredStateFn` is a test seam mirroring the existing `nmstatectlShowFn` pattern.
- The e2e test is flake-safe: if the pod kill lands after the target node finished applying, the test still asserts meaningful convergence with a zeroed counter map.
- Development note: this change was developed with AI assistance (Claude Fable 5 via OpenCode); each commit carries an `Assisted-By` trailer. Design/code were human-directed and the full unit suite plus vet were verified on the final tree. Please verify before acting on any claim herein.

**Release note**:

```release-note
Fixed a deadlock where NodeNetworkConfigurationPolicies could remain Progressing forever with enactments reporting MaxUnavailableLimitReached after an interrupted apply (for example an ungraceful reboot or handler restart). The unavailable-node counter is now audited against live enactments and repaired automatically; recreating the NNCP is no longer necessary.
```

### DIFF
--- a/api/shared/nodenetworkconfigurationenactment_types.go
+++ b/api/shared/nodenetworkconfigurationenactment_types.go
@@ -83,6 +83,13 @@ const (
 	NodeNetworkConfigurationEnactmentConditionMaxUnavailableLimitReached ConditionReason = "MaxUnavailableLimitReached"
 	NodeNetworkConfigurationEnactmentConditionConfigurationProgressing   ConditionReason = "ConfigurationProgressing"
 	NodeNetworkConfigurationEnactmentConditionConfigurationAborted       ConditionReason = "ConfigurationAborted"
+	NodeNetworkConfigurationEnactmentConditionConfigurationInterrupted   ConditionReason = "ConfigurationInterrupted"
+	// NodeNetworkConfigurationEnactmentConditionConfigurationFinalizing marks the
+	// post-apply phase: the desired state was applied and only releasing the
+	// maxUnavailable slot and recording success remain. A reconcile that finds
+	// the enactment in this phase must not re-apply the (already committed)
+	// configuration.
+	NodeNetworkConfigurationEnactmentConditionConfigurationFinalizing ConditionReason = "ConfigurationFinalizing"
 )
 
 func EnactmentKey(node, policy string) types.NamespacedName {
--- a/cmd/handler/main.go
+++ b/cmd/handler/main.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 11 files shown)


mkowalski:
@emy this is for you, vibecoded OCPBUGS-74261. Feel free to review or feel free to drop it completely.

mkowalski:
/retest

_Comment posted with AI assistance (via OpenCode). Please verify before acting on it._

mkowalski:
/test pull-kubernetes-nmstate-e2e-upgrade-k8s

_Comment posted with AI assistance (via OpenCode). Please verify before acting on it._

Issue #1579, opened 2026-09-17
CI Failure: periodic-knmstate-e2e-handler-k8s-latest / The bridge-disabled-VLAN test failed at


qinqon:
Periodic CI job **periodic-knmstate-e2e-handler-k8s-latest** failed in run [2100404445790605312](https://storage.googleapis.com/kubevirt-prow/logs/periodic-knmstate-e2e-handler-k8s-latest/2100404445790605312/build-log.txt).

## Summary
The bridge-disabled-VLAN test failed at `test/e2e/handler/simple_bridge_and_bond_test.go:164` because SSH-based inspection of `br13` on `node04` repeatedly failed during severe node stalls. This was **not an observed VLAN-value mismatch**; the suite finished with 104 passed, 1 failed, and 2 pending.

## Root Cause
**Supported causal chain:** node04 kernel stalls made node operations unresponsive, SSH authentication timed out and subsequent connections were rejected, bridge inspection returned errors, and the test’s `Eventually` assertion expired.

Evidence from the [full build log](https://storage.googleapis.com/kubevirt-prow/logs/periodic-knmstate-e2e-handler-k8s-latest/2100404445790605312/build-log.txt) and [node04 journal](https://storage.googleapis.com/kubevirt-prow/logs/periodic-knmstate-e2e-handler-k8s-latest/2100404445790605312/artifacts/node04-journalctl.log), with journal timestamps converted from EDT to UTC:

1. **Bridge activation succeeded before the stall.** At **02:29:21 UTC**, NetworkManager reported `device (br13): Activation: successful, device activated`; both bridge ports entered forwarding state. The test setup waits for policy availability through `test/e2e/handler/utils.go:243–250`.

2. **Node04 experienced prolonged kernel stalls.** At **02:30:06**, its watchdog reported `soft lockup - CPU#0 stuck for 35s`. At **02:31:08**, RCU reported stalled CPUs/tasks. By **02:53:03**, watchdog reports included CPU stalls of **240 seconds** and **187 seconds**.

3. **SSH became unavailable during the assertion.** At **02:52:32**, sshd logged `Timeout before authentication` for a connection from the runner-side address `192.168.66.2`. At **02:53:20–29**, it repeatedly dropped connections with `penalty: exceeded LoginGraceTime`.

[...]

PR #1585, opened 2026-09-21
cluster: use NetworkManager-main-debug for latest builds

qinqon:

/kind bug

Switch `NM_VERSION=latest` to the `networkmanager/NetworkManager-main-debug` COPR and update the local-cluster documentation and agent guidance.

After #1584 moved the default provider to Kubernetes 1.37 / CentOS Stream 10, the latest-NetworkManager periodic job fails during `cluster-up` because `NetworkManager-main` has no `epel-10-x86_64` repository. `NetworkManager-main-debug` publishes both EPEL 10 and CentOS Stream 10 packages, so existing distribution auto-detection works.

Related to #1579: this addresses the September 19 and 20 cluster-setup failures linked in its comments. The original September 17 kernel-stall/SSH failure has a separate, unresolved cause.

The debug repository follows the same NetworkManager main branch. Matching builds retain `-O2`, `-g`, and LTO, but enable `more_asserts=10000` and `more_logging=true`. These additional runtime checks and diagnostic paths can affect CPU usage, timing, and assertion behavior; this is an intentional change to the latest-build testing configuration.

**Release note**:

```release-note
NONE
```


qinqon:
Closing because NetworkManager-main now has an epel-10 target, and a build is expected shortly. We will keep using the non-debug main-branch repository, so the switch to main-debug is no longer needed.

